PR #5, opened 2019-10-18
Uni10-cytnx

yingjerkao:
This PR creates a new branch of Uni10-cytnx which includes the following features which allows for uni10 build
1. CMake build with CXX, CUDA and pybind
2. Pip install for cytnx
3. Separation for internal and external headers with  modular build design. 
4. Fixed type warnings.


yingjerkao:
I believe documentation build results should not be include in the repo since they change constantly.

PR #25, opened 2021-05-30
Add linalg::Lstsq()

j9263178:
Linear solve routine from Lapacke


kaihsin:
there are some further modifications and several bugs fixed from this PR before integration. please do a diff to see the change to the commit.

Issue #32, opened 2021-07-16
invalid narrowing conversion from "unsigned long long" to "long long"


yingjerkao:
System: Mac OS X 10.15.7 (Catalina)
Compiler: Intel icpc version 2021.1
How to reproduce: Build commit 32b62ff5891f92df72ed08d4a615d3b6813836a3
Short summary: Compiler gives error due to the implicit conversion from `unsigned int` to `int`. I believe this is related in the list initialization. 
Possible Fix: use `static_cast` to explicitly cast the type
References: [Surprises and Undefined Behavior From Unsigned Integer Promotion](https://jeffhurchalla.com/2019/01/16/c-c-surprises-and-undefined-behavior-due-to-unsigned-integer-promotion/)

`====================[ Build | all | Debug ]=====================================
/Applications/CLion.app/Contents/bin/cmake/mac/bin/cmake --build /Users/yjkao/cytnx/cmake-build-debug --target all -- -j 3
[  1%] Building CXX object CMakeFiles/cytnx.dir/src/linalg/Lstsq.cpp.o
[  1%] Building CXX object CMakeFiles/cytnx.dir/src/linalg/linalg_internal_cpu/Matmul_internal.cpp.o
[  2%] Building CXX object CMakeFiles/cytnx.dir/src/linalg/linalg_internal_cpu/Matmul_dg_internal.cpp.o
/Users/yjkao/cytnx/src/linalg/Lstsq.cpp(38): error #3291: invalid narrowing conversion from "unsigned long long" to "long long"
                  bin = Tensor::from_storage(bstor).reshape({n,nrhs});
                                                             ^

/Users/yjkao/cytnx/src/linalg/Lstsq.cpp(38): error #3291: invalid narrowing conversion from "unsigned long long" to "long long"
                  bin = Tensor::from_storage(bstor).reshape({n,nrhs});
                                                               ^

/Users/yjkao/cytnx/src/linalg/Lstsq.cpp(55): error #3291: invalid narrowing conversion from "unsigned long long" to "long long"
                  sol.reshape_({n,nrhs});
                                ^

/Users/yjkao/cytnx/src/linalg/Lstsq.cpp(55): error #3291: invalid narrowing conversion from "unsigned long long" to "long long"
                  sol.reshape_({n,nrhs});
                                  ^

make[2]: *** [CMakeFiles/cytnx.dir/src/linalg/Lstsq.cpp.o] Error 2
make[2]: *** Waiting for unfinished jobs....
[...]

PR #39, opened 2021-08-25
Fixes issue #32

yingjerkao:
This PR fixes issue #32

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,13 +2,12 @@
 ######################################################################
 ### CMAKE VERSION
 ######################################################################
-cmake_minimum_required(VERSION 3.11)
+cmake_minimum_required(VERSION 3.12)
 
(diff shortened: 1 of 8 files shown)


yingjerkao:
> 
> 
> Actually, intrusive_ptr does have a moving constructor, but I need to think about whether it will mess up the ref count.

If it is implemented properly, the ref_count should be also moved.

kaihsin:
no no, that's not what I am worrying about. It is taking cared of by boost. The current implementation doesn't copy the content of the data, it creates a new pointer. Without move, it creates a new pointer that copies the things of intrusive_ptr, not the actual content. The move will be like moving input to the current pointer. 

The thing I am worrying is whether the old one is moved and destroy will lead to a problem here. 

I think in this specific case of Tproxy it is ok.

kaihsin:
[Note] 
complex_arithmetic* changes are rejected.

PR #40, opened 2021-09-13
Add example/vumpsMPO.py

j9263178:
1. Need Arnoldi for solving eigenvector of transfer matrix. 
2. Lancoz_Gnd doesn't work, Lancoz_ER is found to be much slower than scipy.eigsh (Investigating).
3. Might need bicgstab (currently called from scipy)


pcchen:
I would suggest to add wrapper to arpack-ng so that one can use Arnoldi type algorithms for eigenvalues/vectors and SVD.

PR #62, opened 2022-04-27
Prevent tensor being scalar

jeffry1829:
Do not delete labels, bonds of one element tensor.

This results in some bug that's very hard to debug.


kaihsin:
I think there is a reason for this when doing Network, disable it probably will affect some functions there

jeffry1829:
The difference between the old version and new version is that the bonds and labels of Scalar tensor will be cleared.
I guess there'll be no function depend on the emptiness of bonds and labels?

PR #63, opened 2022-04-27
change Kron_internal_dtd and fix padding performance

jeffry1829:
I fix the O(n^2) inserting in Kron.cpp, and remove redundant vector in Kron_internal_dtd.

(I had only change dtd function, maybe we should consider generic type instead of copy same code?)


kaihsin:
@jeffry1829 I thought we don't want to use macro but using template instead?

jeffry1829:
@kaihsin Initially, I want to use only template.
However, there’re 11*11=121 functions,~6000 lines with only one line(invoking template function with different Types). I thought this is extremely hard to maintain, and the logic of different implementations of different functions are not so clear.

I agree that macro is not suitable for every internal.cpp functions, however, I guess macro works at least for Kron_internal.

(We may do explict implementation to functions we want, and then use macros to generate mass majorities of other identical functions? )

kaihsin:
@jeffry1829 Why is there 6000 lines? Maybe we can use the template on the body fucntion such that they accept pointers, and the internal function calling it?  
I am a bit worrying about the debugging, when doing macro... 

template<class T1, class T2, class T3>
Kron_body(T1* out, T2* Lin, T3* Rin,....){
     //implementation
}

Kron_internal_*d*(boost::intrusive_ptr<Storage_base> & out, const boost::intrusive_ptr<Storage_base> & Lin, const boost::intrusive_ptr<Storage_base> & Rin ...){
    cytnx_complex128 *_out = (cytnx_complex128*)out->Mem;
    cytnx_complex128 *_Lin = (cytnx_complex128*)Lin->Mem;
    cytnx_complex128 *_Rin = (cytnx_complex128*)Rin->Mem;

    Kron_body(pout,pLin,pRin ...)

}


...

jeffry1829:
6000 lines comes from unoptimized Kron with maximum line length 100.

Yes, debugging is a big problem. Maybe I’ll comment the macro used to generate these functions, and copy the generated function definitions explicitly to Kron…cpp?
This, we know the logic behind generating these function while we can debug with ease.

PR #102, opened 2022-06-27
Add .clang-format to project

jeffry1829:
The only thing that might be annoying is that pragma cannot be formatted correctly.
We can only do
// clang-format off
CODES WE DON'T WANT BE MODIFIED
// clang-format on


jeffry1829:
For it to be easier to merge PRs after this. We can inspect only commits before clang-format
Or, compare master with PR branch, and follow https://stackoverflow.com/questions/55479729/github-pull-request-shows-wrong-diff
We can get true diff

PR #146, opened 2022-12-27
Add unit test for Hstack and Vstack

hunghaoti:
Add unit test for Hstack and Vstack, and fix the error in Hstack and Vstack.
1. Add unit test for Hstack and Vstack
2. Fix the error in Hstack and Vstack when checking the input data.


hunghaoti:
Fix the unit test for Hstack and Vstack.

PR #219, opened 2023-08-02
Hao ti

hunghaoti:
Modify the api doc as Chia-Min comments:
1. add pre-condition to Exp needs to be a Hermitian (rank-2) tensor.
2. For Tensor operator +-*/ need to have same shape.
3. fix wrong link for operator== in seealso.
4. Use convenstion M=VDV^{-1}.

### DIFF
--- a/include/linalg.hpp
+++ b/include/linalg.hpp
@@ -20,6 +20,7 @@ namespace cytnx {
    * @param[in] Lt The left UniTensor.
    * @param[in] Rt The right UniTensor.
    * @return [UniTensor] The result of the addtion.
+   * @pre \p Lt and \p Rt must have the same shape.
    * @see linalg::Add(const UniTensor &Lt, const UniTensor &Rt)
    */
   cytnx::UniTensor operator+(const cytnx::UniTensor &Lt, const cytnx::UniTensor &Rt);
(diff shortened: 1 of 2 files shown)


include/linalg.hpp
```diff
@@ -2103,18 +2108,31 @@ namespace cytnx {
...
+    @param[in] in input Tensor, should be Hermitian
+    @param[in] a rescale factor
+    @param[in] b bias
     @return
         [Tensor]
-
-    description:
-        perform matrix exponential with \f$O = \exp{aM + b}\f$.
+    @pre the \p in should be a Hermitian matrix.
+    @warning If \p in is not a Hermitian matrix, the result will be undefined.
```
kaihsin, line 2121: I suggest:
```
If `p` is not a Hermitian matrix, only the lower triangular matrix will be used. (This is strongly not recommended. use ExpM[link to ExpM] instead)
```

Also since ExpM and ExpH both use Eig and Eigh you might need to do modifications there as well

kaihsin (changes requested):
Please make corresponding change(s) thx

include/linalg.hpp
kaihsin: remember to modify the Eigh as well (only lower triangular will be used)

kaihsin (approved):

kaihsin (approved):

Issue #223, opened 2023-08-04
Svd_truncate for gpu


jeffry1829:
1. Return all truncated singular values
2. Return cuda gesvdp error as truncation error in Svd_truncate

kaihsin:
This should be also related to CPU.

Issue #202, opened 2023-07-24
Svd_truncate, the returned value when return_err=True


chiamin:
When running `Svd_truncate` with argument `return_err=True`, it will return the largest singular value that is truncated. I suggest to return all the truncated singular values. In DMRG the truncation error is defined by summing all the truncated p_i=s_i*s_i where s_i is the singular value. In this way we can compute the truncation error.

kaihsin:
I suggest to extend the return_err:


return_err="full" (or simply a integer ) will return all the truncated eigenvalues, for example

PR #225, opened 2023-08-05
svdtrunc return

kaihsin:
This PR address #202, and part of #223, 

[NOTE] This PR depending on #224, dont merge this before #224. 

1. Svd/Gesvd_truncate, return_err is now bool -> unsigned int. 
   when return_err = False (or 0) : no return
              return_err = True (or 1): return max err in the truncated list
              retunr_err = other int: return full truncated errors

### DIFF
--- a/src/linalg/Svd_truncate.cpp
+++ b/src/linalg/Svd_truncate.cpp
@@ -177,7 +177,7 @@ namespace cytnx {
 
       outCyT = linalg::Svd(Tin, is_UvT);
 
-      // process runcate:
+      // process truncate:
       // 1) concate all s vals from all blk
       Tensor Sall = outCyT[0].get_block_(0);
       for (int i = 1; i < outCyT[0].Nblocks(); i++) {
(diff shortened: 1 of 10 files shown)


src/linalg/Svd_truncate.cpp
```diff
@@ -167,12 +172,12 @@ namespace cytnx {
...
     void _svd_truncate_Block_UT(std::vector<UniTensor> &outCyT, const cytnx::UniTensor &Tin,
                                 const cytnx_uint64 &keepdim, const double &err, const bool &is_UvT,
-                                const bool &return_err) {
+                                const int &return_err) {
       cytnx_uint64 keep_dim = keepdim;
 
       outCyT = linalg::Svd(Tin, is_UvT);
 
-      // process truncate:
+      // process runcate:
```
jeffry1829, line 180: why runcate

jeffry1829 (approved):

PR #229, opened 2023-08-07
Gpu

jeffry1829:
fix CMakeList output LD_LIBRARY file
rewrite cuGesvd(still not working)
add equivelem api for Tensor&Storage


jeffry1829:
I found the way to fix cuGesvd, will commit soon

clang-formater.sh
kaihsin, line 1: btw, you can just install `pre-commit` tool, it will automatically format for you everytime you `git commit` (See Developer_tools/)

include/Tensor.hpp
kaihsin, line 1: Maybe renaming it to approx_eq or eq_approx?

src/BoolStorage.cpp
kaihsin, line 1: This will not work. if the rhs is not bool, this will fail. Same as other type. 
Also, approx eq you will only need for floating types. So basically for each Storage type, you need to have all the data types (e.g. 

ex: 

BoolStorage:

     bool approx_eq_d(const Storage/or intrusive_ptr<Storage_base>, double tol);
     bool approx_eq_f(const Storage/or intrusive_ptr<Storage_base>, double tol);   
     bool approx_eq_cd(const Storage/or intrusive_ptr<Storage_base>, double tol);
     bool approx_eq_cf(const Storage/or intrusive_ptr<Storage_base>, double tol);
    ......

include/Storage.hpp
kaihsin, line 1: See below

src/linalg/Gesvd.cpp
kaihsin, line 1: The second one you don't need to permute the input tensor! you only need to 

1. make the cusolver do hermitian conjugate (with CUDA_H or something like that) 
2. swap the return U, Vt -> Vt -> U
3. permute/conjugate the U and Vt

kaihsin (changes requested):

src/linalg/Gesvd.cpp
jeffry1829: I don't understand why I swap the N,M passed to cugesvd api didn't transpose the input tensor, I'm kinda stuck in this idea.

Also, I don't find hermitian conjugate parameter passable for cugesvd :(

src/BoolStorage.cpp
jeffry1829: I guess I'll convert input storage to lhs dtype if the type is "less" than lhs.
This way, we can compare elements more naturally

jeffry1829: fix pushed

src/linalg/Gesvd.cpp
jeffry1829: I suggest we deal with this after this PR

jeffry1829: Fixed, substitute gesvd with gesvdj

kaihsin (changes requested):
The scalar will still not work... you can't have only one function because you need to deal with strong/weak type. converting the rhs using .to_<dtype> will not work when lhs is stronger than rhs. In the case rhs is stronger, it will either fail to convert or just give you even non-correct results! 

You need to do for all the types separately, like I do (for example greater())

jeffry1829:
But I check the type order first, and then choose to convert lhs or rhs, won’t that work?

jeffry1829:
Why it doesn’t work when lhs>rhs? Sorry I don’t get it

jeffry1829:
Oh I do the bi-directional astype conversion inside every XXXStorage.cpp
And forgot to do it on Scalar or Tensor approx_eq

kaihsin:
Yes. Sorry I wasn't being specific

jeffry1829:
> Yes. Sorry I wasn't being specific

So does it work if I convert both lhs and rhs in Scalar and Tensor approx_eq?

jeffry1829:
commented out approx_eq for this PR.
Note that linalg_Test.BkUt_Svd_truncate1 test failed

kaihsin (approved):

Issue #239, opened 2023-08-18
The Svd_truncate is extremely slower than Svd in gpu version.


hunghaoti:
As title, 

>Svd with D=500 on gpu:  87 ms
>Svd_truncate with D=100 on gpu:  11818 ms

you can reproduce by running the following code and adjust the bond dimension.
https://github.com/Cytnx-dev/Cytnx/blob/dffe31ddb6965d11f681d9c69115bf552a43a0ae/bm_tests/linalg/Svd_truncate_bm.cpp#L49-L59
https://github.com/Cytnx-dev/Cytnx/blob/dffe31ddb6965d11f681d9c69115bf552a43a0ae/bm_tests/linalg/Svd_bm.cpp#L83-L93

kaihsin:
@hunghaoti  have you managed to test this yet?

hunghaoti:
> @hunghaoti have you managed to test this yet?

Hi @kaihsin 
for bond dimension d*D= [10, 100, 200, 400, 600, 800, 1000, 2000, 5000]
The bench mark test results as following (on A100 GPU):
for this commit: a18c307a0624c3a420377960db802121acabdfce (the original version)
time = [5.94, 662, 11818, X, X, X, X, X, X] ms

for this commit: c1d4bc665daa311aaea76e350694ae00ae1edaea (using cuQuantum)
time = [5.09, 9.79, 18.1, 41.3, 70.2, 109, 157, 640, 6919] ms

for this commit: a3f7ce3a09ac82337520f06d51fa48d2800c4d6f (after merge to dev-master)
time = [5.58, 11.3, 20.4, 46.1, 77.0, 118, 169, 666, 7049] ms

 I think this issue can be close if there is no problem.

kaihsin:
Cool thanks!

PR #243, opened 2023-08-24
cuQuantum SVD(truncate)

j9263178:
This PR address https://github.com/Cytnx-dev/Cytnx/issues/239, also #68

Now `Gesvd_truncate()` support cuQuantum/cuTensornet's SVD routine when the input is on cuda device.

Currently only the complex double type is supported, will finished the rest types in the following commits.

### DIFF
--- a/src/linalg/Gesvd_truncate.cpp
+++ b/src/linalg/Gesvd_truncate.cpp
@@ -5,10 +5,9 @@
 #include "UniTensor.hpp"
 #include "algo.hpp"
 
-
 #ifdef UNI_GPU
   #ifdef UNI_CUQUANTUM
-      #include "linalg_internal_gpu/cuQuantumGeSvd_internal.hpp"
+    #include "linalg_internal_gpu/cuQuantumGeSvd_internal.hpp"
   #endif
 #endif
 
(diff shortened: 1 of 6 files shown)


j9263178:
I noticed that the way it handles the `keepdim` and `err` argument is not consistent with other gesvd_truncate functions, will change the behavor in the following commits.

j9263178:
It seems that we cannot return desired result for `return_err != false`, I didn't find API in cuTensornet allow us to access the discarded singular values and states.

kaihsin:
There is a way to get the largest truncated singular value, but I also didn't find the way to get full truncated svs

j9263178:
> is it possible that we call different Svd internal base on return_err ?
> 
> * False, or largest truncated error -> we use cutensornet SVDtruncate
> * full truncated error -> we use cutensornet SVD, and manually do truncation

Yes, I can do that switch within this PR.

In the future we might still need to optimize the manual truncation step for gpu, see #239

kaihsin:
I wondering if cutensornet/cuquantum has any slicing API we can call?

kaihsin:
a naive solution is to make Svd_internal_gpu to also handle slicing (which is just resize a matrix by dropping out the tails for U, and for vT its stride copy), before you permute the U and vT back

kaihsin (approved):

j9263178:
The behavior for passing `keepdim` and  `return_err` is now consistent with the other Svd_truncate functions.
Note that a manual truncation in addition to the cuquantum svd routine will happen when:

1. The `return_err` is required.
2. `dim_after_cuquantum_truncation`<`keepdim` (remove those zeros).

We might need to do a benchmark to see if there are still overheads in this manual truncation step.

kaihsin (approved):

PR #251, opened 2023-09-07
Dev master merge to master

kaihsin:
This PR merge the dev-master into master


j9263178 (approved):

kaihsin:
@j9263178 @jeffry1829 
There is segfault for one of the test!
image

kaihsin:
Also a couple failed

image

kaihsin:
test case 186: 
    What is the correct testing data? the one I got from the file is actually U1 diagonal UniTensor not a U1 diagonal UniTensor with single element! 
    
    It failed becase diagonal Svd is not support. 
    
test case 14: 
   a few bugs there (1) Seg fault is due to reading `output.txt` and `answer.txt` path is wrong, it should be `utils/output.txt` instead 
   After fixing that there is still bugs causing it failed.

Issue #203, opened 2023-07-24
[Bug] QR doesn't work


chiamin:
The following code does not work

```
import cytnx
# Create a rank-3 Tensor object with shape [2,2,2]
T = cytnx.arange(4*4*4).reshape(4,4,4)
# Convert it to a UniTensor
M = cytnx.UniTensor(T, labels=["a","b","c"])
# 1. Permute the indices to ("c","a","b") ordering,
M.permute_(["c","a","b"])
# 2. Set rowrank=1, so that "c" is the row index,
# and "a", "b" are the column indices.
M.set_rowrank(1)

# Perform SVD without truncation and returns only the singular values
Q, R = cytnx.linalg.Qr (M)
#
M2 = cytnx.Contract(Q, R)

diff = M - M2
print(diff.Norm())
```

chiamin:
Here is the error message.

```
Traceback (most recent call last):
  File "/home/chiamin/project/2023/cytnx/test.py", line 13, in <module>
    Q, R = cytnx.linalg.Qr (M)
RuntimeError: 
# Cytnx error occur at void cytnx::linalg_internal::QR_internal_d(const boost::intrusive_ptr<cytnx::Storage_base>&, boost::intrusive_ptr<cytnx::Storage_base>&, boost::intrusive_ptr<cytnx::Storage_base>&, boost::intrusive_ptr<cytnx::Storage_base>&, boost::intrusive_ptr<cytnx::Storage_base>&, const cytnx::cytnx_int64&, const cytnx::cytnx_int64&, const bool&)
# error: Error in Lapack function 'dorglq': Lapack INFO =  -5
# file : /home/kaihsinwu/.pyenv/versions/anaconda3-5.3.1/conda-bld/cytnx_1689515964188/work/src/linalg/linalg_internal_cpu/QR_internal.cpp (177)
```

kaihsin:
You are trying to do QR on a matrix with all zeros. This is a mathematical ill-defined procedure therefore you get lapack error

kaihsin:
Can we add the above example into a testing case? @j9263178 @hunghaoti

hunghaoti:
> Can we add the above example into a testing case? @j9263178 @hunghaoti

Of course, but maybe we need to convert to C++ version.

j9263178:
> Can we add the above example into a testing case? @j9263178 @hunghaoti

I will do it in next PR.

kaihsin:
[...]

PR #254, opened 2023-09-09
Fix QR bug and cuQuantum QR

j9263178:
Part of this PR address https://github.com/Cytnx-dev/Cytnx/issues/203 and https://github.com/Cytnx-dev/Cytnx/issues/68.

cuQuantum QR routine will now be used when the input UniTensor is on cuda device.
In that routine I currently disable the `return_tau = true` option since the API doesn't store tau or provide ways to retrieve it.


kaihsin:
Make sure all the  tests are passed

kaihsin (approved):

Issue #245, opened 2023-08-27
Gesvd issue


hunghaoti:
I would like to ask whether we should use get_block() instead of get_block_().
Since it looks like it will change the data in `Tin`.
https://github.com/Cytnx-dev/Cytnx/blob/c1d4bc665daa311aaea76e350694ae00ae1edaea/src/linalg/Gesvd.cpp#L92-L92
https://github.com/Cytnx-dev/Cytnx/blob/c1d4bc665daa311aaea76e350694ae00ae1edaea/src/linalg/Gesvd_truncate.cpp#L89-L89

You can reproduce be the following code:
```c++
int D = 10;
auto bd_vi = Bond(D, BD_IN);
auto bd_pi = Bond(2, BD_IN);
auto bd_po = Bond(2, BD_OUT);
auto bd_vo = Bond(D, BD_OUT);
std::vector<Bond> bonds = {bd_vi, bd_pi, bd_po, bd_vo};
cytnx_int64 row_rank = 2;
std::vector<std::string> labels = {};
bool is_diag = false;
auto UT = UniTensor(bonds, labels, row_rank, Type.Double, Device.cuda, is_diag);
double low = -1000, high = 1000;
int rnd_seed = 0;
random::Make_uniform(UT, low = -1000, high = 1000, rnd_seed);
std::cout << UT;
std::vector<UniTensor> svds = linalg::Gesvd_truncate(UT, D);
std::cout << UT;

```

jeffry1829:
This will be fixed in the next commit, seems like in Gesvd_truncate(cuquantum version), we didn't reshape_ tmp Tensor to its original shape

hunghaoti:
I have checked this issue and it is solved.

PR #261, opened 2023-09-11
Cuda memcpy truncation

j9263178:
Unified the cuda svd truncation step into `cytnx::linalg_internal::cudaMemcpyTruncation_X()`, since several svd functions share the same behavior( `cuQuantumGeSvd`, cuSolver version of `Gesvd_truncate` and `Svd_truncate`.).

Also part of this PR address https://github.com/Cytnx-dev/Cytnx/issues/245.

### DIFF
--- a/src/linalg/Svd_truncate.cpp
+++ b/src/linalg/Svd_truncate.cpp
@@ -4,6 +4,7 @@
 #include "Tensor.hpp"
 #include "UniTensor.hpp"
 #include "algo.hpp"
+#include "linalg_internal_interface.hpp"
 
 namespace cytnx {
   namespace linalg {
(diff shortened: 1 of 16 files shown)


src/linalg/Svd_truncate.cpp
```diff
@@ -12,53 +13,77 @@ namespace cytnx {
...
+
+        return tmps;
+      } else {
+#ifdef UNI_GPU
+        std::vector<Tensor> tmps = Svd(Tin, is_UvT);
+        Tensor terr({1}, Tin.dtype(), Tin.device());
 
-      return tmps;
+        cytnx::linalg_internal::lii.cudaMemcpyTruncation_ii[Tin.dtype()](
+          tmps[1], tmps[2], tmps[0], terr, keepdim, err, is_UvT, is_UvT, return_err);
```
kaihsin, line 70: Cool! should we do the same for CPU side ?

j9263178: Yes, I will add the similar functions for cpu case.

j9263178:
Added truncation for cpu, while currently there will be segfault, will fix soon.

kaihsin (approved):

PR #275, opened 2023-09-27
tensor test 2

daniel0321forever:


tests/Tensor_test_2.cpp
```diff
@@ -0,0 +1,762 @@
...
+  for (int i = 0; i < 2; i++){
+    for (int j = 0; j < 2; j++){
+      T1(i,j) = values[i][j];
+    }
+  }
+  auto SVDs = T1.Svd();
+  std::cout << SVDs[0] << std::endl;
+  std::cout << SVDs[1] << std::endl;
+  std::cout << SVDs[2] << std::endl;
+}
```
kaihsin, line 759: This test need to have checking on the elements before/after SVD, as well as element of U, vT

kaihsin (changes requested):

kaihsin:
Also, make sure you comply with clang format, the easiest way is to use pre-commit tool

hunghaoti:
I will merge and review it.

Issue #286, opened 2023-10-02
relabel_ with string list args throw errors


j9263178:
```
import cytnx
from cytnx import Tensor, zeros, linalg, Device, Type, Contract, UniTensor,eye

A = UniTensor(zeros([3,4,5]))
print(A.labels())
A.relabels_(["0","2"],["a","b"])
print(A.labels())
```

gives 

```
RuntimeError: 
# Cytnx error occur at void cytnx::UniTensor_base::set_labels(const std::vector<std::__cxx11::basic_string<char> >&)
# error: [ERROR][set_labels][UniTensor] the len(new_labels) does not match the rank of the UniTenosr

# file : /home/j9263178/Cytnx/src/UniTensor_base.cpp (265)
```

j9263178:
#231

kaihsin:
This is fixed on #291

PR #287, opened 2023-10-02
fix @addmethod override for UniTensor relabel_/relabels_/at

### LINKED ISSUE #231: add_method macro for overloading with signature overlapping does not work
```python
@add_method(UniTensor)
def relabels_(self, a:List[str],b:List[str]):
      ...

@add_method(UniTensor)
def relabels_(self, a:List[str]):
    ...
```

The second one will be ignored (not accessible)

jeffry1829:
Fixed #231, #286

### DIFF
--- a/cytnx/UniTensor_conti.py
+++ b/cytnx/UniTensor_conti.py
@@ -180,14 +180,20 @@ def set_name(self,name):
 
 
 @add_method(UniTensor)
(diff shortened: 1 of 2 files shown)


kaihsin (changes requested):
Per discussion, This should be address on the `@add_method` macro to make wrapping of overload functions possible

PR #305, opened 2023-10-14
- fixed wrong argument name in UniTensor_base::relabel(const std::str…

manuschneider:
- fixed wrong argument name in UniTensor_base::relabel(const std::string &inx, const std::string &new_label)
- changed argument names in many functions from lbl(s) to label(s), to be consistent with old_label(s)/new_label(s) and allow for a more intuitive call when argument names are specifically used in python

### DIFF
--- a/include/Network.hpp
+++ b/include/Network.hpp
@@ -132,8 +132,8 @@ namespace cytnx {
     virtual UniTensor Launch();
 
     virtual void construct(const std::vector<std::string> &alias,
-                           const std::vector<std::vector<std::string>> &lbls,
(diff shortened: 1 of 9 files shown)


jeffry1829:
Hi, maybe you will want to install pre-commit tool
It'll automatically check format for you when commiting

kaihsin:
Is this only involved arguments variable name changes?

manuschneider:
Yes, only argument and some variable names changed.
Thanks @jeffry1829 for fixing the problems with clang-format (lines became too long).

jeffry1829 (approved):

PR #306, opened 2023-10-14
Add Scalar backend

jeffry1829:
Not fully tested, at least compile passed.

It's important to review SProxy, since we're not using straightforward way to write into memory.
Instead, we first convert memory to at::Tensor, then use `tn[loc] = elem; `

### DIFF
--- a/include/backend_torch/Scalar.hpp
+++ b/include/backend_torch/Scalar.hpp
@@ -4,22 +4,23 @@
 #include "Type.hpp"
 #ifdef BACKEND_TORCH
   #include <torch/torch.h>
-
 namespace cytnx {
   class Scalar : public c10::Scalar {
    public:
     int _dtype = Type.Void;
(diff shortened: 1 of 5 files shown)


kaihsin:
This need to be careful to not mess-up the gradient part.

include/backend_torch/Scalar.hpp
kaihsin, line 1: is it possible to not using switch? I think torch must have their own function you can call?

jeffry1829: Since we're dealing with cytnx types. I think we cannot use pytorch.

IvanaGyro:
I think this PR is outdated because Scalar is small in memory. There is no need to add one more layer like Storage to reuse the memory.  Furthermore, connecting pytorch as a backend is not a near-term milestone now. We may be able to close this PR.

Issue #311, opened 2023-10-18
linalg::Norm for UniTensor


j9263178:
also a couple functions that is trivial to implement for denseUt.

PR #312, opened 2023-10-19
Adding denseUt linalg::Norm/Eig/Eigh

j9263178:
This PR address https://github.com/Cytnx-dev/Cytnx/issues/311,
The Inv/InvM part is not fully done since there are lapack errors from very basic test, I'm investigating.

### DIFF
--- a/src/linalg/Eig.cpp
+++ b/src/linalg/Eig.cpp
@@ -2,6 +2,7 @@
 
 #include <iostream>
 #include "Tensor.hpp"
+using namespace std;
 
 #ifdef BACKEND_TORCH
 #else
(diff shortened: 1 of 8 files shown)


j9263178:
Will do `Inv` and `InvM` in another PR.

src/linalg/Eig.cpp
```diff
@@ -92,4 +93,107 @@ namespace cytnx {
...
+      tmps.clear();
+      vector<string> oldlabel = Tin.labels();
+
+      // collapse as Matrix:
+      cytnx_int64 rowdim = 1;
+      for (cytnx_uint64 i = 0; i < Tin.rowrank(); i++) rowdim *= tmp.shape()[i];
+      tmp.reshape_({rowdim, -1});
+
+      vector<Tensor> outT = cytnx::linalg::Eig(tmp, is_V, row_v);
+      if (Tin.is_contiguous()) tmp.reshape_(oldshape);
```
jeffry1829, line 126: Why reshape_ when contiguous?

src/linalg/Eig.cpp
```diff
@@ -92,4 +93,107 @@ namespace cytnx {
...
+        labelU.push_back(Cy_S.labels()[0]);
+        Cy_U.set_labels(labelU);
+        t++;  // U
+      }
+      // if tag, then update  the tagging informations
+      if (Tin.is_tag()) {
+        Cy_S.tag();
+        t = 1;
+        if (is_V) {
+          cytnx::UniTensor &Cy_U = outCyT[t];
```
jeffry1829, line 158: The label informations are removed

src/linalg/Norm.cpp
```diff
@@ -58,6 +58,17 @@ namespace cytnx {
       cytnx_error_msg(uTl.uten_type() != UTenType.Dense,
                       "[Error][Norm] Can only use Norm on DenseUniTensor or Tensor%s", "\n");
```
jeffry1829, line 59: Can't we use Norm on BkUT?

j9263178: added in the latest commit.

src/linalg/Eig.cpp
j9263178: tmp refer to the Tin block so we need to reshape it back.

src/linalg/Eig.cpp
j9263178: Fixed in the latest commit. Note that in current version output eigenvector matrix will be untagged dense square ut regardless of Tin's rank and tags, since I find maintaining certain shape, tag and label info in the output with respect to the Tin might be complicated and confusing to users.

jeffry1829: @kaihsin Should we do that?

PR #318, opened 2023-10-20
Add label argument for put_block/put_block_

### LINKED ISSUE #285: put_block() using labels
Similar to `get_block().`

jeffry1829:
To resolve #285


pybind/unitensor_py.cpp
kaihsin, line 1: I think the keyword args should be `py::arg("qidx")` instead of `py::arg("qnum")`? 

Could you modify them for binding on `get_block` and `put_block`

kaihsin (changes requested):

pybind/unitensor_py.cpp
jeffry1829: done

kaihsin (approved):

Issue #319, opened 2023-10-20
get() set() for UniTensor


manuschneider:
At the moment, the elements of a UniTensor can only be accessed by
1) .at(), which only returns one element
2) accessing the block (Tensor) and then the data

Since we want to avoid dealing with Tensors, it would be great if get() and set() can be supported, at least for dense UniTensor.

Sugar code like uT[0,:,1:4:2] would be amazing for UniTensors as well, but for the paper we need at least get() and set().

kaihsin:
get and set us already there

pcchen:
What is the syntax of get() and get() for UniTensor? (If it exists.)

j9263178:
> What is the syntax of get() and get() for UniTensor? (If it exists.)

Currently it has no pybind API, will add them soon.

j9263178:
> What is the syntax of get() and get() for UniTensor? (If it exists.)

@manuschneider  @pcchen I think things like `Ut[:, 0:1, 1]`  should already work in python.

pcchen:
On which branch and after which commit?

pcchen:
OK. It seems to work for me with version 0.9.5.

pcchen:
This works 
````
UT1 = cytnx.UniTensor(cytnx.arange(4*4*4).reshape(4,4,4),labels=["a","b","c"])
UT2 = cytnx.UniTensor(cytnx.arange(8*8*8).reshape(8,8,8),labels=["a","b","c"])
print(UT1[1:3,2,3])
print(UT2[1:3,2,3])
````

But these will fail (for Tensor it works)
````
UT1[1,2,3] = -1.0
UT1[1:3,2,3] = UT2[1:3,2,3]
````

j9263178:
> This works
> 
> ```
> UT1 = cytnx.UniTensor(cytnx.arange(4*4*4).reshape(4,4,4),labels=["a","b","c"])
> UT2 = cytnx.UniTensor(cytnx.arange(8*8*8).reshape(8,8,8),labels=["a","b","c"])
> print(UT1[1:3,2,3])
> print(UT2[1:3,2,3])
> ```
> 
> But these will fail (for Tensor it works)
> 
> ```
> UT1[1,2,3] = -1.0
> UT1[1:3,2,3] = UT2[1:3,2,3]
> ```

The last two lines should work now in the corresponding PR, plx build and check.

pcchen:
This is to confirm that last two lines work in the corresponding PR.

[...]

PR #323, opened 2023-10-23
Adding slice set between DenseUniTensor

j9263178:
https://github.com/Cytnx-dev/Cytnx/issues/319.

### DIFF
--- a/include/UniTensor.hpp
+++ b/include/UniTensor.hpp
@@ -3009,7 +3009,12 @@ namespace cytnx {
     void set(const std::vector<Accessor> &accessors, const Tensor &rhs) {
       this->_impl->set(accessors, rhs);
     }
-
+    void set(const std::vector<Accessor> &accessors, const UniTensor &rhs) {
+      cytnx_error_msg(
+        rhs.uten_type() != UTenType.Dense,
+        "[ERROR] cannot set elements from UniTensor with symmetry. Use at() instead.%s", "\n");
+      this->_impl->set(accessors, rhs.get_block());
+    }
     /**
     @brief Reshape the UniTensor.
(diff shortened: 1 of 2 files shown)


include/UniTensor.hpp
kaihsin, line 1: You didn't consider if the  incoming UT has i`s_diag=True` and/or the current UT `is_diag=True`

kaihsin (changes requested):

include/UniTensor.hpp
j9263178: @kaihsin  What's the expected behavior of get/set for `is_diag=True` ut? Currently I assume that user should always regard `is_diag=True` ut as an ut having rank-1 block when doing get/set, so they can only do things like `diag_ut[i]` (they can only access the diag part)  but not `diag_ut[i,j,k]` which might gives the non-diag part. Let me know if you think user should be able to do the latter.

kaihsin: It's not the tensor anymore so the user should be able to assign (is_diag=true) block to the UT with dense, and vice versa (which include changing the current tensor to is_diag=false)

j9263178: So to be more specific u want sth like user can slice a diag_true_ut (which create a diag_false_ut since there will be non diag zeros due to the slicing), and user can assign  things(might be an element or a block) to a diag_true_ut which may make it into an dense diag_false_ut.

kaihsin:
Yes. The most important one is if the incoming Unitensor is diag UT, you
need to make conversion first before assign

kaihsin:
The behavior currently will be:

Reject all set by unitensor with is_diag=True (both incoming, and currrent instance)

jeffry1829 (approved):

PR #327, opened 2023-10-25
Implement UniTensor arange/linspace with shape

### LINKED ISSUE #325: UniTensor.arange(int lst) does not exist
It would be great if arange would work with different shapes then vectors as well.
In the draft, there is a codeline:
```
uT = cytnx.UniTensor.arange([2,3,4], labels=["a","b","c"], name="tensor uT")
```

It does not work, since arange only takes integer arguments as size. I would suggest that this is implemented, otherwise it is a bit complicated to use arange (then reshape, then relabel).

jeffry1829:
Fixed #325

### DIFF
--- a/include/Generator.hpp
+++ b/include/Generator.hpp
@@ -142,26 +142,6 @@ namespace cytnx {
   */
   Tensor arange(const cytnx_double &start, const cytnx_double &end, const cytnx_double &step = 1,
                 const unsigned int &dtype = Type.Double, const int &device = Device.cpu);
-  /**
-    @brief create an rank-1 Tensor with elements defined in range [start,end) with assigned
-    step-size, then reshape to the shape specified by the input shape. Where Nelem is the total
-    number of elements in the input shape.
-    @param shape the shape of the Tensor
-    @param start the start value of the range
-    @param end the end value of the range
-    @param step the step-size of the range
-    @param dtype the dtype of the Tensor. It can be any type defined in \link cytnx::Type
-    cytnx::Type \endlink
-    @param device the device that the Tensor is put on. It can be any device defined in \link
-    cytnx::Device cytnx::Device \endlink
-
-    @return
-        [Tensor]
-
-    */
-  Tensor arange(const std::vector<cytnx_uint64> &shape, const cytnx_double &start,
-                const cytnx_double &end, const cytnx_double &step = 1,
-                const unsigned int &dtype = Type.Double, const int &device = Device.cpu);
   //@}
 
   /**
@@ -178,20 +158,6 @@ namespace cytnx {
   Tensor linspace(const cytnx_double &start, const cytnx_double &end, const cytnx_uint64 &Nelem,
                   const bool &endpoint = true, const unsigned int &dtype = Type.Double,
                   const int &device = Device.cpu);
-  /**
-      @brief create an rank-1 Tensor with Nelem elements defined in range [start,end), which is
-     evenly spaced. Then reshape to the shape specified by the input shape.
-      @param start the start value of the range
-      @param end the end value of the range
-      @param Nelem the number of elements
-      @param endpoint if true, the end value is included in the range.
-      @param dtype the dtype of the Tensor. It can be any type defined in \link cytnx::Type \endlink
-      @param device the device that the Tensor is put on. It can be any device defined in \link
-      cytnx::Device \endlink
-    */
-  Tensor linspace(const std::vector<cytnx_uint64> &shape, const cytnx_double &start,
-                  const cytnx_double &end, const cytnx_uint64 &Nelem, const bool &endpoint = true,
-                  const unsigned int &dtype = Type.Double, const int &device = Device.cpu);
 
   // Tensor rand(const std::vector<cytnx_uint64> &size);
   // Tensor randn(const std::vector<cytnx_uint64> &size);
--- a/include/Generator.hpp
+++ b/include/Generator.hpp
@@ -38,7 +38,7 @@ namespace cytnx {
       [Tensor]
 
   */
-  Tensor zeros(const std::vector<cytnx_uint64> &Nelem, const unsigned int &dtype = Type.Double,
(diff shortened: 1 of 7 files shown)


kaihsin:
What is going on with the test No. 520 (BkUT_Norm)

hunghaoti:
> What is going on with the test No. 520 (BkUT_Norm)

Hi @kaihsin , did you mean this modification:
df12df8
Because for original code user can use put_block to put the Tensor into the UniTensor even there are not same dType. After the commit 
https://github.com/Cytnx-dev/Cytnx/commit/58897042447f4c6ab4d18cad8a71b66c118f13bf#L123
this function will check whether the dType of the input tensor same as UniTensor, see
https://github.com/Cytnx-dev/Cytnx/blob/58897042447f4c6ab4d18cad8a71b66c118f13bf/include/UniTensor.hpp#L1430-L1433
So the original test code will throw the error at line 340, see 
https://github.com/Cytnx-dev/Cytnx/blob/276507974c4cc974291a7318ac906d0825dcdf77/tests/linalg_test/linalg_test.cpp#L330-L345

kaihsin:
Why not just try to cast it, I don't know what is going on

include/Generator.hpp
```diff
@@ -132,11 +142,56 @@ namespace cytnx {
...
+    @param dtype the dtype of the Tensor. It can be any type defined in \link cytnx::Type
+    cytnx::Type \endlink
+    @param device the device that the Tensor is put on. It can be any device defined in \link
+    cytnx::Device cytnx::Device \endlink
+
+    @return
+        [Tensor]
+
+    */
+  Tensor arange(const std::vector<cytnx_uint64> &shape, const cytnx_double &start,
```
kaihsin, line 162: Dont do this

include/Generator.hpp
```diff
@@ -132,11 +142,56 @@ namespace cytnx {
...
+     evenly spaced. Then reshape to the shape specified by the input shape.
+      @param start the start value of the range
+      @param end the end value of the range
+      @param Nelem the number of elements
+      @param endpoint if true, the end value is included in the range.
+      @param dtype the dtype of the Tensor. It can be any type defined in \link cytnx::Type \endlink
+      @param device the device that the Tensor is put on. It can be any device defined in \link
+      cytnx::Device \endlink
+    */
+  Tensor linspace(const std::vector<cytnx_uint64> &shape, const cytnx_double &start,
```
kaihsin, line 192: remove

include/UniTensor.hpp
```diff
@@ -4165,6 +4165,26 @@ namespace cytnx {
...
+    @param[in] Nelem the number of elements.
+    @param[in] in_labels the labels of the UniTensor.
+    @param[in] name the name of the UniTensor.
+    @return
+        [UniTensor]
+    @see arange(const std::vector<cytnx_uint64> &shape, const cytnx_double &start, const
+    cytnx_double &end, const cytnx_double &step, const unsigned int &dtype, const int &device) \n
+    @see arange(const cytnx_int64 &Nelem)
+    */
+    static UniTensor arange(const std::vector<cytnx_uint64> &shape,
```
kaihsin, line 4182: same as above

include/UniTensor.hpp
```diff
@@ -4189,6 +4209,32 @@ namespace cytnx {
...
+    @param[in] in_labels the labels of the UniTensor.
+    @param[in] dtype the data type of the UniTensor, see cytnx::Type for more information.
+    @param[in] device the device type of the UniTensor, see cytnx::Device for more information.
+    @param[in] name the name of the UniTensor.
+    @return
+        [UniTensor]
+    @see arange(const std::vector<cytnx_uint64> &shape, const cytnx_double &start, const
+    cytnx_double &end, const cytnx_double &step, const unsigned int &dtype, const int &device)
+    */
+    static UniTensor arange(const std::vector<cytnx_uint64> &shape, const cytnx_double &start,
```
kaihsin, line 4229: same as above

include/UniTensor.hpp
```diff
@@ -4217,6 +4263,35 @@ namespace cytnx {
...
+    @param[in] in_labels the labels of the UniTensor.
+    @param[in] dtype the data type of the UniTensor, see cytnx::Type for more information.
+    @param[in] device the device type of the UniTensor, see cytnx::Device for more information.
+    @param[in] name the name of the UniTensor.
+    @return
+        [UniTensor]
+    @see linspace(const cytnx_double &start, const cytnx_double &end, const cytnx_uint64 &Nelem,
+    const bool &endpoint, const unsigned int &dtype, const int &device)
+    */
+    static UniTensor linspace(const std::vector<cytnx_uint64> &shape, const cytnx_double &start,
```
kaihsin, line 4285: same above

pybind/generator_py.cpp
kaihsin, line 1: same adjustment as c++ side

pybind/unitensor_py.cpp
kaihsin, line 1: same adjustment as c++ side

src/Generator.cpp
kaihsin, line 1: please remove all the things with arange with shape as list

tests/DenseUniTensor_test.cpp
kaihsin, line 1: testing case accrodingly

kaihsin (changes requested):

kaihsin:
@hunghaoti , the user can always put unitensor with different type, you just need to cast it to the proper type. I think you code in C++ and Python enough to know it should be intrinsic casting for UX

jeffry1829:
@kaihsin You mean we only keep `arange(list)`, and remove all other added function?

include/UniTensor.hpp
jeffry1829: I think this should be kept?
We should remove arange(shape, start,end,...) ?

kaihsin: you are combining arange and reshape at the same API call. numpy does not have this kind of APIs, and its inconsistent

Issue #328, opened 2023-10-26
random seed reset each time a tensor is created randomly


manuschneider:
If I create a random tensor, it always has the same entries:
```
T = cytnx.random.uniform([4,4], low=-1., high=1.)
print(T)
T2 = cytnx.random.uniform([4,4], low=-1., high=1.)
print(T2)
```
gives the same tensors `T` and `T2`. This is not what I expect (and different from the behavior of other libraries that I used). It seems a standard seed is set if no `seed` argument is given, which is always the same. Isn't there a standard seed that is evolved each time I generate random numbers? I think this would be a better default initialization.

kaihsin:
The initializer use random_device() which will create new instance everytime.  The simple fix is to use rand() as default on C++ side

kaihsin:
or srand()

kaihsin:
the design philosophy is that the API call should have seed, which the user can assign. if not assign, it should get seed from pseudo random gen engine.

jeffry1829:
I cannot think of any way to fix this quickly.
I'll soon make an PR that overloads 2 methods for every function that has parameter seed. One with seed, one without.

jeffry1829:
I tried replacing default seed as `static` random_device()(), which should generate different number everytime we call the function.

I tested it in C++, and it works fine. I guess the problem is that pybind _might_ not support dynamical default function parameter.

PR #332, opened 2023-10-26
Fix random seed for generator function

jeffry1829:
#328 

This draft works for c++ now. Not working for pybind.


jeffry1829:
Actually, before this PR, c++ part works fine.
I think the problem is all on pybind side.

j9263178 (approved):

PR #347, opened 2023-11-01
Added tests for BlockUniTensor and Bond

yingjerkao:
Bump up test coverage

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -417,3 +417,8 @@ if(BUILD_PACKAGES)
   include("Package.cmake")
   include(CPack)
 endif()
+
+if(RUN_TESTS)
+  include(CTest)
+endif()
+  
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -417,3 +417,8 @@ if(BUILD_PACKAGES)
   include("Package.cmake")
   include(CPack)
 endif()
+
```
jeffry1829, line 420: This might not be neccessary?

IvanaGyro: Do you mean `include(CTest)` is not neccessary?

jeffry1829: But current cmakelist already contains include(CTest)?

IvanaGyro: It should be there: https://github.com/Cytnx-dev/Cytnx/blob/f43dcbd960b955611cddcbc036cc07e113dbbb8b/CMakeLists.txt#L144
I don't know why it doesn't show in the comparison tab.

IvanaGyro (approved):

PR #368, opened 2023-11-14
fix arithmetic bug

jeffry1829:


src/linalg/Add.cpp
```diff
@@ -613,6 +613,9 @@ namespace cytnx {
 
     cytnx::UniTensor Add(const cytnx::UniTensor &Lt, const cytnx::UniTensor &Rt) {
       UniTensor out = Lt.clone();
+      if (Lt.dtype() > Rt.dtype()) {
+        out = out.astype(Rt.dtype());
+      }
       out.set_labels(vec_range<std::string>(Lt.rank()));
       out.set_name("");
```
kaihsin, line 621:
```c++
UniTensor out;
if(Lt.dtype() > Rt.dtype()){
     out = Rt.clone();
     out.add_(Lt);
}else{
     out = Lt.clone();
     out.add_(Rt);
}

```

kaihsin: and then dont forget to set label and name before `return out`

src/linalg/Mul.cpp
kaihsin, line 1: Suggest modify as Add.cpp

kaihsin (changes requested):
make corresponding change for Add and Mul

PR #372, opened 2023-11-16
Eig,Eigh for Block UniTensor

j9263178:


pcchen:
Can someone fix the failed tests and then approve/merger this pull request. I have tested the basic function and it works ok.

PR #400, opened 2023-12-03
Add arg tol for dense to block unitensor conversion

j9263178:


yingjerkao:
This PR is from dev-master. 
We probably need a new PR from master if we want this feature.

Issue #404, opened 2023-12-18
Lanczos approixmation for matrix exponentiation


TensorNetwork18:
Dear Cytnx developers,

It seems to me that the library so far only implemented Lanczos for ground-states.
I would like to request a Lanczos routine to approximate exp(A)*v.
Please let me know if it is possible to implement this.

Thanks,
Jheng-Wei Li

ianmccul:
`Lanczos_Exp` now exists, but the recent fixes and follow-up issues suggest that the implementation needs further numerical review.

Relevant follow-ups:

- #900 audits the other non-ARPACK Krylov algorithms for the same style of numerical issues that were fixed recently in `Lanczos_Exp`.
- #901 tracks the projected exponential call in `Lanczos_Exp`; it currently premultiplies the Hermitian projected matrix by `tau` and calls `ExpM`, instead of preserving the Hermitian structure and passing the scalar separately.
- #902 tracks the general robustness of `ExpM`/`ExpH`.

Opened by Codex on behalf of @ianmccul.

PR #407, opened 2024-02-21
Add function:Lanczos-like algorithm to approximate exponetial operator.

hunghaoti:
Add function:Lanczos-like algorithm to approximate exponetial operator performing on the vector. The algorithm is based on the reference:
https://doi.org/10.48550/arXiv.1111.1491
The corresponding issue is #404

### DIFF
--- a/src/linalg/Lanczos_Exp.cpp
+++ b/src/linalg/Lanczos_Exp.cpp
@@ -178,22 +178,22 @@ namespace cytnx {
       out.set_rowrank_(v0.rowrank());
     }
 
-    // Lanczos_Exp_Ut
-    UniTensor Lanczos_Exp_Ut(LinOp *Hop, const UniTensor &Tin, const double &CvgCrit,
-                             const unsigned int &Maxiter, const bool &verbose) {
+    // Lanczos_Exp
+    UniTensor Lanczos_Exp(LinOp *Hop, const UniTensor &Tin, const double &CvgCrit,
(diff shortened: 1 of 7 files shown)


src/linalg/Lanczos_Exp.cpp
```diff
@@ -0,0 +1,227 @@
...
+#ifdef BACKEND_TORCH
+#else
+
+namespace cytnx {
+  namespace linalg {
+    typedef Accessor ac;
+    using namespace std;
+
+    // <A|B>
+    static Scalar Dot(const UniTensor &A, const UniTensor &B) {
```
kaihsin, line 23: do we have dot?

src/linalg/Lanczos_Exp.cpp
```diff
@@ -0,0 +1,227 @@
...
+      };
+      auto r = (b - I_plus_A_Op(Tin)).relabels_(b.labels());
+      auto r0 = r;
+      auto x = Tin;
+      auto p = Dot(r0, r);
+      auto pv = r;
+      auto p_old = p;
+      auto pv_old = pv;
+      auto x_old = x;
+      auto r_old = r;
```
kaihsin, line 43: Why is this so scattered? maybe we should considered grouping parameters into a struct

kaihsin (commented):
lgtm, just maybe put some more comments in your code, also try to use struct to organize things especially when doing low level routines

PR #413, opened 2024-02-24
Haoti/arnoldi ut

hunghaoti:
Add function Arnoldi support for Dense UniTensor.


src/linalg/Arnoldi.cpp
```diff
@@ -172,9 +205,80 @@ namespace cytnx {
...
+        buffer.push_back(nextTens);
+        for (cytnx_uint32 j = 0; j < krydim; j++) {
+          auto h = Dot(buffer[i], buffer[j]).conj();
+          kry_mat_buffer[{i - 1, j}] = h;
+          buffer[i] -= h * buffer[j];
+        }
+        auto h = std::sqrt(static_cast<double>(Dot(buffer[i], buffer[i]).real()));
+        kry_mat_buffer[{i - 1, i}] = h;
+        buffer[i] /= h;
+        Tensor kry_mat = ResizeMat(kry_mat_buffer, krydim, krydim);
```
kaihsin, line 241: Are you sure we don't have this ResizeMat functionality implemented somewhere?

kaihsin (commented):
I would suggest for any internal functions that you don't want to expose, using underscore and lowercase with underscore separators for naming it, instead of Capitalize separator. for example:


`GetEig` -> `_get_eigenval` 


Other than that LGTM

src/linalg/Arnoldi.cpp
hunghaoti: I tried to find it but I could not find. Maybe we could provide this function in the future.

src/linalg/Arnoldi.cpp
```diff
@@ -172,9 +205,80 @@ namespace cytnx {
...
+        dim *= UT_init_shape[i];
+      }
+      const cytnx_uint64 imp_maxiter = std::min(maxiter, Hop->nx() + 1);
+      const cytnx_complex128 unit_complex = 1.0;
+      auto eigvals = std::vector<Scalar>(k, Scalar());
+      std::vector<UniTensor> eigTens_s;
+      Tensor kry_mat_buffer =
+        cytnx::zeros({imp_maxiter + 1, imp_maxiter + 1}, Hop->dtype(), Hop->device());
+      bool is_cvg = false;
+      auto eigvals_old = std::vector<Scalar>(k, Scalar::maxval(Type.Double));
```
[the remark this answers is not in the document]
hunghaoti: I see. How about use a rank-1 Tensor to store the eigenvalues? Actually, I think the length of this vector is only k which is usually be 1. But if there is more efficient way then I will modify it.

hunghaoti:
> I would suggest for any internal functions that you don't want to expose, using underscore and lowercase with underscore separators for naming it, instead of Capitalize separator. for example:
> 
> `GetEig` -> `_get_eigenval`
> 
> Other than that LGTM

OK, I will modify it.

src/linalg/Arnoldi.cpp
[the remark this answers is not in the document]
kaihsin: ok lets keep this for now, but you might need to do performance check for this with Lanczos

src/linalg/Arnoldi.cpp
kaihsin: create another issue for this

hunghaoti: OK, see issue #415 .

PR #437, opened 2024-05-23
Fix tdvp example and add Lanczos benchmark:

hunghaoti:
1. Fix the tdvp example code to correct order.
2. Add benchmark test for Lanczos_Gnd and Lanczos_Exp.


hunghaoti:
Optimize Lanczos method by storing the Krylov subspace vectors in the first iteration.

src/linalg/Lanczos_Gnd_Ut.cpp
```diff
@@ -166,44 +132,21 @@ namespace cytnx {
...
       if (cvg_fin == false) {
-        cytnx_warning_msg(true,
-                          "[WARNING] iteration not converge after Maxiter!.\n :: Note :: ignore if "
-                          "this is intended%s",
-                          "\n");
+        // cytnx_warning_msg(true,
+        //                   "[WARNING] iteration not converge after Maxiter!.\n :: Note :: ignore
+        //                   if " "this is intended%s",
+        //                   "\n");
       }
```
kaihsin, line 139: should we just removing this part completely?

kaihsin (changes requested):

src/linalg/Lanczos_Gnd_Ut.cpp
hunghaoti: How about we print this info out if **verbose==ture**?

hunghaoti: I have remain the warning. Could you review again, thanks.

hunghaoti: I have remain the warning. Could you review again, thanks.
https://github.com/Cytnx-dev/Cytnx/blob/47984fc8f61358a994a362fbf39375bd2cf9aa11/src/linalg/Lanczos_Gnd_Ut.cpp#L134-L139

kaihsin (approved):

Issue #439, opened 2024-05-29
Transition to cuTENSOR 2.x


j9263178:
https://docs.nvidia.com/cuda/cutensor/latest/api_transition.html

There are some API changes in the new version so the current source code cannot be compiled with cuTENSOR 2.x.

IvanaGyro:
I have finished this job that migrates cuTENSOR to 2.x. I will create a PR after getting the permission to create a new branch on this repo.

IvanaGyro:
Done with #447

PR #441, opened 2024-06-07
Transition to cutensor2.x

j9263178:
https://github.com/Cytnx-dev/Cytnx/issues/439


IvanaGyro:
This is finished in #447

Issue #440, opened 2024-06-02
Wrong Optimized Network Contraction Order


Chan-Sing-Hong:
I have encountered a problem when setting the optimized network contraction order using Cytnx version 0.9.7. The attachment is a demo code demonstrating the issue: [show_problem.zip](https://github.com/user-attachments/files/15525686/show_problem.zip)

The T tensor and its dagger are rank-4 tensors with dimension _chi_ for each bond:
````
               0                           2
         ┌─────┘                           └────┐
         │ ┏━━━╳━━┓                    ┏━━━╳━━┓ │
         └─┨      ┃                    ┃      ┠─┘
      1 ───┨   T  ┠─── 2          0 ───┨  T.d ┠─── 3
           ┃      ┠─┐                ┌─┨      ┃
           ┗━━━━━━┛ │                │ ┗━━━━━━┛
               ┌────┘                └─────┐
               3                           1
````
Below, I use the network:
````
              ┌──────101───────┐
              │                │
        ┌─────┘                └────┐      
        │ ┏━━━╳━━┓         ┏━━━╳━━┓ │
        └─┨      ┃         ┃      ┠─┘
     0 ───┨  Tu  ┠───102───┨ Tu.d ┠───2
          ┃      ┠─┐     ┌─┨      ┃                    ┏━━━╳━━┓ 
          ┗━━━━━━┛ │     │ ┗━━━━━━┛               0 ───┨      ┠─── 2
        ┌────103───┘     └────104───┐      -->         ┃ TOUT ┃
        │ ┏━━━╳━━┓         ┏━━━╳━━┓ │             1 ───┨      ┠─── 3 
        └─┨      ┃         ┃      ┠─┘                  ┗━━━━━━┛
     1────┨  Td  ┠───105───┨ Td.d ┠────3
          ┃      ┠─┐     ┌─┨      ┃
          ┗━━━━━━┛ │     │ ┗━━━━━━┛
              ┌────┘     └─────┐
              │                │
              └──────106───────┘
````
or
````
net = cytnx.Network()
net.FromString(["Tu  : 101,0,102,103",
                "Td  : 103,1,105,106",
                "Tu.d: 102,104,101,2",
                "Td.d: 105,106,104,3",
                "TOUT: 0, 1; 2, 3"])
````

For this network, the optimized contraction order should be ((Tu, Tu.d), (Td, Td.d)), with
[...]

PR #458, opened 2024-08-31
Re-fix setOrder

jeffry1829:
Previous fix on this problem cause GPU test to fail.

This time the RegularNetwork::CtTree_to_eisumpath function is fixed.

This is related to #440

### DIFF
--- a/src/RegularNetwork.cpp
+++ b/src/RegularNetwork.cpp
@@ -261,7 +261,6 @@ namespace cytnx {
             tns.erase(tns.begin() + id2);
           else
             tns.erase(tns.begin() + id2 - 1);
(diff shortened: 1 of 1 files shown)


src/RegularNetwork.cpp
```diff
@@ -261,7 +261,9 @@ namespace cytnx {
             tns.erase(tns.begin() + id2);
           else
             tns.erase(tns.begin() + id2 - 1);
-          tns.push_back(root->name);
+          // tns.push_back(root->name);
```
IvanaGyro, line 264: We can delete this line. Git will record the history.

IvanaGyro (approved):

Issue #449, opened 2024-08-27
Problem with installing on Mac


josephinius:
First, I had to add `add_compile_options(-Wno-c++11-narrowing)` into CMakeLists.txt to get rid of this error:

```
error: non-constant-expression cannot be narrowed from type 'value_type' (aka 'int') to 'unsigned long long' in initializer list [-Wc++11-narrowing]
        auto eigTens = P_inv.at({0, maxIdx}) * qs[0];
```

But I still have the following error while installing: 
```
error: implicit conversion from '_Complex double' to 'double' is not permitted in C++
        od[0] *= (cytnx_complex128)_Rin[i * N + i];
```

I use clang as a compiler.

josephinius:
Here are the whole error messages:

```
/Cytnx/src/backend/linalg_internal_cpu/Det_internal.cpp:31:36: error: implicit conversion from '_Complex double' to 'double' is not permitted in C++
        od[0] *= (cytnx_complex128)_Rin[i * N + i];
                 ~                 ^~~~~~~~~~~~~~~
/Cytnx/src/backend/linalg_internal_cpu/Det_internal.cpp:55:35: error: implicit conversion from '_Complex float' to 'float' is not permitted in C++
        od[0] *= (cytnx_complex64)_Rin[i * N + i];
```

IvanaGyro:
#455 fixed the `no-c++11-narrowing` problem. For the implicit conversion, I can't reproduce the error while using clang++. @josephinius, does the error still happen?

pcchen:
with commit 0be640dcb08dcfd4de8ed9a2f7da73dec7ddfaf2 
and clang version:
-- The CXX compiler identification is AppleClang 15.0.0.15000309
-- The C compiler identification is AppleClang 15.0.0.15000309

I still encounter this
````
[ 83%] Building CXX object CMakeFiles/cytnx.dir/src/backend/linalg_internal_cpu/Ger_internal.cpp.o
/Users/pcchen/github/Cytnx/src/backend/linalg_internal_cpu/Det_internal.cpp:31:36: error: implicit conversion from '_Complex double' to 'double' is not permitted in C++
        od[0] *= (cytnx_complex128)_Rin[i * N + i];
                 ~                 ^~~~~~~~~~~~~~~
/Users/pcchen/github/Cytnx/src/backend/linalg_internal_cpu/Det_internal.cpp:55:35: error: implicit conversion from '_Complex float' to 'float' is not permitted in C++
[...]

PR #461, opened 2024-09-01
Suppress clang error

IvanaGyro:
This PR eases the problem of #449.

### DIFF
--- a/src/backend/linalg_internal_cpu/Det_internal.cpp
+++ b/src/backend/linalg_internal_cpu/Det_internal.cpp
@@ -1,7 +1,12 @@
 #include "Det_internal.hpp"
-#include "../utils_internal_interface.hpp"
-#include "utils/utils.hpp"
+
+#include <complex.h>
+#include <type_traits>
+
 #include "backend/lapack_wrapper.hpp"
+#include "backend/utils_internal_interface.hpp"
+#include "utils/utils.hpp"
+
 #ifdef UNI_OMP
   #include <omp.h>
 #endif
@@ -25,10 +30,22 @@ namespace cytnx {
       cytnx_error_msg(
         info != 0, "%s %d",
         "[ERROR][Det_internal] Error in Lapack function 'zgetrf': Lapack INFO = ", info);
+
+      // Whether lapack_complex_T is defined as std::complex<T> (c++ complex) or T _Complex
+      // (C complex) depends on whether MKL is used.
+      auto lapack_complex_to_cytnx_complex = [](auto x) {
+        if constexpr (std::is_same_v<std::complex<double>, decltype(x)>) {
+          return static_cast<cytnx_complex128>(x);
+        } else if constexpr (std::is_same_v<double _Complex, decltype(x)>) {
+          return cytnx_complex128{creal(x), cimag(x)};
+        } else {
+          assert(false);
+        }
+      };
       od[0] = 1;
       bool neg = 0;
       for (lapack_int i = 0; i < N; i++) {
-        od[0] *= (cytnx_complex128)_Rin[i * N + i];
+        od[0] *= lapack_complex_to_cytnx_complex(_Rin[i * N + i]);
         if (ipiv[i] != (i + 1)) neg = !neg;
       }
       free(ipiv);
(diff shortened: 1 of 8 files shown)


src/backend/linalg_internal_cpu/Det_internal.cpp
```diff
@@ -27,8 +31,8 @@ namespace cytnx {
         "[ERROR][Det_internal] Error in Lapack function 'zgetrf': Lapack INFO = ", info);
       od[0] = 1;
       bool neg = 0;
-      for (lapack_int i = 0; i < N; i++) {
-        od[0] *= (cytnx_complex128)_Rin[i * N + i];
+      for (int i = 0; i < N; i++) {
+        od[0] *= cytnx_complex128(_Rin[i * N + i]);
```
kaihsin, line 35: please make sure this casting will work if the element is not complex

src/backend/linalg_internal_cpu/Det_internal.cpp
```diff
@@ -27,8 +31,8 @@ namespace cytnx {
         "[ERROR][Det_internal] Error in Lapack function 'zgetrf': Lapack INFO = ", info);
       od[0] = 1;
       bool neg = 0;
-      for (lapack_int i = 0; i < N; i++) {
-        od[0] *= (cytnx_complex128)_Rin[i * N + i];
+      for (int i = 0; i < N; i++) {
```
kaihsin, line 34: the reason for having  lapack_int because of backend agnostic issue.  Here is to make it compatible to when the lapack/mkl is using 64bit backend/api v.s. 32bit. by replacing it with int you are actually limited the scope

tests/gpu/BlockUniTensor_test.h
kaihsin, line 1: please use the cytnx type here

tests/gpu/linalg_test/Directsum_test.cpp
kaihsin, line 1: use cytnx type instead.

kaihsin (dismissed):

kaihsin:
Actually maybe I miss something but what exactly is the issue?

kaihsin:
```
[ 83%] Building CXX object CMakeFiles/cytnx.dir/src/backend/linalg_internal_cpu/Ger_internal.cpp.o
/Users/pcchen/github/Cytnx/src/backend/linalg_internal_cpu/Det_internal.cpp:31:36: error: implicit conversion from '_Complex double' to 'double' is not permitted in C++
        od[0] *= (cytnx_complex128)_Rin[i * N + i];
```

It seems the issue is the implicit cast upon calling `*=` where its going from complex to double? (specifically od[0] is double and someone try to do this op?) Then I am not sure the fix in this OR goes into the root of problem? @IvanaGyro maybe you can put some more insight on what happends?

IvanaGyro:
> ```
> [ 83%] Building CXX object CMakeFiles/cytnx.dir/src/backend/linalg_internal_cpu/Ger_internal.cpp.o
> /Users/pcchen/github/Cytnx/src/backend/linalg_internal_cpu/Det_internal.cpp:31:36: error: implicit conversion from '_Complex double' to 'double' is not permitted in C++
>         od[0] *= (cytnx_complex128)_Rin[i * N + i];
> ```
> 
> It seems the issue is the implicit cast upon calling `*=` where its going from complex to double? (specifically od[0] is double and someone try to do this op?) Then I am not sure the fix in this OR goes into the root of problem? @IvanaGyro maybe you can put some more insight on what happends?

Sorry, I don't have a Mac device. This original problem of implicit conversion only happens on Mac. Here is the commend I left in the issue. https://github.com/Cytnx-dev/Cytnx/issues/449#issuecomment-2323257106

src/backend/linalg_internal_cpu/Det_internal.cpp
IvanaGyro: Whether `lapack_complex_double` is defined as `std::complex<double>`(c++ complex) or `double _Complex`(C complex) depends on whether MKL is used.  The spec of C++17 doesn't promise to construct `std::complex<double>` from `double _Complex`, but it is allowed in the implementation of GCC and clang.

src/backend/linalg_internal_cpu/Det_internal.cpp
IvanaGyro: Updated

kaihsin:
Yes. That's what I was asking for agnostic implementation. It needs to take
all those deps into consideration. And previous solution is to use macro (
actually idk if there is any other solutions even exist in C/C++) to deal
with this

pcchen:
I would like to suggest to keep this kind of technical note somewhere in the repository.

src/backend/linalg_internal_cpu/Det_internal.cpp
IvanaGyro: A safe solution is updated.

tests/gpu/linalg_test/Directsum_test.cpp
IvanaGyro: Done.

IvanaGyro:
> I would like to suggest to keep this kind of technical note somewhere in the repository.

Mentioned in the comment in the latest update.

kaihsin (approved):

PR #464, opened 2024-09-10
fixed segfault when a BlockUniTensor with no valid blocks was printed…

manuschneider:
Fixed a segfault in case no valid block exists. For example, the following code resulted in a kernel crash:
```
bond_d = cytnx.Bond(
     cytnx.BD_IN, [cytnx.Qs(5)>>1, cytnx.Qs(-5)>>1],
     [cytnx.Symmetry.U1()]
)
bond_e = cytnx.Bond(
     cytnx.BD_IN, [cytnx.Qs(1)>>1, cytnx.Qs(-1)>>1],
     [cytnx.Symmetry.U1()]
)
bond_f = cytnx.Bond(
     cytnx.BD_OUT,
     [cytnx.Qs(2)>>1, cytnx.Qs(0)>>2, cytnx.Qs(-2)>>1],
     [cytnx.Symmetry.U1()]
)

Tsymm = cytnx.UniTensor([bond_d, bond_e, bond_f], name="symm. tensor",
labels=["d", "e", "f"])

Tsymm.print_diagram() 
```

### DIFF
--- a/include/UniTensor.hpp
+++ b/include/UniTensor.hpp
@@ -1167,34 +1167,36 @@ namespace cytnx {
     };
 
     unsigned int dtype() const {
+        //[21 Aug 2024] This is a copy from BlockUniTensor;
   #ifdef UNI_DEBUG
       cytnx_error_msg(this->_blocks.size() == 0, "[ERROR][internal] empty blocks for blockform.%s",
                       "\n");
   #endif
-      return this->_blocks.size() < 1 ? N_Type + 1 : this->_blocks[0].dtype();
+      return this->_blocks.size() < 1 ? Type.Void : this->_blocks[0].dtype();
     };
     int device() const {
+        //[21 Aug 2024] This is a copy from BlockUniTensor;
   #ifdef UNI_DEBUG
       cytnx_error_msg(this->_blocks.size() == 0, "[ERROR][internal] empty blocks for blockform.%s",
                       "\n");
   #endif
-      return this->_blocks.size() < 1 ? 404 : this->_blocks[0].device();
+      return this->_blocks.size() < 1 ? -404 : this->_blocks[0].device();
     };
     std::string dtype_str() const {
+        //[21 Aug 2024] This is a copy from BlockUniTensor;
   #ifdef UNI_DEBUG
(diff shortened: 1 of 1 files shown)


include/UniTensor.hpp
```diff
@@ -1171,28 +1171,30 @@ namespace cytnx {
...
-      return this->_blocks[0].dtype();
+      return this->_blocks.size() < 1 ? N_Type + 1 : this->_blocks[0].dtype();
     };
     int device() const {
   #ifdef UNI_DEBUG
       cytnx_error_msg(this->_blocks.size() == 0, "[ERROR][internal] empty blocks for blockform.%s",
                       "\n");
   #endif
-      return this->_blocks[0].device();
+      return this->_blocks.size() < 1 ? 404 : this->_blocks[0].device();
```
IvanaGyro, line 1181: Do we assume the device value larger and equal to 0 is a GPU device?

include/UniTensor.hpp
```diff
@@ -1171,28 +1171,30 @@ namespace cytnx {
       cytnx_error_msg(this->_blocks.size() == 0, "[ERROR][internal] empty blocks for blockform.%s",
                       "\n");
   #endif
-      return this->_blocks[0].dtype();
+      return this->_blocks.size() < 1 ? N_Type + 1 : this->_blocks[0].dtype();
```
IvanaGyro, line 1174: Is `Type.Void` a choice here?

ianmccul:
It worries me that it is doing something completely different in UNI_DEBUG mode (displaying an error, and throwing an exception) versus non-debug mode (return a normal value).

include/UniTensor.hpp
kaihsin: Yes. See Device.hpp

kaihsin:
> It worries me that it is doing something completely different in UNI_DEBUG mode (displaying an error, and throwing an exception) versus non-debug mode (return a normal value).

It is legacy from uni10 where we have two builds for test and performances. I don't think it's necessary anymore.

manuschneider:
I updated the standard values, thanks @IvanaGyro for checking.

IvanaGyro (approved):

PR #471, opened 2024-09-14
Fix cutt install and (Ge)Svd(_truncate) error when use cuda

jeffry1829:


CytnxBKNDCMakeLists.cmake
```diff
@@ -154,7 +154,8 @@ if(USE_CUDA)
         include_directories(${install_dir}/include)
         message(STATUS "cutt install dir: ${install_dir}")
         add_dependencies(cytnx cutt)
-        set_property(TARGET cytnx PROPERTY CUDA_ARCHITECTURES 52 53 60 61 62 70 72 75 80 86)
+        # set_property(TARGET cytnx PROPERTY CUDA_ARCHITECTURES 52 53 60 61 62 70 72 75 80 86)
```
IvanaGyro, line 157: Forget to remove this?

IvanaGyro (approved):
Good job! It would be great if you could `fixup` two commits for formatting into their previous commit, and fix the format of the first commit. Currently, it is really hard to know that you only change the device variable passing to `Init()`.

Issue #380, opened 2023-11-23
permute_ and reshape_ return none


hunghaoti:
Function 'permute_' and 'reshape_' return None. For example:
```python
A = cytnx.ones([2,3])
B = A.permute_(1,0)
print(B) #None
```
```python
A = cytnx.UniTensor.ones([2,3])
B = A.permute_([1,0])
print(B) #None
```

pcchen:
I think the in-place version returns `None` is proper?

IvanaGyro:
The behaviour was changed after #478.

PR #478, opened 2024-09-20
Modify reshape_ and permute_ to return itself.

hunghaoti:
I modified reshape_, permute_ functions in both UniTensor and Tensor object. So now they will return itself, just like relabel_(), instead of None as a void function.
I have test the following code and work well:
```python
A = cytnx.UniTensor.zeros([2,3])
B = A.permute([1,0]) # A and B share tensor elements, different metadata
B[0,1] = 1.
C = A.clone().permute([1,0]) # C is independent of A
C[0,1] = 2.
print(A) # Original shape, element A[1,0] == 1.
print(B) # Permuted shape, element B[0,1] == 1.
print(C) # Permuted shape, element C[0,1] == 2.
D = A.permute_([1,0])
# D is a reference to A
print(A) # Permuted shape, element A[0,1] == 1.
print(D is A) # Output: True

```
```python
A = cytnx.zeros([2,3])
A[0,1] = 1 
print(A)
D = A.permute_(1,0)
print(D)
print(D is A) # Output: True
```
`print(D is A)` gives True
This PR solve the issue #380

### DIFF
--- a/pybind/tensor_py.cpp
+++ b/pybind/tensor_py.cpp
@@ -184,7 +184,7 @@ void tensor_binding(py::module &m) {
          [](cytnx::Tensor &self, py::args args) {
            std::vector<cytnx::cytnx_uint64> c_args = args.cast<std::vector<cytnx::cytnx_uint64>>();
            // std::cout << c_args.size() << std::endl;
-           self.permute_(c_args);
+           return &self.permute_(c_args);
          })
     .def("permute",
          [](cytnx::Tensor &self, py::args args) -> cytnx::Tensor {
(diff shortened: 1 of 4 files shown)


pybind/tensor_py.cpp
```diff
@@ -184,7 +184,7 @@ void tensor_binding(py::module &m) {
          [](cytnx::Tensor &self, py::args args) {
            std::vector<cytnx::cytnx_uint64> c_args = args.cast<std::vector<cytnx::cytnx_uint64>>();
            // std::cout << c_args.size() << std::endl;
-           self.permute_(c_args);
+           return &self.permute_(c_args);
```
IvanaGyro, line 187: Should it be `return self.permute_(c_args);` instead?

IvanaGyro (approved):

Issue #482, opened 2024-09-26
Replacing magma_zgetrf_gpu by cusolverDn<Z><getrf>


yingjerkao:
Now that CuSolver has implemented LU decomposition 

https://github.com/NVIDIA/CUDALibrarySamples/tree/master/cuSOLVER/Xgetrf

I propose to replace `magma_zgetrf_gpu` in `cuDet_internal.cu` with `cusolverDn<Z><getrf>` 

https://docs.nvidia.com/cuda/cusolver/index.html#cusolverDnXgeqrf

With this update, we can drop the dependency on MAGMA.

IvanaGyro:
It seems that there is an entry for magma in the git submodule.
https://github.com/Cytnx-dev/Cytnx/blob/733d434521c2ae673703cfd14f3faf154b9967ca/.gitmodules#L11

yingjerkao:
I will clean things up in #536

PR #483, opened 2024-09-26
Remove Magma dependency by using CuSolver Routines

yingjerkao:
This PR attempts to remove the Magma dependency by replacing magma routines with the corresponding cusolver ones. 

It compiles CUDA 11.8, CuTensor 2.0.2.5, CuQuantum 24.08.0.5.


jeffry1829:
> It looks like the tests and coverage don't cover the gpu version?

Yes, one possible option is that we might need to link some machine to github action

jeffry1829:
I'll push test for Det later(along with some fixes), however, cusolver getrf is ridiculously slow, might investigate more

jeffry1829:
Nevermind, I didn't use cutensor and cuquantum

IvanaGyro:
@jeffry1829  Could you create another PR for https://github.com/Cytnx-dev/Cytnx/pull/483/commits/35c95aa2296d8a37afb56dab50dddbb38d04070d ? So this PR doesn't have to wait for the commit for fixing the Det.

jeffry1829:
> @jeffry1829 Could you create another PR for [35c95aa](https://github.com/Cytnx-dev/Cytnx/commit/35c95aa2296d8a37afb56dab50dddbb38d04070d) ? So this PR doesn't have to wait for the commit for fixing the Det.

Reverted

yingjerkao:
Fix #482

PR #490, opened 2024-10-06
Fix functionality of linalg::Sum() for complex data types

### LINKED ISSUE #467: `linalg::Sum(const Tensor&)` gives wrong answer for complex types
The test below should pass.

```cpp
#include "cuda_runtime_api.h"
#include "gtest/gtest.h"

#include "cytnx_error.hpp"
#include "linalg.hpp"
#include "Tensor.hpp"
#include "Type.hpp"

using namespace cytnx;

TEST(Linalg, Sum) {
  Tensor block(/* shape */ {5}, Type.ComplexDouble, Device.cuda, /* init_zero */ true);
  checkCudaErrors(cudaSetDevice(block.device()));
  block.fill(cytnx_complex128{2., 3.});
  Tensor sum_result = cytnx::linalg::Sum(block);
  checkCudaErrors(cudaDeviceSynchronize());

  EXPECT_EQ(sum_result.shape().size(), 1);
  EXPECT_EQ(sum_result.shape()[0], 1);
  EXPECT_DOUBLE_EQ(sum_result.at<cytnx_complex128>({0}).real(), 10.);
  EXPECT_DOUBLE_EQ(sum_result.at<cytnx_complex128>({0}).imag(), 15.);
}
```

However, the test fails.

```
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from Linalg
[ RUN      ] Linalg.Sum
test.cpp:20: Failure
Expected equality of these values:
  sum_result.at<cytnx_complex128>({0}).real()
    Which is: 0
  10.
    Which is: 10

test.cpp:21: Failure
Expected equality of these values:
  sum_result.at<cytnx_complex128>({0}).imag()
    Which is: 0
  15.
    Which is: 15

[  FAILED  ] Linalg.Sum (496 ms)
[----------] 1 test from Linalg (496 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (496 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] Linalg.Sum

 1 FAILED TEST


0% tests passed, 1 tests failed out of 1

Total Test time (real) =   4.98 sec

```

This issue may be the culprit for failing the tests `BlockUniTensorTest.gpu_Trace` and `DenseUniTensorTest.gpu_Trace`.

IvanaGyro:
This pull request addresses several issues related to the functionality of the `linalg::Sum()` function, particularly for complex data types on GPU, and resolves #467. The changes include:

1. **Fix for Complex Data Types**: 
   - Resolved issues with the `linalg::Sum()` function when handling complex data types on GPU. 
   - Added tests to validate the accuracy of summing floating-point numbers, as the previous algorithm for floating-point data types on CPU did not pass accuracy tests.

2. **Assertion for Number Equality**:
   - Introduced an assertion that checks the equality of numbers across all supported data types. This is particularly useful in type-parameterized tests.
   - The new assertion, `EXPECT_NUMBER_EQ`, compares floating-point numbers based on their ULP (Units in Last Place) distance, differentiating it from gtest's `EXPECT_EQ`.

3. **Code Cleanup**:
   - Removed an unused test file to streamline the codebase.
   - Moved the `TestTools` namespace under the `cytnx` namespace to avoid using directives in header files.
   - Removed an unused argument from the `linalg_internal::lii.Sum_ii` and `linalg_internal::lii.cuSum_ii` functions, as the output type is now determined by the input tensor.


IvanaGyro:
The algorithm of `linalg::Sum()` on CPU is not accurate enough to pass the new tests. We may want to apply the Kahan summation algorithm or to use `long double` if `long double` has more bits than `double` on the target machine.

manuschneider:
Why is the namespace `TestTools` avoided (but still existing), and testing functions are added to `Cytnx`? Maybe I have missed the discussion. Isn't it a bad practice to have `Cytnx` in the testing environment differently defined than a user would have who installed Cytnx?

Besides this, looks fine to me but maybe someone else should double check as well, especially about the GPU part.

IvanaGyro:
> Why is the namespace TestTools avoided (but still existing), and testing functions are added to Cytnx? Maybe I have missed the discussion. Isn't it a bad practice to have Cytnx in the testing environment differently defined than a user would have who installed Cytnx?

I agree that we don't need the nested `TestTools` namespace.

manuschneider:
Should the `TestTools` namespace be removed then? Could someone please have a look at the GPU part?

IvanaGyro:
> Should the TestTools namespace be removed then?

There are a few files using the `TestTools` namespace now. I will create another PR for removing that namespace after merging this PR.

manuschneider (approved):
`TestTools` namespace will be removed in another PR.

PR #510, opened 2024-11-13
Add a helper class to check if a type is `complex`

IvanaGyro:
Many refactoring works including #504, #508, and #509 use this change.

### DIFF
--- a/include/Type.hpp
+++ b/include/Type.hpp
@@ -1,14 +1,12 @@
-#ifndef _H_TYPE_
-#define _H_TYPE_
+#ifndef INCLUDE_TYPE_H_
+#define INCLUDE_TYPE_H_
 
-#include <string>
 #include <complex>
+#include <cstdint>
+#include <string>
+#include <type_traits>
 #include <vector>
-#include <stdint.h>
-#include <climits>
-#include <typeinfo>
-#include <unordered_map>
-#include <typeindex>
+
 #include "cytnx_error.hpp"
 
 #define MKL_Complex8 std::complex<float>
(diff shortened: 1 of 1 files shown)


manuschneider:
Why not call the function to use `is_complex` directly? So I would suggest
'is_complex_v` -> 'is_complex`
'is_complex` -> 'is_complex_type`
or something similar. This would affect #508 and #509 as well.

Generally, a short comment for functions and classes about what they do or how they are used is always welcome. Like
```
// is_complex checks if a data type is of type std::complex
// usage: is_complex<T> returns true or false for a data type T
```

ianmccul:
The terminology `is_complex<T>` for the structure and `is_complex_v<T>` matches the C++ convention of the [`type_traits`](https://en.cppreference.com/w/cpp/header/type_traits), eg `is_integral` -> `is_integral_v`, `is_pointer` -> `is_pointer_v` etc.  It is perhaps slightly unfortunate naming, but the structure metafunctions were added (and got the nice name) in C+11, before template variables were added in C++14, so we got the variables with the `_v` suffix.  For the same reason, metafunctions that return a type conventionally have a helper template alias that adds `_t` suffix.

The terminology @IvanaGyro has used also matches a [proposal to add `is_complex` and `is_complex_v` to the C++ standard library](https://github.com/cplusplus/papers/issues/805), and the [boost version](https://www.boost.org/doc/libs/1_86_0/libs/type_traits/doc/html/boost_typetraits/reference/is_complex.html).

manuschneider (approved):

PR #513, opened 2024-11-21
Add cmake configuration for the location of the test data

ianmccul:
Added CYTNX_TEST_DATA_DIR variable for the location of the test data root directory

Update all tests that reference the test data directory to load data from CYTNX_TEST_DATA_DIR

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -2,6 +2,9 @@
 if(BACKEND_TORCH)
 else()
 
+# Define the path to the test data directory
+set(CYTNX_TEST_DATA_DIR "${CMAKE_CURRENT_SOURCE_DIR}/test_data_base")
+
 add_executable(
   test_main
   test_tools.cpp
@@ -35,6 +38,9 @@ add_executable(
   algo_test/Vsplit_test.cpp
   algo_test/Vstack_test.cpp
 )
+
+# Add the preprocessor definition for CYTNX_TEST_DATA_DIR
+target_compile_definitions(test_main PUBLIC CYTNX_TEST_DATA_DIR="${CYTNX_TEST_DATA_DIR}")
 
 if(USE_CUDA)
       add_subdirectory(gpu)
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -5,6 +5,9 @@ else()
 # Define the path to the test data directory
 set(CYTNX_TEST_DATA_DIR "${CMAKE_CURRENT_SOURCE_DIR}/test_data_base")
 
+# Add the preprocessor definition for CYTNX_TEST_DATA_DIR
+add_compile_definitions(CYTNX_TEST_DATA_DIR="${CYTNX_TEST_DATA_DIR}")
+
 add_executable(
   test_main
   test_tools.cpp
@@ -38,9 +41,6 @@ add_executable(
   algo_test/Vsplit_test.cpp
   algo_test/Vstack_test.cpp
 )
-
-# Add the preprocessor definition for CYTNX_TEST_DATA_DIR
-target_compile_definitions(test_main PUBLIC CYTNX_TEST_DATA_DIR="${CYTNX_TEST_DATA_DIR}")
 
 if(USE_CUDA)
       add_subdirectory(gpu)
--- a/tests/BlockUniTensor_test.h
+++ b/tests/BlockUniTensor_test.h
@@ -9,7 +9,7 @@ using namespace cytnx;
 using namespace TestTools;
 class BlockUniTensorTest : public ::testing::Test {
  public:
-  std::string data_dir = "../../tests/test_data_base/common/BlockUniTensor/";
+  std::string data_dir = CYTNX_TEST_DATA_DIR "/common/BlockUniTensor/";
 
   Bond B1 = Bond(BD_IN, {Qs(0) >> 1, Qs(1) >> 2});
   Bond B2 = Bond(BD_IN, {Qs(0), Qs(1)}, {3, 4});
--- a/tests/DenseUniTensor_test.h
+++ b/tests/DenseUniTensor_test.h
@@ -10,7 +10,7 @@ using namespace std;
 using namespace TestTools;
 class DenseUniTensorTest : public ::testing::Test {
  public:
-  std::string data_dir = "../../tests/test_data_base/common/DenseUniTensor/";
+  std::string data_dir = CYTNX_TEST_DATA_DIR "/common/DenseUniTensor/";
 
   UniTensor ut_uninit;
   UniTensor utzero345;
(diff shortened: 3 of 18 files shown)


yingjerkao:
In the old uni10 days, the tests and related files are installed in a test subdirectory so that the `CTEST` can be run separately. This install location can be set in the main `CmakeLIst.txt`.

ianmccul:
I have no idea about the codecov test, it doesn't appear to be anything to do with the changes in the patch

tests/CMakeLists.txt
```diff
@@ -36,6 +39,9 @@ add_executable(
   algo_test/Vstack_test.cpp
 )
 
+# Add the preprocessor definition for CYTNX_TEST_DATA_DIR
+target_compile_definitions(test_main PUBLIC CYTNX_TEST_DATA_DIR="${CYTNX_TEST_DATA_DIR}")
```
IvanaGyro, line 43: Is it fine to set the visibility to `PRIVATE`?

Doesn't these lines also need to be added into `tests/gpu/CMakeLists.txt`?

IvanaGyro:
> In the old uni10 days, the tests and related files are installed in a test subdirectory so that the `CTEST` can be run separately. This install location can be set in the main `CmakeLIst.txt`.

What does `CTEST` can be run separately mean?

tests/CMakeLists.txt
ianmccul: I am not an expert on CMAKE (this is the first time I use it), but I gather here PUBLIC means that other targets that link with test_main will also get that symbol defined.  I guess there are no such targets, so PRIVATE is fine.

However, this definition is, in principle, not specific to `test_main`, but anything that gets compiled in this directory.  So would it be even better to use
```
add_compile_definitions(CYTNX_TEST_DATA_DIR="${CYTNX_TEST_DATA_DIR}")
```
?

I missed `tests/gpu/CMakeLists.txt` (I haven't tried compiling the gpu tests yet).  I can add it there too.

IvanaGyro: I tried. `add_compile_definitions()` doesn't propagate to the subdirectories. `add_compile_definitions` only affects the targets defined in the same cmake file, so it is the same with `target_compile_definitions` in this case. I feel that `target_compile_definitions` is safer because it will not cause a surprise result if someone adds a new target in the same camke file in the future.

ianmccul: Did you put `add_compile_definitions` at the right place?  According to the cmake documentation, it is a directory property, so it will apply to subdirectories too.  But it needs to be set before the `test_main` executable is added:
```CMakeLists.txt

if(BACKEND_TORCH)
else()

# Define the path to the test data directory
set(CYTNX_TEST_DATA_DIR "${CMAKE_CURRENT_SOURCE_DIR}/test_data_base")

# Add the preprocessor definition for CYTNX_TEST_DATA_DIR
add_compile_definitions(CYTNX_TEST_DATA_DIR="${CYTNX_TEST_DATA_DIR}")

add_executable(
  test_main
  test_tools.cpp
   ...
```

Works for me.

I think this is probably the best solution? If someone does add a new target then it is quite likely that it will want access to the test data directory.  If not, having a defined symbol `CYTNX_TEST_DATA_DIR` is not going to cause a problem.  And anyway, when they add the new target to `CMakeLists.txt`, they could choose to add it either before or after the `add_compile_definitions()` statement, to either get the symbol or not.

ianmccul: Ok, the tests compile and run on gpu.  I am getting a bunch of GPU test failures though,
```
[  FAILED  ] 40 tests, listed below:
[  FAILED  ] NetworkTest.gpu_Network_dense_no_order
[  FAILED  ] NetworkTest.gpu_Network_dense_find_optimal
[  FAILED  ] NetworkTest.gpu_Network_dense_order_line
[  FAILED  ] NetworkTest.gpu_Network_dense_specified_order
[  FAILED  ] NetworkTest.gpu_Network_dense_reuse
[  FAILED  ] NconTest.gpu_ncon_default_order
[  FAILED  ] NconTest.gpu_ncon_specified_order
[  FAILED  ] NconTest.gpu_ncon_optimal_order
[  FAILED  ] ContractTest.gpu_Contract_denseUt_optimal_order
[  FAILED  ] ContractTest.gpu_Contract_denseUt_default_order
[  FAILED  ] ContractTest.gpu_Contract_denseUt_specified_order
[  FAILED  ] ContractTest.gpu_Contract_denseUt_optimal_specified_order
[  FAILED  ] ContractTest.gpu_Contracts_denseUt_optimal_order
[  FAILED  ] ContractTest.gpu_Contracts_denseUt_default_order
[  FAILED  ] ContractTest.gpu_Contracts_denseUt_specified_order
[  FAILED  ] ContractTest.gpu_Contracts_denseUt_optimal_specified_order
[  FAILED  ] BlockUniTensorTest.gpu_Trace
[  FAILED  ] DenseUniTensorTest.gpu_Trace
[  FAILED  ] ExpM.gpu_ExpM_test
[  FAILED  ] Lanczos_Gnd.gpu_Bk_Lanczos_Gnd_test
[  FAILED  ] Arnoldi.gpu_which_LM_test
[  FAILED  ] Arnoldi.gpu_which_LR_test
[  FAILED  ] Arnoldi.gpu_which_LI_test
[  FAILED  ] Arnoldi.gpu_which_SM_test
[  FAILED  ] Arnoldi.gpu_which_SR_test
[  FAILED  ] Arnoldi.gpu_which_SI_test
[  FAILED  ] Arnoldi.gpu_mat_type_real_test
[  FAILED  ] Arnoldi.gpu_k1_test
[  FAILED  ] Arnoldi.gpu_k_max
[  FAILED  ] Arnoldi.gpu_smallest_dim
[  FAILED  ] Svd.gpu_dense_one_elem
[  FAILED  ] Svd.gpu_dense_nondiag_test
[  FAILED  ] Svd.gpu_U1_zeros_test
[  FAILED  ] Gesvd.gpu_dense_one_elem
[  FAILED  ] Gesvd.gpu_dense_nondiag_test
[  FAILED  ] linalg_Test.gpu_BkUt_Svd_truncate3
[  FAILED  ] linalg_Test.gpu_BkUt_Qr1
[  FAILED  ] linalg_Test.gpu_BkUt_expM
[  FAILED  ] linalg_Test.gpu_DenseUt_Gesvd_truncate
[  FAILED  ] linalg_Test.gpu_DenseUt_Svd_truncate
```

I looked a bit closer at the first one, error is coming from here: https://github.com/Cytnx-dev/Cytnx/blob/384bafaeee4bd99cf28b3999bf0e5ea72ac4ee26/src/UniTensor_base.cpp#L412
called from: https://github.com/Cytnx-dev/Cytnx/blob/384bafaeee4bd99cf28b3999bf0e5ea72ac4ee26/include/UniTensor.hpp#L3432

I didn't look any further.  Is this expected?

IvanaGyro: > And I think this should automatically get propagated to the gpu tests? I need to set up a cuda build environment to test it.

Aha, you are correct. `add_compile_definitions` should be put before `add_subdirectory`. I didn't notice that.

IvanaGyro: > I didn't look any further. Is this expected?

I think it's not expected.

IvanaGyro: I guess the tests failed because your base branch is `master`. Some bugs are fixed on the `dev-master` branch only.

IvanaGyro:
I merged this PR into the dev branch at #523.

PR #514, opened 2024-11-21
Refactor Type.hpp

ianmccul:
Refactor Type.hpp to automatically deduce properties of the types at compile time.

Type_class now contains only static members, and all except `Type_class::getname()` and `Type_class::c_typename_to_id()` are `constexpr`.

The global variable `Type` still exists, but it is now `constexpr`.  Calls to `Type.X` are now  equivalent to `Type_class::X`. But I do not suggest using `Type_class` directly, but instead consider removing the `Type` global variable and renaming `Type_class` to `Type`. It could also be a `namespace` rather than a `class`, which could be useful in some circumstances.

I wasn't able to make `Type_class::getname()` constexpr because it returns a `std::string`.  It *could* return a `const char*`, since all of the strings that it can return are string literals, but this causes a lot of compilation failures because of code like:

    cytnx_error_msg(dtype != Type.Float,
                    "[ERROR] type mismatch. try to get <float> type from raw data of type %s",
                    Type.getname(dtype).c_str());

where it is calling `getname().c_str()`, which obviously fails if `getname()` is returning a `const char*`.

The proper fix is to improve `cytnx_error_msg()` so that it can print values in a type-safe way.  After that is done, `Type_class::getname()` could be changed to constexpr and return a `const char*`.

I wasn't able to make `Type_class::c_typename_to_id()` constexpr because `typeid(T).name()` is not constexpr. That is a limitation of C++, and is unlikely to change.

A possible extension would be to add a wrapper around `typeid(T).name()` that returns `Type.getname()` if `T` is a cytnx type, with a fallback if it isn't a cytnx type.  It could also use some compiler-specific functions if available, for example both gcc and clang have `abi::__cxa_demangle()` function.  MSVC users are out of luck ;)

The second patch in the series  removes __type::__pybind_type.

The enumeration __type::__pybind_type is a duplicate of information already in Type_class, so we can remove it and make the Python bindings use the same enumeration as the C++ code.


ianmccul:
I am not planning to work further on this.  But the next step would probably be to refactor instances of arithmetic operations on `dype()` and magic numbers, and replace them with some functions to select the appropriate type for a common type of two (or more) parameters, type promotion etc.

ianmccul:
Reformatted using clang-format, also removed a comment that appears to be left over from a previous refactor

include/Type.hpp
```diff
@@ -63,125 +50,239 @@ namespace cytnx {
...
+    // type_size returns the sizeof(T) for the supported types. This is the same as
+    // sizeof(T), except that size_type<void> is 0.
+    template <typename T>
+    constexpr int type_size = sizeof(T);
+    template <>
+    constexpr int type_size<void> = 0;
+  }  // namespace internal
+
+  // the list of supported types. The dtype() of an object is an index into this list.
+  // This **MUST** match the ordering of __type::__pybind_type
```
IvanaGyro, line 113: Is `__type::__pybind_type` removed?

include/Type.hpp
```diff
@@ -63,125 +50,239 @@ namespace cytnx {
...
     bool is_complex;
     bool is_float;
     bool is_int;
     unsigned int typeSize;
   };
 
-  static const int N_Type = 12;
-  const int N_fType = 5;
+  template <typename T>
+  struct Type_struct_t {
```
IvanaGyro, line 189: Is is possible unpack the `Type_list` directly in `make_type_array_helper` and not to use this helper structure?

IvanaGyro (commented):
> But I do not suggest using Type_class directly, but instead consider removing the Type global variable and renaming Type_class to Type. It could also be a namespace rather than a class, which could be useful in some circumstances.

I also suggest renaming `Type_class`.

There are some new functions and types named in a similar "style" in the present files. Do you consider naming the new functions and types following the Google coding style? By doing so, we keep getting close to unifying the variable names.

include/Type.hpp
ianmccul: Yes, it could be replaced by some helper function, eg
```c++
    template <typename T>
    constexpr Type_struct make_type_struct() {
        return {
            Type_names<T>,
            Type_enum_name<T>,
            std::is_unsigned_v<T>,
            is_complex_v<T>,
            std::is_floating_point_v<T> || is_complex_floating_point_v<T>,
            std::is_integral_v<T> && !std::is_same_v<T, bool>,
            type_size<T>
        };
    }
    
    template <typename Tuple, std::size_t... Indices>
    constexpr auto make_type_array_helper(std::index_sequence<Indices...>) {
        return std::array<Type_struct, sizeof...(Indices)>{
            make_type_struct<std::tuple_element_t<Indices, Tuple>>()...};
    }
```

but I was thinking that Type_struct_t is actually an interesting class, and potentially more useful than Type_struct itself.  `is_float(unsigned int)` works for runtime calculations, but if you have some type T at compile time and you want to find that out, then a good way is `Type_struct_t<T>::is_float`. But that name isn't great, perhaps call it `Type_traits` instead?  Or simply `Traits` ? And then `Type_struct` can move into `internal` namespace, as it is an implementation detail of the `is_float(unsigned int)` etc family of functions.

I was also thinking to add `is_real`, meaning a floating point type but not complex.

IvanaGyro: I get your intention. However, at the compile time, the type is known, so we can directly use type traits template in STL to check the type. Using STL may make the code a little bit longer, but it is straightforward to people who hasn't checked `Type.hpp`.

For example, in case where needs `Type_struct_t<T>::is_float`, I will use `std::is_floating_point_v<T> || is_complex_floating_point_v<T>` or just `std::is_floating_point_v<T> || is_complex_v<T>` instead. BTW, the name of `is_float()` is ambiguous with the meaning of `std::is_same_v<T, float>`. If someone needs to get the display name of the type or the size of type, `Type_names<T>` and `sizeof(T)` satisfy the requirement.

ianmccul: What is easier to type, and less likely to go wrong, `Traits<T>::is_float`, or `std::is_floating_point_v<T> || is_complex_floating_point_v<T>` ?  And it could easily go wrong, for example your suggestion `std::is_floating_point_v<T> || is_complex_v<T>` will give the wrong answer if `T` is `std::complex<int>` (OK, that is not standard C++, and I've never used it, but I think some compilers allow it.)

The idea is that `Traits<T>::is_XXX` is guaranteed to map exactly same value as `Type.is_XXX(dtype)`, and with the same spelling.

For `is_float`, that is the name of the existing definition.  I have no objection to change it, but it is used elsewhere:
```
grep -r 'Type.is_float('
src/linalg/Lanczos_Exp.cpp:      cytnx_error_msg(!Type.is_float(Hop->dtype()),
src/linalg/Lanczos_ER.cpp:        !Type.is_float(Hop->dtype()),
src/linalg/Arnoldi.cpp:        !Type.is_float(Hop->dtype()),
src/linalg/Arnoldi.cpp:        !Type.is_float(Hop->dtype()),
src/linalg/Lanczos_Gnd.cpp:        !Type.is_float(Hop->dtype()),
src/linalg/Lanczos_Gnd_Ut.cpp:        !Type.is_float(Hop->dtype()),
bm_tests/linalg/Lanczos_bm.cpp:    if (Type.is_float(A.dtype())) {
bm_tests/linalg/Lanczos_bm.cpp:    if (Type.is_float(HEff.dtype())) {
tests/linalg_test/Arnoldi_test.cpp:    if (Type.is_float(this->dtype())) {
tests/linalg_test/Arnoldi_test.cpp:    if (Type.is_float(this->dtype())) {
tests/linalg_test/Lanczos_Exp_test.cpp:    if (Type.is_float(A.dtype())) {
tests/linalg_test/Lanczos_Exp_test.cpp:    if (Type.is_float(HEff.dtype())) {
tests/linalg_test/Arnoldi_Ut_test.cpp:    if (Type.is_float(this->dtype())) {
tests/gpu/linalg_test/Arnoldi_test.cpp:    if (Type.is_float(this->dtype())) {
tests/gpu/linalg_test/Arnoldi_test.cpp:    if (Type.is_float(this->dtype())) {
```

IvanaGyro: > The idea is that Traits<T>::is_XXX is guaranteed to map exactly same value as Type.is_XXX(dtype), and with the same spelling.

I agree with this statement, which is why I prefer using std directly. One potential issue with `is_complex_v<T>` arises if we decide to support complex number types beyond `std::complex<float>` and `std::complex<double>`. This is because `is_complex_v` is defined by us. To clarify, the suggestion for `is_complex_v` is only valid if we limit our support to these two types of complex numbers in the foreseeable future. 

My philosophy is not to reproduce the functions that we feel it has a bad name like `is_float`. Instead, those functions should be marked as deprecated. If we really want to pack `std::is_floating_point_v<T> || is_complex_floating_point_v<T>`,

```cpp
template <class T>
using is_float = std::is_floating_point_v<T> || is_complex_floating_point_v<T>;
```

is shorter and also satisfies the requirement.

I am okay with `Type_struct_t`. I just wanted to throw an idea to avoid wrapping std functions too much. Feel free to resolve this thread.

ianmccul: That is very neat!

I think `Type_struct` still has its uses though, because it collects all of the traits used by the library into a single location, rather than spreading them out into a bunch of individual traits. I can imagine it is useful, eg
```c++
void ShowTypeInfo(std::ostream& out, const Type_struct& t) {
   out << "Type:  " << t.name 
       << "\nId: " << t.cy_typeid 
       << "is_complex: " << t.is_complex
       << ....
}
```
If `type_struct` doesn't exist then you would need to implement this quite differently.

You probably still want some helper functions, since `is_int(1)` is much easier to type than `TypeTrait<cytnx::is_int>(1)`, but the helper functions could be implemented in terms of `TypeTrait`. And you definitely need `cytnx::is_int`, rather than `std::is_integral`. The reason is that the traits defined by the library are a *point of customization*, where the library, or users who are extending the library with their own types, can determine their own behavior.  `std::is_integral` is only true for the built-in C++ integral types, which, for historical reasons, includes `bool`.  Cytnx does not regard bool as integral, which is a reasonable choice for numerical algorithm purposes.  (Hence `static constexpr bool is_int = std::is_integral_v<T> && !std::is_same_v<T, bool>;`.)

if you have your own `bigint` class, for example, that you wanted to add to the library, how do you do it? You are not allowed to provide your own specializations of classes in `namespace std`, so you cannot cause `std::is_integral<bigint>` to be `std::true_type`. But you *can* specialize `cytnx::is_int<bigint>` and thereby configure the library to treat your `bigint` class as an integer.

Similarly for floating point types. I was thinking some more about this, and I think the distinctions that are useful for numerical algorithms are probably something like `is_real<T>`, `is_complex<T>` (used to select symmetric vs hermitian, or orthogonal vs unitary) and `is_real_or_complex<T>`.  There are many other traits that one could imagine, eg for rational arithmetic.  These are all points of customization where the library (and users writing library extensions) can choose how customized types should behave.  Even for the built-in types, none of these map directly onto the `std::` traits (except `is_real<T>` is probably the same as `std::is_floating_point_v<T>` the builtin types).

ianmccul:
## [Dispatch via std::variant, and type-safe conversion of Tensor::_impl->_storage._impl->Mem to a pointer](https://github.com/Cytnx-dev/Cytnx/pull/514/commits/710c98c73d24dba02bdbecad54dcfc8dcdf5b6fc)

### Type-safe conversions

```c++
pointer_types Tensor::ptr()
```
returns `this->_impl->_storage._impl->Mem`, contained in a `std::variant` of pointers to the supported types (excluding void).  If the dtype is void, then throw an error

```c++
template <typename T>
T* Tensor::ptr_as<T>() 
```
returns `this->_impl->_storage._impl->Mem` cast to `T*`.  If dtype does not match T, then throw an error

There are also versions for GPU storage, enabled if UNI_GPU is defined:

```c++
gpu_pointer_types Tensor::gpu_ptr()
```
returns a `std::variant` of pointers to the supported types, as required for CUDA (i.e. using cuComplex etc).

```c++
template <typename T>
T* Tensor::gpu_ptr_as<T>()
```
returns `this->_impl->_storage._impl->Mem` cast to `T*`.  If dtype does not match the CUDA type T, then throw an error

There is also a compile-time mechanism to promote types, as `Type_class::type_promote_t<T, U>`, which uses the  type_promote() function to determine the type corresponding to `type_promote(dtype_T, dtype_U)` (yay for constexpr functions!)  Calculating the promoted type via pointers is likely to be a common operation, so there is also `Type_class::type_promote_from_pointer_t<T,U>`, which requires that `T` and `U` are pointer types.  This is almost equivalent to `type_promote_t<std::remove_pointer_t<T>, std::remove_pointer_t<U>>`, except that if `T` is not actually a pointer type then `std::remove_pointer_t<T>` is just `T`, and hence gives the same result as `type_promote_t<T,U>`, but `type_promote_from_pointer_t<T,U>` is `void` if `T` or `U` are not pointer types (with the intent that this will eventually result in a compile error if it was not indented).  Note that `type_promote_from_pointer_t<T,U>` is NOT itself a pointer type.

Similarly, for GPU types, there is `type_promote_gpu_t` and `type_promote_from_gpu_pointer_t`.

### Dispatch via `std::variant`

[src/linalg/Kron.cpp](https://github.com/ianmccul/Cytnx/blob/98e3fca56d8dbbd44464e2917b08ecebbc75b5a9/src/linalg/Kron.cpp#L70-L78) demonstrates dispatch to a two-parameter function via `std::variant` and `Tensor::ptr()`. The key code is

```c++
        std::visit(
          [&](auto tl, auto tr) {
            // tl and tr are pointer types here.
            using out_type = Type_class::type_promote_from_pointer_t<decltype(tl), decltype(tr)>;
            static_assert(!std::is_same_v<out_type, void>);
            cytnx::linalg_internal::Kron_general(out.ptr_as<out_type>(), tl, tr, pad_shape1,
                                                 pad_shape2);
          },
          Tl.ptr(), Tr.ptr());
```

here, `Tl` and `Tr` are `Tensor` objects, and this does a double dispatch on the dtypes of `Tl` and `Tr`, via

```c++
std::visit(function, Tl.ptr(), Tr.ptr())
```

with `function` being a lambda function to forward the pointers to the template kernel `linalg_internal::Kron_general()` defined in[src/backend/linalg_internal_cpu/Kron_internal.hpp (https://github.com/ianmccul/Cytnx/blob/98e3fca56d8dbbd44464e2917b08ecebbc75b5a9/src/backend/linalg_internal_cpu/Kron_internal.hpp) (defined in the header, since it is a template function).

`out_type` is computed from the promotion of the pointer types of `tl` and `tr`, and the pointer to the output tensor is obtained from `out.ptr_as<out_type>()`.  This will throw an error if out_type does not match the previously assigned dtype, although it represents logic error in the code, rather than a user error.  In this case we can be confident it should never trigger, since the initialization of Tensor out immediately before dispatching to the kernel sets the dtype from `type_promote()`, which is using the same code to compute the promoted type as `type_promote_from_pointer_t`.

A slight disadvantage of this approach is that this requires that the template kernel is included into Kron.hpp, which increases compile time marginally but it means that linalg_internal_cpu/Kron_internal.cpp is no longer used.  Since this is a very long file that must be kept in sync with any changes to dtypes, the type promotion algorithm, and the Kron kernel itself, it is a substantial redunction in source code and maintenance.

Similarly, the static jump tables for the Kron function in backend/linalg_internal_interface.cpp have been removed (these were actually missing some entries).

The corresponding changes have also been done for the CUDA version, now using Tl.gpu_ptr() and Tr.gpu_ptr(), and dispatching to the CUDA kernel cuKron_general, and using type_promote_from_gpu_pointer_t to get the correct CUDA type of the output tensor.

A complication in the CUDA case is that the CUDA template kernel now needs to be included into Kron.cpp, so the cuKron_internal.cu has been renamed cuKron_internal.cuh (CUDA header file), and if UNI_GPU is defined then Kron.cpp must be compiled using the CUDA compiler.  This is managed by the local CMakeLists.txt file:

```cmake
 # This must be before the target_sources_local(), so that the target is added as CUDA
if(USE_CUDA)
  set_source_files_properties(Kron.cpp DIRECTORY ${CMAKE_SOURCE_DIR} PROPERTIES LANGUAGE CUDA)
endif()
```

set_source_files_properties() is directory-specific and by default only applies to targets defined in the current directory (and subdirectories?), hence it is necessary to explicitly set the `DIRECTORY` where the targets (cytnx) are defined.

### Possible future directions

Apply the above techniques to the other linear algebra functions, and eventually drop src/backend/linalg_internal_interface.[hpp|cpp].  This should be simple in some cases; other cases would need more substantial refactoring to merge type-specific kernels into template functions.

The `Tensor::ptr(), gpu_ptr()`, `ptr_as<T>()`, `gpu_ptr_as<T>()` functions do not really belong in class Tensor, it would be nice to move them into the Storage, ideally `ptr()` and `ptr_as<T>()` functions would only exist for some CpuStorage class, and similarly gpu_ptr() and gpu_ptr_as() would only exist for the GPU storage class. The dispatch to a back-end kernel could then go via two stages: firstly dispatch based on the device to a back-end storage (i.e. either cpu, or gpu), and then
that back-end handles the device-specific code.  Banish `gpu_ptr()` and `gpu_ptr_as<T>()` to backend/gpu/storage/....

The appearance of 'void' as one of the types in dtype was a complication that needed a hack to remove 'void' when constructing the variant of pointers. If void* is a possible type in the std::variant, then the dispatch via std::visit() is expecting that void* is a viable option to consider, and hence the first version of the Kron dispatch was attempting to compile the Kron kernel for tensors of type void...  Removing void from consideration in this way works, but it hasa disadvantage that the index of the type in the variant (i.e. `T.ptr().index()`) is no longer aligned with the dtype variable.  Not really a problem since there should not be any need to do that, but there is potential for that to go badly wrong if someone unwittingly expects them to be the same. Perhaps removing void completely from Type_class would be feasible, and instead treat any tensor with zero dimension as the equivalent of 'void'.

include/Tensor.hpp
```diff
@@ -493,6 +493,59 @@ namespace cytnx {
...
+    // appearing first ...
+    struct internal {
+      template <typename Variant>
+      struct exclude_first;
+
+      template <typename First, typename... Rest>
+      struct exclude_first<std::variant<First, Rest...>> {
+        using type = std::variant<Rest...>;
+      };
+    };  // internal
```
IvanaGyro, line 506: Maybe this can be put in Type.hpp. And add `dtypes` which is a variant or a tuple that does not contain void, as you suggested before.

include/Type.hpp
```diff
@@ -75,6 +76,20 @@ namespace cytnx {
...
   }  // namespace internal
 
+  // helper metafunction to transform a variant into another variant via a
+  // transform template alias
+  template <typename V, template <typename> class Transform>
+  struct make_variant_from_transform;
+
+  template <template <typename> class Transform, typename... Args>
+  struct make_variant_from_transform<std::variant<Args...>, Transform> {
+    using type = std::variant<typename Transform<Args>::type...>;
```
IvanaGyro, line 86: `using type = std::variant<typename Transform<Args>...>;` would enable us to transform the variant to the variant of the StorageImeplementation (typed storage).

To be compatible with `std`, we can use `make_variant_from_transform<std::variant<...>, std::add_pointer_t>`.

IvanaGyro (approved):
Excellent! Would you rebase the commits onto the dev branch?

include/Tensor.hpp
ianmccul: Perhaps.  I would first look at how feasible it is to remove `void` from `Type_class`, rather than having two versions of the type list that differ only by one element in the public interface.

include/Type.hpp
ianmccul: Yes, that is why I put this metafunction in `Type.hpp`, it could be used in other places.

ianmccul:
As I said in my email, I am not going to spend more time working on this, but anyone else is welcome to carry this forward.

I didn't try rebasing this on dev-master, but from a quick look, it will probably apply cleanly or at most minor changes.

It would probably be good to move the implementation of `ptr()`, `ptr_as()` etc to Storage_base, and the `Tensor` versions could just forward to the storage.  I didn't do this already as I wasn't sure how it would fit in with @IvanaGyro's refactoring of Storage.

IvanaGyro:
I got it. I will carry this forward.

PR #515, opened 2024-11-21
Remove some duplicate code

ianmccul:
There is a lot of duplicate code in cytnx. This patch removes some duplication of code in the file I/O in src/Tensor.cpp, src/UniTensor.cpp, src/backend/Storage.cpp.  There are sometimes 3 copies of a function, that differ only by taking a filename as an std::string, a const char*, or an fstream. These should be consolidated to one function. I made a start at this, and also improved some of the error messages.

A reasonable model is:

```c++
// Save(std::string) forwards to Save(const char*)
void Tensor::Save(const std::string &fname) const {
  Tensor::Save(fname.c_str());
}

// Save(const char*) does some error checking and forwards to Save(fstream&)
void Tensor::Save(const char *fname) const {
  fstream f;
  string ffname = string(fname) + ".cytn";
  f.open(ffname, ios::out | ios::trunc | ios::binary);
  cytnx_error_msg(!f.is_open(), "[ERROR] invalid file path for save: %s", fname);
  this->_Save(f);
  f.close();
}

void Tensor::_Save(fstream &f) const {
// non-trivial code in here
}
```

I also improved some of the error messages, although I notice that I missed the one that I copy-pasted above.

A common pattern I see is code like this:

```
if (!f.is_open()) {
  cytnx_error_msg(true, "[ERROR] invalid file path for save.%s", "\n");
}
```

I wonder if this is partly due to the name cytnx_error_msg -- it is not obvious from the name what this does, does it unconditionally show an error message, or does it do some check, or what?  I suggest to rename `cytnx_error_msg` to
```c++
cytnx_error_if(Condition, Message, ...)
```
to make it clear that it is doing a test of a condition and it does not need to be enclosed in an if (Condition) statement.

Almost every cytnx_error_msg call has "[ERROR] " at the start of the string. The few calls that do not are probably unintentional. So a nice improvement would be to move the "[ERROR] " part of the message into the implementation of cytnx_error_xxx, so the caller doesn't have to repeat the same string every time.

There are also a lot of cytnx_error_xxx calls that include some information about the sub-module, eg "[ERROR][Trace] invalid Type.%s".  A better mechanism would be to specify the sub-module separately, for example something like
```c++
CYTNX_SUBMODULE("Trace");
```
which could go at the top of a C++ file (or anywhere else, really), and some macro magic would include the module string in the error message.

### DIFF
--- a/src/Tensor.cpp
+++ b/src/Tensor.cpp
@@ -367,7 +367,12 @@ namespace cytnx {
   // wrapper
 
   void Tensor::Tofile(const std::string &fname) const {
-    this->Tofile(fname.c_str());
+    if (!this->is_contiguous()) {
+      auto A = this->contiguous();
+      A.storage().Tofile(fname);
+    } else {
+      this->_impl->_storage.Tofile(fname);
+    }
   }
   void Tensor::Tofile(const char *fname) const {
     if (!this->is_contiguous()) {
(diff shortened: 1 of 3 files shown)


ianmccul:
I didn't rename the `_Save` function, although it should be: it is a name that is reserved.

kaihsin:
Nice. This is great.

kaihsin (approved):

PR #523, opened 2024-11-29
Add cmake configuration for the location of the test data

IvanaGyro:
This is a PR that rebases #513 onto the dev branch. Please see #513 for the details.


manuschneider (approved):
Works fine on my system, also if I move the test data folder elsewhere and change the path in `tests/CMakeLists.txt` file accordingly.

yingjerkao (approved):

Issue #517, opened 2024-11-22
Catastrophically slow usage of OpenMP


ianmccul:
I came across this fragment of code:

https://github.com/Cytnx-dev/Cytnx/blob/fa4596e1620cc304f698286905818a2118b2bd8e/src/backend/linalg_internal_cpu/QR_internal.cpp#L23-L28

This is a misuse of OpenMP. As discussed in #511, OpenMP with schedule(dynamic), without specifying a chunk size, does a single iteration of the loop in each thread. It is intended for cases where each iteration does a significant amount of work that may take a variable amount of time.  It is not appropriate to use `schedule(dynamic)` in loops where the iterations will take a fixed amount of time.  Also, if you use `schedule(dynamic)`, unless each iteration takes a very long time, you should set the chunk size, eg 
```
#pragma omp parallel for schedule(dynamic,100000)
```

The effect of using `schedule(dynamic)` on a tight loop is catastrophic. From the `Fill` example discussed in #511,
| Benchmark            | Time (ms) |
|----------------------|-----------|
|              FillCpu | 1.20837     |
|        FillCpuStatic | 1.6324     |
|       FillCpuDynamic | 278.007     |
|  FillCpuDynamicChunk | 1.8086     |

As this benchmark shows, there really isn't a useful benefit from using OpenMP at all in memory-bound tight loops such as the above; at best you get a few percent improvement even with lots of threads. That particular example is a reimplementation of BLAS `dcopy` anyway, and it is hard to imagine `N` ever being large enough to benefit from hand-optimization (the QR decomposition is $O(N^3)$, so speeding up an $O(N)$ loop won't do much). But using `schedule(dynamic)` means the loop will be running ~ 200x slower.

On further checks, the string `schedule(dynamic)` appears 2360 in the codebase.  I checked a random sample of 10 of these. **7** of them were definitely erroneous, with code essentially identical to the fragment above.  The remaining 3 are most likely erroneous, but *might* benefit from OpenMP if done carefully (they were also 3 copies of the same code fragment, that probably should be a helper function).  To confirm, I checked an additional sample of 10, with the same mix of 7 definitely erroneous, and 3 likely erroneous.

It looks like there is some attempt to have some benchmarks in the `bm_tests/` directory, but it looks like it is disabled in CMakeLists.txt.  Is there a way to build the benchmarks?

I honestly don't know what to do here.  Some options:

- use `sed` to remove all `#pragma omp` lines that contain `schedule(dynamic)`
[...]

PR #526, opened 2024-11-30
Remove UNI_OMP guards surrounding `#pragma omp...`

IvanaGyro:
## Remove UNI_OMP guards surrounding `#pragma omp...`

The compilers don't stop compiling if they cannot find the token
following `#pragma`, so we don't have to use the compile definition to
guard `#pragma omp...` statements.

## Remove `find_package(GTest)` from the root CMake file 
We have used `include(GoogleTest)`, which is the new mechanism, to discover the
tests.

This is a change for #517.

## Remove `-fopenmp` flag when only enabling MKL

We don't guard `#pragam omp...` statements now, so they take effects
when `-fopenmp` compiler flag is set. `set()` function only affect the
targets defined in the same scope (the same directory) and the
subdirectories after the line of `set()`. Setting `-fopenmp` flag when
the user only enable `USE_MKL` but not enable `USE_OMP` makes the user
suprised.

By default, MKL links to their own openmp (not gomp) for threading. The
build process has no problem with no setting `-fopenmp` compiler flag
when using MKL as the provider of BLAS and LAPACK. The above claim is
true for the MKL version later then 2023.2.0. I am not sure if it is
also true for the older version of MKL.


manuschneider (approved):
Seems good for now. But in #517 suggests to remove `#pragma omp parallel for schedule(dynamic)` everywhere, so the issue still needs some work.

ianmccul:
> Seems good for now. But in #517 suggests to remove `#pragma omp parallel for schedule(dynamic)` everywhere, so the issue still needs some work.

Yes, all of the `#pragma omp` code I saw uses `schedule(dynamic)` in a way that makes it much slower than not using OpenMP at all.  The majority of OpenMP loops that I looked at were level-1 operations, i.e. one new memory access per loop iteration.  These don't benefit much from OpenMP, because they are saturating the memory bandwidth. Replacing the `schedule(dynamic)` with `schedule(static)` would certainly be better than the current situation, but that will at best give a few percent speedup, on code that is not a bottleneck anyway.  I  suggest simply removing all of the `#pragma omp` lines completely.

Perhaps try some test case with 3 benchmarks:

1. the code as it is
2. globally replace `schedule(dynamic)` with `schedule(static)`
3. disable OpenMP - either compiler flag or globally remove all lines `#pragma omp ...`

I tried (1) and (3), using `C = A+B` for `A`, `B`, `C` being 10000x10000 tensors of `Double`.  `bench-add-noomp` is linking against cytnx compiled without OpenMP, `bench-add` is the same code linking against an OpenMP build.
```
$ ./bench-add-noomp
Average time for Cytnx Tensor: 0.212623 seconds
$ OMP_NUM_THREADS=1 ./bench-add
Average time for Cytnx Tensor: 0.781158 seconds
$ OMP_NUM_THREADS=10 ./bench-add
Average time for Cytnx Tensor: 3.15877 seconds
```
It gets worse the more threads you add because they are all fighting each other for the same L1 cache entries.

IvanaGyro:
Yes, the issue is still there. Because I cannot confirm that all `#pragma omp` code is redundant, I didn't remove it by meta programming. I will check them by hand, but this will take some time.

IvanaGyro:
For non-complex, maybe just use `std::max_element` :)

ianmccul:
> For non-complex, maybe just use `std::max_element` :)

And for the complex case too, just need to decide what the comparison function is.

PR #528, opened 2024-12-01
Encapsulate Storage_base class

IvanaGyro:
**The first two commits are the same as #525. Please start to review from the third commit.**
**There are too many files changed to review. I leave comments on the changed lines which are not trivially renaming the variables. You can just review those parts.**

With the current class definition, developers are able to change the variable storing the size of the storage without allocating the underlying storage, which potentially causes memory leaks.

To ensure there will be no surprising behavior caused by directly accessing the variables of Storage_base, all variables of Stroage_base are moved to its subclass, StorageImplementation<DType>, and are set as private. All functions outside of StorageImplementation<DType> should access the value of the variable via the getters.

The next step in refactoring the components related to Storage is to remove the Storage_base class.

## Test result

This PR breaks no tests with CUDA, cuQuantum, cuTENSOR, MKL, CUTT, and HPTT enabled.

### DIFF
--- a/include/backend/Scalar.hpp
+++ b/include/backend/Scalar.hpp
@@ -393,6 +393,9 @@ namespace cytnx {
       return nullptr;
     }
 
+    /**
+     * @deprecated This method is not in use anymore.
+     */
     virtual void *get_raw_address() const {
       cytnx_error_msg(true, "[ERROR] Void Type Scalar cannot have operation!!%s", "\n");
       return nullptr;
--- a/src/Tensor.cpp
+++ b/src/Tensor.cpp
@@ -76,7 +76,7 @@ namespace cytnx {
     cytnx_error_msg(this->dtype() == 0, "[ERROR] operation not allowed for empty (void) Tensor.%s",
                     "\n");
     // dtype()-1 here because we have removed void from the variant
-    return void_ptr_to_variant_impl(this->_impl->_storage._impl->Mem, this->dtype() - 1,
+    return void_ptr_to_variant_impl(this->_impl->_storage._impl->data(), this->dtype() - 1,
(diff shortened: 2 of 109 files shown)


include/backend/Scalar.hpp
```diff
@@ -392,6 +392,9 @@ namespace cytnx {
       return nullptr;
     }
 
+    /**
+     * @deprecated This method is not in use anymore.
+     */
     virtual void *get_raw_address() const {
       cytnx_error_msg(true, "[ERROR] Void Type Scalar cannot have operation!!%s", "\n");
       return nullptr;
```
IvanaGyro, line 400: Mark this function as deprecated. I may be deleted or renamed to `data()` like `std` convention.

src/backend/Storage.cpp
```diff
@@ -136,18 +136,21 @@ namespace cytnx {
...
     f.write((char *)&IDDs, sizeof(unsigned int));
-    f.write((char *)&this->size(), sizeof(unsigned long long));
-    f.write((char *)&this->dtype(), sizeof(unsigned int));
-    f.write((char *)&this->device(), sizeof(int));
+    auto write_number = [&f](auto number) {
+      f.write(reinterpret_cast<char *>(&number), sizeof(number));
+    };
+    write_number(this->size());
+    write_number(this->dtype());
+    write_number(this->device());
```
IvanaGyro, line 144: `this->size()` and the other statements are not `lvalue` now, so we cannot access the address by `&`.

src/Tensor.cpp
manuschneider, line 1: Where are these functions used?

hunghaoti (approved):

hunghaoti (approved):

PR #529, opened 2024-12-04
Uniform randomizing function respects given range

IvanaGyro:
This change fixes the problem that the uniform random number generator ignores the given lower bound and higher bound when the view types (Storage, Tensor, and UniTensor) are assigned on the GPU.


IvanaGyro:
Some the helper classes for manipulating types are duplicated with #490. This PR doesn't base on #490, so this PR implement them again. I will integrate them after merging these two PRs.

yingjerkao (approved):

PR #530, opened 2024-12-04
Det() returns zero for the sigular matrix

IvanaGyro:
According to cuSOLVER's document, returned `devInfo` greater than zero means U[devInfo - 1, devInfo - 1] (zero-indexed) is zero, which implies the determinant is zero.


IvanaGyro:
Link to the document: https://docs.nvidia.com/cuda/cusolver/index.html#cusolverdn-t-getrf

> If LU factorization failed, i.e. matrix A (U) is singular, The output parameter devInfo=i indicates U(i,i) = 0.
> If output parameter devInfo = -i (less than zero), the i-th parameter is wrong (not counting handle).

yingjerkao (approved):

PR #533, opened 2024-12-05
Det() handles the sigular matrix on CPU

IvanaGyro:
According to LAPACK's document, returned `info` greater than zero means U[info - 1, info - 1] (zero-indexed) is zero, which implies the determinant is zero.


manuschneider (approved):
See https://www.netlib.org/lapack//explore-html/db/d04/group__getrf_ga85d0f9ae367fd540a745f558a77717db.html

PR #534, opened 2024-12-06
Fix and refactor tests for Det()

IvanaGyro:
Changes:

1. Set the tolerence to 1e-6 * expected_value
2. Used EXPECT_THAT and matchers to provide better error messages
3. Renamed the tests to follow the rules of gtest: do not use "_" on the name of the test

The failed tests will pass after merging #533.

### DIFF
--- a/tests/gpu/linalg_test/Det_test.cpp
+++ b/tests/gpu/linalg_test/Det_test.cpp
@@ -1,18 +1,38 @@
-#include <gtest/gtest.h>
+#include <exception>
+#include <iomanip>
+
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
 
 #include "../test_tools.h"
+#include "Device.hpp"
+#include "Type.hpp"
 
-using namespace cytnx;
-using namespace testing;
-using namespace TestTools;
+namespace cytnx {
 
-namespace DetTest {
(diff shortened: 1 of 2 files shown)


yingjerkao (approved):

IvanaGyro:
@hunghaoti  Merge this to fix the build with tests.

hunghaoti:
> @hunghaoti Merge this to fix the build with tests.

@IvanaGyro , I have tried to build it but still cannot build successfully. The error message is 
```
/home/hunghaoti/workplace/Cytnx/Cytnx_gpu/src/backend/utils_internal_gpu/cuFill_gpu.cu(35): error: no suitable user-defined conversion from "const cytnx::cytnx_complex128" to "CudaDType" exists
          detected during instantiation of "void cytnx::utils_internal::FillGpu(void *, const DType &, cytnx::cytnx_uint64) [with DType=cytnx::cytnx_complex128]" 
(38): here

```

IvanaGyro:
The error is not related to this PR. Upgrading the CUDA version above 11.4 should solve the problem. Below 11.4, there is no `cuda::std::complex`. Refer to: https://nvidia.github.io/cccl/libcudacxx/standard_api/numerics_library.html

PR #540, opened 2024-12-14
Create CITATION.cff, to add doi and url for web site.

y3jo6:
hi Authors,
We try to find the doi for your work andn create the CITATION.cff file with realted information. Since we can't clarify if that is the doi you want to attached in this repo.  kinldy confirm that.


manuschneider:
It should be version 1.0.0 now. Not sure how I can change the patch.

IvanaGyro (approved):

Issue #505, opened 2024-11-09
`linalg::Svd()` outputs NaN when the input tensor contains only zeros


IvanaGyro:
`linalg::Svd()` outputs NaN when the input tensor contains only zeros. This issue only happens on GPU and doesn't happen when the data type is float. This bug is the culprit of the broken `Svd.gpu_U1_zeros_test` test case. 

Below is the minimal test case to reproduce the bug.

```cpp
  TEST(Svd, GpuZeros) {
    cudaSetDevice(Device.cuda);
    Tensor all_zeros = zeros(64, Type.Double, Device.cuda);
    all_zeros.reshape_(16, 4);
    std::vector<Tensor> svd_output = linalg::Svd(all_zeros, /* is_UvT */ true);
    cudaDeviceSynchronize();
    ASSERT_EQ(svd_output.size(), 3);
    Tensor& singular_values = svd_output[0];
    ASSERT_EQ(singular_values.shape().size(), 1);
    ASSERT_EQ(singular_values.shape()[0], 4);
    EXPECT_EQ(singular_values.at({0}), 0);
    EXPECT_EQ(singular_values.at({1}), 0);
    EXPECT_EQ(singular_values.at({2}), 0);
    EXPECT_EQ(singular_values.at({3}), 0);
  }
```

current output:

```
```

IvanaGyro:
Only when the size of singular values are 4, 23, 28, and 33, the output 

```cpp
  TEST(Svd, GetZerosWhenInputtingZeros) {
    cudaSetDevice(Device.cuda);
    auto any_non_zero = [](const Tensor& tensor) -> bool {
      Storage storage = tensor.storage();
      size_t size = storage.size();
      double* ptr = reinterpret_cast<double*>(storage.data());
      for (int i = 0; i < size; ++i) {
        if (*ptr++ != 0) {
          return true;
        }
      }
      return false;
    };
    std::vector<std::pair<size_t, size_t>> shapes_get_nan;
    for (size_t m = 1; m < 61; ++m) {
      for (size_t n = 1; n < 61; ++n) {
        Tensor all_zeros = zeros(m * n, Type.Double, Device.cuda);
        all_zeros.reshape_(m, n);
        std::vector<Tensor> svd_output = linalg::Svd(all_zeros, /* is_UvT */ true);
        Tensor& singular_values = svd_output[0];
        if (any_non_zero(singular_values)) {
          shapes_get_nan.emplace_back(m, n);
        }
      }
[...]

PR #545, opened 2024-12-27
Fix gpu unit test.

hunghaoti:
1. Fix Lanczos Gnd unit test cause from putblock using different devices.
2. Skip ExpM for gpu test since Eig still not implemted, which is needed to do the exponential.
3. cuda svd cannot give the correct answer if almost elements are zeros. So we skip Svd U1_zeros tests.
4. The answer of Svd_truncate3 is not correct also cause from the above issue. So I delete it.

### DIFF
--- a/tests/gpu/linalg_test/Directsum_test.cpp
+++ b/tests/gpu/linalg_test/Directsum_test.cpp
@@ -24,11 +24,11 @@ namespace DirectsumTest {
     axes:{1}
   ====================*/
   TEST(Directsum, gpu_allDType) {
-    for (auto device : device_list) {  // now only test for cpu device.
+    for (auto device : device_list) {
       for (auto dtype1 : dtype_list) {
         for (auto dtype2 : dtype_list) {
-          Tensor T1 = Tensor({12, 5, 7}, dtype1, device);
-          Tensor T2 = Tensor({12, 5, 8}, dtype2, device);
+          Tensor T1 = Tensor({12, 5, 7}, dtype1, device).to(cytnx::Device.cuda);
+          Tensor T2 = Tensor({12, 5, 8}, dtype2, device).to(cytnx::Device.cuda);
           InitTensorUniform(T1, rand_seed1 = 0);
           InitTensorUniform(T2, rand_seed2 = 1);
           std::vector<cytnx_uint64> shared_axes = {1};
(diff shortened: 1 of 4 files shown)


IvanaGyro:
Just leave a reference. The third problem was addressed in #505.

tests/gpu/linalg_test/linalg_test.cpp
```diff
@@ -25,6 +25,7 @@ TEST_F(linalg_Test, gpu_BkUt_Svd_truncate2) {
   auto con_T2 = Contract(Contract(res[1], res[0]), res[2]);
 }
 
+/*
 TEST_F(linalg_Test, gpu_BkUt_Svd_truncate3) {
   std::vector<UniTensor> res = linalg::Svd_truncate(svd_T, 200, 0, true);
```
IvanaGyro, line 30: Maybe just skip them or disable them, so do we know we have to fix them in the future.

tests/gpu/linalg_test/Directsum_test.cpp
```diff
@@ -24,11 +24,11 @@ namespace DirectsumTest {
     axes:{1}
   ====================*/
   TEST(Directsum, gpu_allDType) {
-    for (auto device : device_list) {  // now only test for cpu device.
+    for (auto device : device_list) {
       for (auto dtype1 : dtype_list) {
         for (auto dtype2 : dtype_list) {
-          Tensor T1 = Tensor({12, 5, 7}, dtype1, device);
-          Tensor T2 = Tensor({12, 5, 8}, dtype2, device);
+          Tensor T1 = Tensor({12, 5, 7}, dtype1, device).to(cytnx::Device.cuda);
```
IvanaGyro, line 30: The device_list in test_tools.h seems wrong. Could you create an issue to address it?

IvanaGyro (approved):
Does the first problem imply there is a bug in the function to put blocks while the device is the GPU?

IvanaGyro:
Feel free to close the conversations.

tests/gpu/linalg_test/linalg_test.cpp
hunghaoti: This problem occurs since the answer from dense is wrong. I think gpu_BkUt_Svd_truncate1 is enough.

hunghaoti:
> Does the first problem imply there is a bug in the function to put blocks while the device is the GPU?

The first problem cause since we cannot put a Tensor in GPU device to the UniTensor in CPU device. So we just create Tensors in CPU and put it in UniTensor in CPU. And then transfer to GPU finally.

tests/gpu/linalg_test/Directsum_test.cpp
hunghaoti: I have leave the issue #548 .

PR #547, opened 2024-12-27
Update readme.

hunghaoti:


manuschneider:
I think a longer description of the features UniTensor would be helpful (labels, symmetries, block structure etc). The linear algebra part is also missing, we should at least mention Svd (truncate) and Eig.
These are the things that we actually want people to use. Maybe also change the order from UniTensor -> Linear algebra -> Tensor -> Storage

hunghaoti:
Move to PR #549 . Please discuss at the stream PR #549 .
Instead of merging into dev-master, I merged to pre-releasev1.0.0.

PR #553, opened 2025-01-02
Fix macos building error.

hunghaoti:
Fix macos building error.


include/Type.hpp
```diff
@@ -134,9 +134,9 @@ namespace cytnx {
     // type_size returns the sizeof(T) for the supported types. This is the same as
     // sizeof(T), except that size_type<void> is 0.
     template <typename T>
-    constexpr int type_size = sizeof(T);
```
IvanaGyro, line 137: Which compiler rejects this?

hunghaoti: Apple Clang. It will return error like
```
  duplicate symbol 'cytnx::internal::type_size<void>' in:
      CMakeFiles/pycytnx.dir/pybind/cytnx.cpp.o
      CMakeFiles/pycytnx.dir/pybind/bond_py.cpp.o
```
See [this workflow](https://github.com/Cytnx-dev/Cytnx/actions/runs/12579093565/job/35058724705) and [ref](https://stackoverflow.com/questions/64610024/duplicate-symbols-with-global-template-variable-using-clang)

IvanaGyro: Great! Thanks for the ref. However, if this issue also happens with Clang, can we support Clang and not support Apple Clang? Apple Clang only picks a subset of Clang (see https://github.com/Cytnx-dev/Cytnx/issues/449#issuecomment-2323257106) and fixing compile issues without a Mac is so frustrating.

hunghaoti: I found this issue when I tried to merge to master branch. Since there are a workflow need to build on macos.

hunghaoti: Hi @kaihsin , what do you think. If we don't want to support `macos`, we may need to delete this [workflow](https://github.com/Cytnx-dev/Cytnx/actions/workflows/conda_build_macos.yml). However, if we want to support, we also need to add the workflow for branch `dev-master` to test macos building. Otherwise, it is hard to debug when I try to merge `dev-master` to `master` branch.

CMakeLists.txt
```diff
@@ -96,7 +96,7 @@ project(CYTNX VERSION ${CYTNX_VERSION} LANGUAGES CXX C)
 
 # C++ uses link-time optimization anyway; this enables additionally -flto=auto,
 # for parallel compilation
-set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
+#set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
```
yingjerkao, line 99: It's better guard this with a check for MacOS using APPLE, instead directly comment out this. 
https://cmake.org/cmake/help/latest/variable/APPLE.html

```
IF (APPLE)  
set(CMAKE_INTERPROCEDURAL_OPTIMIZATION FALSE)
ELSE
set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
ENDIF

```

ianmccul: And add a comment explaining why it was done (with details of the compiler, and the error it causes etc)

hunghaoti: Just also leave the comment here. We need to disable `CMAKE_INTERPROCEDURAL_OPTIMIZATION`on MacOS since the building issue (See this [MacOS building event](https://github.com/Cytnx-dev/Cytnx/actions/runs/12600085136/job/35118332595) Line 941.

yingjerkao (approved):

Issue #563, opened 2025-02-27
The  DMRG sweeping make the memory(RAM) increase in Cytnx 1.0.0 version


Heliumky:
Because some parts of the code are private, I have already sent the dmrg.py file to Haoti. In this script, we frequently use the functions `cytnx.Network`, `cytnx.linalg.Lanczos`, and `cytnx.linalg.Svd_truncate`. I am not sure whether updating Cytnx from 0.9.7 to 1.0.0 will affect these functions. Additionally, it seems that these functions may have a memory increase, as the RAM is not properly released.

hunghaoti:
It seems that the function **Contract()** has some problem (file:test_2.py), which cause this issue:
``` python
import sys,os
sys.path.insert(0,"/home/hunghaoti/Cytnx_lib")
import cytnx
print(cytnx.__version__)

@profile #comment this line if you don't use Memory Profiler
def test_Contract(A, B): 
    for i in range(100):
        res = cytnx.Contract(A, B)
    return res 

D = 100 
tmp = cytnx.random.uniform([D,D,D], low=-1., high=1.)
A = cytnx.UniTensor (tmp, labels=['c','a1','a2'])
test_Contract(A, A)
```
If using [Memory Profiler](https://pypi.org/project/memory-profiler/)
```bash
python -m memory_profiler test_2.py
```
The output for version **v1.0.0**:

```
1.0.0
Filename: test_2.py

Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
     6   86.004 MiB   86.004 MiB           1   @profile
     7                                         def test_Contract(A, B):
     8 1614.949 MiB    0.000 MiB         101       for i in range(100):
     9 1614.949 MiB 1528.945 MiB         100           res = cytnx.Contract(A, B)
    10 1614.949 MiB    0.000 MiB           1       return res

```
The output for version **v0.9.7**:
```
0.9.7
Filename: test_2.py

Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
     6   81.391 MiB   81.391 MiB           1   @profile
     7                                         def test_Contract(A, B):
     8   84.723 MiB    0.000 MiB         101       for i in range(100):
     9   84.723 MiB    3.332 MiB         100           res = cytnx.Contract(A, B)
    10   84.723 MiB    0.000 MiB           1       return res

```
And the performance (efficiency) is much better thant **v1.0.0**.

hunghaoti:
[...]

PR #564, opened 2025-03-02
fix leak bug in calling override/virtual func in destructor

kaihsin:
This PR address #563

### DIFF
--- a/include/backend/Storage.hpp
+++ b/include/backend/Storage.hpp
@@ -44,7 +44,7 @@ namespace cytnx {
     virtual const unsigned long long size() const {
       cytnx_error_msg(true, "Not implemented.%s", "");
     }
-    virtual ~Storage_base();
+    ~Storage_base();
 
     template <class T>
     T &at(const cytnx_uint64 &idx) const;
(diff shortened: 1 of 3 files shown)


include/backend/Storage.hpp
```diff
@@ -44,7 +44,7 @@ namespace cytnx {
     virtual const unsigned long long size() const {
       cytnx_error_msg(true, "Not implemented.%s", "");
     }
-    virtual ~Storage_base();
+    ~Storage_base();
```
IvanaGyro, line 47: The destructor of the base class must be virtual to let the destructor of the subclass be called properly while deleting the pointer.  I intended to move all arguments into the subclass because `Storage_base` was planned to be removed or be designed as a pure virtual class.

kaihsin: Either this or you need to implement a destructor for `StorageImplementation`. Can't just use the parent one

kaihsin: I am ok with both. If the intent is to  remove Storage_base, then lets add destructor for StorageImplementation

IvanaGyro: Moving the function of destructor of `Storage_base` to `StorageImplementation` is better.

kaihsin: done

kaihsin: feel free to close the issue and tag+release a patch once its merged

hunghaoti: Should we bump up the version to v1.0.1 or just retag v1.0.0?

kaihsin: > Should we bump up the version to v1.0.1 or just retag v1.0.0?

Please don't retag. just do a patch version 1.0.1

yingjerkao (approved):

IvanaGyro (approved):

PR #567, opened 2025-03-10
Upgrade project C++ standard to C++20

IvanaGyro:
C++20 includes several valuable features:
- Concepts (clearer template constraints)
- Ranges (improved algorithm readability)
- Three-way comparison operator (<=>) (simplifies comparison logic)
- `std::span` (easier handling of contiguous data)
- `std::format` (modern string formatting)

These features will help improve code readability.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@ include(ExternalProject)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules")
 message(STATUS ${CMAKE_MODULE_PATH})
 
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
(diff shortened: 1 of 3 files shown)


kaihsin:
Let's chat on this a bit more. Specifically I also want to explore a bit on module from C++20

kaihsin:
Would you mind focus on the refactor of storage first before bumping this up to C++20? Or you need any features from C++20?

yingjerkao:
I would suggest to use `std::mdspan` which will be part of C++23, and has a reference implementation 

https://en.cppreference.com/w/cpp/container/mdspan

https://github.com/kokkos/mdspan

ianmccul:
Modules are not really usable in C++20.  I think everyone is hoping C++26
fixes it....

IvanaGyro:
> I would suggest to use `std::mdspan` which will be part of C++23, and has a reference implementation
> 
> https://en.cppreference.com/w/cpp/container/mdspan
> 
> https://github.com/kokkos/mdspan

GCC somehow [hasn't implemented `mdspan`](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=107761). I learned this from [the compiler support table](https://en.cppreference.com/w/cpp/compiler_support/23). There are two options. One is to drop the support of GCC and use Clang only. The other is making kokkos' implementation as a dependency until GCC implements `mdspan`. Do you have any ideas?

yingjerkao:
> GCC somehow [hasn't implemented `mdspan`](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=107761). I learned this from [the compiler support table](https://en.cppreference.com/w/cpp/compiler_support/23). There are two options. One is to drop the support of GCC and use Clang only. The other is making kokkos' implementation as a dependency until GCC implements `mdspan`. Do you have any ideas?

Since the reference implementation is pretty complete and should be simply replaced to `std::mdspan` when the compiler is ready, we should go this route.

IvanaGyro:
cuda::std also implemented [mdspan](https://nvidia.github.io/cccl/libcudacxx/standard_api/container_library/mdspan.html?highlight=mdspan).

hunghaoti (approved):

IvanaGyro:
I'm not sure it the discussion will happen. In order to move forward, I am merging this PR now. We can revert it in the future.

PR #575, opened 2025-03-13
Haoti/omp bm

hunghaoti:


CytnxBKNDCMakeLists.cmake
hunghaoti, line 1: This file needs to be review.

benchmarks/linalg/QR_bm.cpp
```diff
@@ -0,0 +1,26 @@
+#include <benchmark/benchmark.h>
+#include <cytnx.hpp>
+
+// Cytnx test
+namespace BMTest_QR {
```
IvanaGyro, line 5: The name of namespace should [be snake case](https://google.github.io/styleguide/cppguide.html#Namespace_Names). Maybe we can just put the test functions in `namespace cytnx {}` directly or in `namespace cytnx {namespace benchmark {}}`?

IvanaGyro (approved):

PR #579, opened 2025-03-15
Replace `Install.sh` with CMakePresets.json

IvanaGyro:
## Motivation

[cmake-presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html) is a common way to share settings for CMake and is also supported by some editor extensions, e.g., vscode-cmake-tools. Using cmake-presets to configure the project is more intuitive and convenient than using `Install.sh`, an Unix-only shell script.

## Changes

- All editable settings of this project are collected in a section in CMakeLists.txt and CMakePresets.json.
- Pre-defined some useful setting combinations as presets.
- The presets enabling CUDA also enable cuTENSOR and cuQuantum. This decision is to reduce the number of presets.
- cuTENSOR and cuQuantum are not enabled automatically when CUDA is enabled. If you only want to enable CUDA, the cmake command is changed from `cmake -DUSE_CUDA=ON -DUSE_CUTENSOR=OFF -DUSE_CUQUANTUM=OFF` to `cmake -DUSE_CUDA`. If you want to enable CUDA, cuTENSOR, and cuQuantum, the command is changed from `cmake -DUSER_CUDA` to `cmake -DUSE_CUDA=ON -DUSE_CUTENSOR=ON -DUSE_CUQUANTUM=ON`.

## Limitation

There is no way to parallel the build and test process with the maximum number of processors or unlimited number like `make -j`. [A feature request](https://gitlab.kitware.com/cmake/cmake/-/issues/23648) to CMake was created two years ago. 

`CUTENSOR_ROOT` and `CUQUANTUM_ROOT` are not explicitly added into CMakePresets.json because vscode-cmake-tools, a vscode extension for CMake, doesn't pass environment variables to presets. The users of the extension will have no intuitive way to pass the value of `CUTENSOR_ROOT` and `CUQUANTUM_ROOT` for configuration. As an alternative, we show an error message to the user when they doesn't set `CUTENSOR_ROOT` or `CUQUANTUM_ROOT` in the environment variables.


IvanaGyro:
I will create a PR  in Cytnx_doc.

IvanaGyro:
PR in Cytnx_doc is opened. https://github.com/Cytnx-dev/Cytnx_doc/pull/1

yingjerkao (approved):

Issue #580, opened 2025-03-15
CUDA, CUTENSOR, CUQUANTUM


pcchen:
It seems that if I turn cuda option on via `-DUSE_CUDA=ON`, then one automatically turns on cutensor and cuquantum. If this is the case, then maybe we don't need `-DUSE_CUTENSOR=ON` and `-DUSE_CUQUANTUM=ON`?

IvanaGyro:
Are you discussing the presets created in #579?

With the current settings (without #579), enabling CUDA will not automatically enable cuTENSOR and cuQuantum.

In #579, CMakeLists.txt follows the current settings. cuTENSOR and cuQuantum are not enabled when configuring with the command `cmake -DUSE_CUDA`. cuTENSOR and cuQuantum are only enabled with CUDA when using presets. `cmake --preset openblas-cuda` enables cuTENSOR and cuQuantum.

I updated the PR message of #579.

pcchen:
Question 1:

Can one build gpu support without using CUTENSOR and CUQUANTUM?

pcchen:
I have to explicitly turn off the cutensor/cuquantum. `
cmake -DUSE_CUDA=ON -DUSE_CUTENSOR=OFF -DUSE_CUQUANTUM=OFF ..`

````
-- ------------------------------------------------------------------------
--   Project Cytnx, A Cross-section of Python & C++,Tensor network library 
-- ------------------------------------------------------------------------
-- 
-- /home/pcchen/github/Cytnx/cmake/Modules
--  Generator: Unix Makefiles
--  Build Target: -
--  Installation Prefix: 
--  Version: 1.0.0
-- The CXX compiler identification is GNU 13.3.0
-- The C compiler identification is GNU 13.3.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- The CUDA compiler identification is NVIDIA 12.8.93
-- Detecting CUDA compiler ABI info
-- Detecting CUDA compiler ABI info - done
-- Check for working CUDA compiler: /usr/local/cuda/bin/nvcc - skipped
-- Detecting CUDA compile features
[...]

PR #582, opened 2025-03-18
Fix build when enabling CUDA but not cuQuantum

IvanaGyro:
Contributors usually run tests with both CUDA and cuQuantum enabled. This bug is located at the place that is disabled when the `UNI_CUQUANTUM` feature flag is enabled, so we didn't detect this bug before.


IvanaGyro:
This PR fixes the bug addressed in #580.

yingjerkao (approved):

PR #586, opened 2025-03-19
BREAKING CHANGE: Replace symmetry type with enum

IvanaGyro:
- Replaced `class SymmetryType_class` with `enum SymmetryType`
- Removed `struct __sym` (I mention here becuase C++ doesn't have name mangling like Python. Removing this structure indeed breaks the interface.)
- Removed singaleton instance `SymType`


hunghaoti:
I believe the original design intention was to allow for easy switching between Python and C++. For instance, if the Python code uses `SymType.U`, the C++ code would also use `SymType.U`.

If we change `SymType` to an `enum` , the syntax in C++ would become `SymType::U`. In my point of view it is acceptable. Since we have similar issue when using namespace in C++. For example, in Python we write `cytnx.random.uniform(...)`, and in C++ we write `cytnx::random::uniform(...)`. @kaihsin  , what do you think about it?

If we decide to use `enum`, we can consider using `enum class`.

ianmccul:
eventually you will want to make this something more general, so that you can add new symmetries later, ideally as an extension that doesn't require modifying existing code.

kaihsin:
I agree with @ianmccul . For now, yes. Let's do enum

pybind/symmetry_py.cpp
```diff
@@ -18,9 +18,9 @@ using namespace pybind11::literals;
 using namespace cytnx;
 
 void symmetry_binding(py::module &m) {
-  py::enum_<__sym::__stype>(m, "SymType")
-    .value("Z", __sym::__stype::Z)
-    .value("U", __sym::__stype::U)
+  py::enum_<SymmetryType>(m, "SymType")
```
hunghaoti, line 21: I think we should use "SymmetryType" here rather than "SymType".

hunghaoti (approved):

hunghaoti:
I suggest that in the future we can consider replacing `enum` with `enum class`.

pybind/symmetry_py.cpp
IvanaGyro: That will break the Python API. I can make SymType deprecated now, but it will not be removed until version 2.x.

IvanaGyro:
I thought to merge this PR after changing `SymType` to `SymmetryType` and mark `SymType` deprecated. This job was pending because I set updating build process and supporting Mac at higher priority.

We can merge this PR now and add the new Python API in the future.

PR #589, opened 2025-03-21
Update CITATION.cff

IvanaGyro:
- At least one author must be added.
- Updated the version number and the release date.


hunghaoti:
We have the published paper: 	
https://doi.org/10.21468/SciPostPhysCodeb.53
So I think it will be great if we can use this link.

IvanaGyro:
@hunghaoti I updated it.

hunghaoti (approved):

PR #597, opened 2025-05-05
Fermions

manuschneider:
I added the basic functionality for fermionic UniTensors. I followed [these ideas](https://doi.org/10.21468/SciPostPhysCodeb.20) quite closely, but with a more efficient implementation of the sign flips. See more discussion and conventions on [overleaf](https://www.overleaf.com/project/66861014a28ff7e8d22e953c).
 
The main additions are:
- Symmetries `fermionParity` and `fermionNumber`: they are basically a Z2 and a U(1) Symmetries, but also contain information about the fermion parity of each quantum number (even/odd) when `get_fparity()` is called.
- UniTensor class `BlockFermionicUniTensor`: basically a copy of `BlockUniTensor`, but containing the sign flips corresponding to fermionic anti-commutation relations; `_signflip` contains the information if a given block comes with a negative sign. When indices are permuted, these signs change. The actual sign flip is only applied in linear algebra functions like contractions etc.
- Contractions, basic linear algebra functions, SVD can now all deal with the sign flips correctly

Todos (help appreciated):
- unit tests
- `Svd_truncate` needs to be finished, eigenvalue decomposition, hermitian conjugation
- `P-gates` or similar structures for linear operators and scalar products, [see the graded spaces paper](https://doi.org/10.21468/SciPostPhys.18.1.012)
- documentation (user guide)


include/Symmetry.hpp
```diff
@@ -35,12 +31,25 @@ namespace cytnx {
...
    */
   extern SymmetryType_class SymType;
 
+  /**
+   * @brief fparity
+   * @details the parity of fermionis
+   * EVEN For bosons or an even number of fermions
+   * ODD  For an even number of fermions
+   */
+  enum fparity : bool { EVEN = false, ODD = true };
```
IvanaGyro, line 48: Need to follow the naming rule of enum.

https://google.github.io/styleguide/cppguide.html#Enumerator_Names

include/Symmetry.hpp
```diff
@@ -35,12 +31,25 @@ namespace cytnx {
...
+  // fparity operator+ (fparity const& lhs, fparity const& rhs) {
```
IvanaGyro, line 49: Should this be removed?

include/Symmetry.hpp
```diff
@@ -145,6 +160,56 @@ namespace cytnx {
...
+      return out;
+    }
+    bool check_qnum(const cytnx_int64 &in_qnum);
+    bool check_qnums(const std::vector<cytnx_int64> &in_qnums);
+    void combine_rule_(std::vector<cytnx_int64> &out, const std::vector<cytnx_int64> &inL,
+                       const std::vector<cytnx_int64> &inR);
+    void combine_rule_(cytnx_int64 &out, const cytnx_int64 &inL, const cytnx_int64 &inR,
+                       const bool &is_reverse);
+    void reverse_rule_(cytnx_int64 &out, const cytnx_int64 &in);
+    fparity get_fparity(const cytnx_int64 &in_qnum);
```
IvanaGyro, line 181: Should add override keywords for `get_fparity`, `is_fermionic`, and `stype_str`. Additionally, `virtual` keyword on `is_fermionic` should be removed.

include/Symmetry.hpp
```diff
@@ -89,7 +98,13 @@ namespace cytnx {
     virtual void combine_rule_(cytnx_int64 &out, const cytnx_int64 &inL, const cytnx_int64 &inR,
                                const bool &is_reverse);
     virtual void reverse_rule_(cytnx_int64 &out, const cytnx_int64 &in);
+    virtual fparity get_fparity(const cytnx_int64 &in_qnum);
+    virtual bool is_fermionic() { return false; };
```
IvanaGyro, line 102: Do you consider to declare these method with the `const` specifier?

pybind/symmetry_py.cpp
```diff
@@ -21,6 +21,13 @@ void symmetry_binding(py::module &m) {
   py::enum_<__sym::__stype>(m, "SymType")
     .value("Z", __sym::__stype::Z)
     .value("U", __sym::__stype::U)
+    .value("fPar", __sym::__stype::fPar)
```
IvanaGyro, line 24: Should the element of enum be capitalized and the class name of enum be camel case?

pybind/symmetry_py.cpp
```diff
@@ -53,6 +62,8 @@ void symmetry_binding(py::module &m) {
     .def(
       "reverse_rule", [](Symmetry &self, const cytnx_int64 &qin) { return self.reverse_rule(qin); },
       py::arg("qin"))
+    .def("get_fparity", &Symmetry::get_fparity, py::arg("qnum"))
```
IvanaGyro, line 65: Is using an abbreviation here a good idea? Does everyone using this project know the `f` in `fparity` is fermion?

tests/BlockFermionicUniTensor_test.h
```diff
@@ -0,0 +1,51 @@
+#ifndef _H_BlockFermionicUniTensor_test
```
IvanaGyro, line 1: Follow the rule of header guards in other files instead. Furthermore, it is not recommended to use `using namespace` statement in the header file.

https://google.github.io/styleguide/cppguide.html#The__define_Guard
https://google.github.io/styleguide/cppguide.html#Namespaces

tests/BlockUniTensor_test.cpp
```diff
@@ -1034,3 +1034,9 @@ TEST_F(BlockUniTensorTest, get_bond_ref) {
   EXPECT_EQ(ut1.bonds()[0].qnums(), std::vector<std::vector<cytnx_int64>>({{1}, {0}, {0}}));
   EXPECT_EQ(ut1.bonds()[1].qnums(), std::vector<std::vector<cytnx_int64>>({{1}, {0}, {0}}));
 }
+
+// TEST_F(BlockUniTensorTest, fparity_boson) {
+//   bool par;
+
+//   EXPECT_EQ(par, false);
+// }
```
IvanaGyro, line 1042: Remove if unused.

IvanaGyro (approved):

manuschneider:
It seems that the build process fails on github because of incompatibilities with the Storage refactoring. The signs are included on a low level, when the storage is touched. What is the new API to do this? I currently only copied the old behavior of `BlockUniTensor`.

Strangely, everything compiles and runs on my linux machine. Also, I ran `pre-commit run --all-files`, so I am wondering why the clang check fails.

IvanaGyro:
> Also, I ran pre-commit run --all-files, so I am wondering why the clang check fails.

Maybe you didn't correctly committed the files corrected by pre-commit?

> It seems that the build process fails on github because of incompatibilities with the Storage refactoring.

Here is the present one.
```
this->_blocks[a].storage()._impl->data()
```

tests/BlockFermionicUniTensor_test.h
manuschneider: I copied this from `tests/BlockUniTensor_test.h`. I would suggest to keep the same style (and that is what I am doing for now) or change it everywhere

pybind/symmetry_py.cpp
manuschneider: You mean
```
struct __sym {
    enum __stype { U = -1, Z = 0, FPAR = -2, FNUM = -3 };
  };

  class SymmetryType_class {
   public:
    enum : int { Void = -99, U = -1, Z = 0, fPar = -2, fNum = -3 };
    std::string getname(const int &stype);
  };
  ```
in `/include/Symmetry.h` (and therefore also in `pybind/symmetry_py.cpp`)? Why this convention?

pybind/symmetry_py.cpp
manuschneider: I changed it to `fermionParity` for the type and `get_fermion_parity` for the function.

PR #602, opened 2025-05-28
Fix Mamba error when running GitHub actions

IvanaGyro:
Drop boa and mamba because they almost don't speed the build process up comparing with using the built-in conda-build now. Conda uses libmamba as the default solver since 23.10, which made conda comparable to mamba. Additaionally, boa was archived in 2024 Dec. The replacement of boa is rattler-build. rattler-build may reduce some overheads of conda-build. It's worth to try in the future.

Downgrade the Python version to 3.10 to support older Python version in the future. Alought the oldest Python version under security maintaince is 3.9 now, the support will be terminated in 2025 Oct. To avoid changing the Python version too freqently, downgrade the version to 3.10 instead of 3.9.

### DIFF
--- a/.github/workflows/conda_build_linux.yml
+++ b/.github/workflows/conda_build_linux.yml
@@ -18,44 +18,26 @@ jobs:
     steps:
     - uses: actions/checkout@v3
 
-    - uses: conda-incubator/setup-miniconda@v2
+    - uses: conda-incubator/setup-miniconda@v3
       with:
-        # Specify python version your environment will have. Remember to quote this, or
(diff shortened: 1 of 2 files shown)


IvanaGyro:
> It seems that the tests fail for MacOS because no MKL is available on arm64 build? I think one easy fix is to turn off USE_MKL switch in Mac build.

Yes. To turn off MKL, we may have to modify setup.py. I plan to fix it after finishing the work of caching the build result and stopping supporting Apple Clang.

Issue #595, opened 2025-04-28
conda install cytnx MKL problem.


Heliumky:
I followed the documentation to install Cytnx version 0.9.7.
When I **do not** import the `cytnx` package, all NumPy-dependent code runs correctly, as shown below:

(cytnx) [jerrychen@cn7 2d_new]$ python3 test_bug.py 
(1.0000000000000004+0j)

However, **after importing** cytnx, the code triggers an Intel MKL-related error:

(cytnx) [jerrychen@cn7 2d_new]$ python3 test_bug.py 

Intel oneMKL ERROR: Parameter 3 was incorrect on entry to ZHEEVD.

Intel oneMKL ERROR: Parameter 3 was incorrect on entry to ZHEEVD.

Intel oneMKL ERROR: Parameter 3 was incorrect on entry to ZHEEVD.

Intel oneMKL ERROR: Parameter 3 was incorrect on entry to ZHEEVD.

Intel oneMKL ERROR: Parameter 3 was incorrect on entry to ZHEEVD.

Intel oneMKL ERROR: Parameter 3 was incorrect on entry to ZHEEVD.

Intel oneMKL ERROR: Parameter 3 was incorrect on entry to ZHEEVD.

Intel oneMKL ERROR: Parameter 3 was incorrect on entry to ZHEEVD.

Intel oneMKL ERROR: Parameter 3 was incorrect on entry to ZHEEVD.

Intel oneMKL ERROR: Parameter 3 was incorrect on entry to ZHEEVD.

Intel oneMKL ERROR: Parameter 3 was incorrect on entry to ZHEEVD.

Intel oneMKL ERROR: Parameter 3 was incorrect on entry to ZHEEVD.
Traceback (most recent call last):
  File "/data2/home/jerrychen/Jerry_work/2d_new/test_bug.py", line 130, in <module>
    print(inner_MPS(psi0,psi0))
          ^^^^^^^^^^^^^^^^^^^^
  File "/data2/home/jerrychen/Jerry_work/2d_new/test_bug.py", line 12, in inner_MPS
    assert len(res) == 1
           ^^^^^^^^^^^^^
AssertionError

Since GitHub does not allow attaching `.py` files directly, I have renamed the uploaded script with a different extension.
Please kindly change the filename extension back to `.py` when testing.

Thank you very much for your help!
[test_bug.txt](https://github.com/user-attachments/files/19949556/test_bug.txt)

yingjerkao:
[...]

PR #605, opened 2025-06-02
Refactor MKL integration to use LP64 interface

yingjerkao:
This PR changes default binary build to MKL_LP64 for better compatibility with `numpy` and `spicy`. 
Also, usage of oneAPI cmake configuration for better integration.


CytnxBKNDCMakeLists.cmake
```diff
@@ -1,26 +1,31 @@
...
-  target_compile_definitions(cytnx PUBLIC MKL_ILP64)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fopenmp")
-endif() #use_mkl
+# if (USE_MKL)
+#   option(MKL_SDL "Link to a single MKL dynamic libary." ON)
+#   option(MKL_MLT "Use multi-threading libary. [Default]" ON)
+#   mark_as_advanced(MKL_SDL MKL_MLT)
+#   target_compile_definitions(cytnx PUBLIC UNI_MKL)
+#   target_compile_definitions(cytnx PUBLIC MKL_LP64)
+#   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fopenmp")
```
IvanaGyro, line 7: I remember that `-fopenmp` is needed to make the program linking against to Cytnx successfully find the implementation of openmp functions. I don't know why.

src/backend/linalg_internal_interface.cpp
```diff
@@ -30,7 +30,7 @@ namespace cytnx {
     linalg_internal_interface::linalg_internal_interface() {
       mkl_code = -1;
 #ifdef UNI_MKL
-      this->set_mkl_ilp64();
+      this->set_mkl_lp64();
```
IvanaGyro, line 33: If Cytnx is built with mkl_lp64 instead of mkl_rt, do we stiil have to set the variant of MKL manually?
Another similar but different question. If Cytnx is linked against with mkl_rt, do we have to manually set the variant of MKL to be compatible with numpy or scipy which also link against mkl_rt on conda?

IvanaGyro:
Is it okay to replace the cmake module for MKL with these two [cmake commands](https://cmake.org/cmake/help/latest/module/FindBLAS.html)?

```cmake
set(BLA_VENDOR Intel10_64_dyn)
find_package( BLAS REQUIRED)
```

src/backend/linalg_internal_interface.cpp
yingjerkao: It seems that the original  is hard coded setting mkl_ilp64 regardless of the variant.

IvanaGyro: The output from https://github.com/Cytnx-dev/Cytnx/issues/595#issuecomment-2841947115 confirms current Cytnx on Conda is linked against with mkl_rt. 

> libpython3.11.so.1.0 => not  (real: /home/jerrychen/Downloads/QTTHydrogen-main/onebody/hydrogen/2d_new/not)
libmkl_rt.so.2 => /opt/intel/oneapi/mkl/2025.0/lib/libmkl_rt.so.2  (real: /opt/intel/oneapi/mkl/2025.0/lib/libmkl_rt.so.2)
libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6  (real: /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.33)
libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6  (real: /usr/lib/x86_64-linux-gnu/libm.so.6)
libgomp.so.1 => /lib/x86_64-linux-gnu/libgomp.so.1  (real: /usr/lib/x86_64-linux-gnu/libgomp.so.1.0.0)
libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1  (real: /usr/lib/x86_64-linux-gnu/libgcc_s.so.1)
libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6  (real: /usr/lib/x86_64-linux-gnu/libc.so.6)
libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2  (real: /usr/lib/x86_64-linux-gnu/libdl.so.2)

I think this is the reason we have to switch to MKL ILP64 at runtime. Does the new practice link against mkl_rt or the mkl_lp64 family directly?

yingjerkao: I have changed it back to link against libmkl_rt

yingjerkao:
> The unit test for the Arnoldi algorithm failed because `cytnx::linalg::Vectordot` returned an incorrect result, which used in Arnoldi implementation.
> 
> https://github.com/Cytnx-dev/Cytnx/blob/9debc5feae93a3f375bc5c2b8173b1862fad9e31/src/linalg/Arnoldi.cpp#L165
> 
> 
> The output of Vectordot is wrong when we call lapack function here:
> https://github.com/Cytnx-dev/Cytnx/blob/9debc5feae93a3f375bc5c2b8173b1862fad9e31/src/backend/linalg_internal_cpu/Vectordot_internal.cpp#L11-L50
> 
> 
> Since this PR change to LP64, so sizeof(blas_int) != 8

I wonder why the implementation relies on the explicit size of `blas_int`?

yingjerkao:
According to this https://www.intel.com/content/www/us/en/docs/onemkl/developer-guide-linux/2023-0/using-the-ilp64-interface-vs-lp64-interface.html

`MKL_INT` is 4 bytes in lp64 and the behavior should be the same as openblas. 

`blas_int` is defined here 
https://github.com/Cytnx-dev/Cytnx/blob/18670bdeb83cd57f2a1a3389370e55d8279e6e7a/include/Type.hpp#L23-L29


So I don't understand why there should be an error for MKL lp64 but not for openblas?

CytnxBKNDCMakeLists.cmake
yingjerkao: We are not using openmp any more right?

hunghaoti:
> I looked at the output messages of these tests and they have the same error
> 
> ```
> # Cytnx error occur at void cytnx::linalg_internal::InvM_inplace_internal_cd(boost::intrusive_ptr<Storage_base> &, const cytnx_int64 &)
> # error: Error in Lapack function 'zgetrf': Lapack INFO =  2
> # file : /home/yjkao/Cytnx/src/backend/linalg_internal_cpu/InvM_inplace_internal.cpp (43)
> ```
> 
> According to the LAPACK documentation: https://www.netlib.org/lapack/explore-html/db/d04/group__getrf_ga85d0f9ae367fd540a745f558a77717db.html
> 
> This indicates the resulting `U` is singular so that the inversion is also singular. Is this still related to `VectorDot` ?



> I looked at the output messages of these tests and they have the same error
> 
> ```
> # Cytnx error occur at void cytnx::linalg_internal::InvM_inplace_internal_cd(boost::intrusive_ptr<Storage_base> &, const cytnx_int64 &)
> # error: Error in Lapack function 'zgetrf': Lapack INFO =  2
> # file : /home/yjkao/Cytnx/src/backend/linalg_internal_cpu/InvM_inplace_internal.cpp (43)
> ```
> 
> According to the LAPACK documentation: https://www.netlib.org/lapack/explore-html/db/d04/group__getrf_ga85d0f9ae367fd540a745f558a77717db.html
> 
> This indicates the resulting `U` is singular so that the inversion is also singular. Is this still related to `VectorDot` ?

Yes, this is the issue form `Vectordot`. Since currently the output of `zdotc` not work, so the output of `Vectordot` is zero, which cause the output of matrix is just a zero matrix so is not invertible.
If you replace 
[Vectordot_internal.cpp#L11-L50](https://github.com/Cytnx-dev/Cytnx/blob/9debc5feae93a3f375bc5c2b8173b1862fad9e31/src/backend/linalg_internal_cpu/Vectordot_internal.cpp#L11-L50)
with 
```c++
      _out[0] = 0;
      if (is_conj) {
        for (cytnx_uint64 i = 0; i < len; i++) {
          _out[0] += std::conj(_Lin[i]) * _Rin[i];
        }
      } else {
        for (cytnx_uint64 i = 0; i < len; i++) {
          _out[0] += _Lin[i] * _Rin[i];
        }
      }
```
then it can work.

ianmccul:
presumably it is not directly a 32/64 bit issue, but the same problem as described at https://github.com/flame/blis/issues/433

Intel fortran calling convention uses a hidden first parameter for complex return types (as the code above assumes), but gfortran uses registers (same as C calling convention).  This means that MKL use a different convention to other BLAS libraries.  I think nvidia use Intel convention as well.

ianmccul:
If you want to use `zdotc` you need to have a configuration option to switch between intel and gcc style.  Eg in the MPToolkit I have:
```
#if defined(FORTRAN_COMPLEX_RETURN_FIRST_ARG)
void F77NAME(zdotu)(std::complex<double>* result, const integer *n,
                    const std::complex<double> *dx, const integer *incx,
                    const std::complex<double> *dy, const integer *incy);

void F77NAME(zdotc)(std::complex<double>* result, const integer *n,
                    const std::complex<double> *dx, const integer *incx,
                    const std::complex<double> *dy, const integer *incy);
#elif defined(FORTRAN_COMPLEX_RETURN_IN_REGISTER)
Fortran::complex
F77NAME(zdotu)(const integer *n,
               const std::complex<double> *dx, const integer *incx,
               const std::complex<double> *dy, const integer *incy);
Fortran::complex
F77NAME(zdotc)(const integer *n,
               const std::complex<double> *dx, const integer *incx,
               const std::complex<double> *dy, const integer *incy);
#endif
```
with an autoconf check to see which version works!

ianmccul:
and the other half of the code is
```
#if defined(FORTRAN_COMPLEX_RETURN_FIRST_ARG)
inline
std::complex<double>
zdotu(integer n, std::complex<double> const* dx, integer incx,
      std::complex<double> const* dy, integer incy)
{
   std::complex<double> result;
   TRACE_BLAS1("BLAS1: zdotu")(&result)(n)(dx)(incx)(dy)(incy);
   raw::F77NAME(zdotu)(&result, &n, dx, &incx, dy, &incy);
   return result;
}

inline
std::complex<double>
zdotc(integer n, std::complex<double> const* dx, integer incx,
      std::complex<double> const* dy, integer incy)
{
   std::complex<double> result;
   TRACE_BLAS1("BLAS1: zdotc")(&result)(n)(dx)(incx)(dy)(incy);
   raw::F77NAME(zdotc)(&result, &n, dx, &incx, dy, &incy);
   return result;
}
#elif defined(FORTRAN_COMPLEX_RETURN_IN_REGISTER)
inline
std::complex<double>
zdotu(integer n, std::complex<double> const* dx, integer incx,
      std::complex<double> const* dy, integer incy)
{
   TRACE_BLAS1("BLAS1: zdotu")(n)(dx)(incx)(dy)(incy);
   return raw::F77NAME(zdotu)(&n, dx, &incx, dy, &incy);
}

inline
std::complex<double>
zdotc(integer n, std::complex<double> const* dx, integer incx,
      std::complex<double> const* dy, integer incy)
{
   TRACE_BLAS1("BLAS1: zdotc")(&result)(n)(dx)(incx)(dy)(incy);
   return raw::F77NAME(zdotc)(&n, dx, &incx, dy, &incy);
}
#endif
```

ianmccul:
> The speed is 4-5 times better if we call zdotc(u)
> 

That is a bit surprising.  Make sure that the loop condition is a local variable, not a reference, and make sure that you accumulate into a local variable.  i.e.
```
      cytnx_complex128 r = 0.0;
      auto L = len;  // or pass len by value
      if (is_conj) {
        for (cytnx_uint64 i = 0; i < L; i++) {
          r += std::conj(_Lin[i]) * _Rin[i];
        }
      } else {
        for (cytnx_uint64 i = 0; i < L; i++) {
          r += _Lin[i] * _Rin[i];
        }
      }
      _out[0] = r;
 ```
 Clang should autovectorize a loop like this quite well.  Gcc might need some coaxing to vectorize it, or use SIMD intrinsics.  It won't autovectorize unless it can prove that there is no aliasing (i.e. that _out[0] is not part of _Lin or _Rin arrays [which should be renamed to not use _X pattern too]).

**Comment by hunghaoti** [comment]
> > The speed is 4-5 times better if we call zdotc(u)
> 
> That is a bit surprising. Make sure that the loop condition is a local variable, not a reference, and make sure that you accumulate into a local variable. i.e.
> 
> ```
>       cytnx_complex128 r = 0.0;
>       auto L = len;  // or pass len by value
>       if (is_conj) {
>         for (cytnx_uint64 i = 0; i < L; i++) {
>           r += std::conj(_Lin[i]) * _Rin[i];
>         }
>       } else {
>         for (cytnx_uint64 i = 0; i < L; i++) {
>           r += _Lin[i] * _Rin[i];
>         }
>       }
>       _out[0] = r;
> ```
> 
> Clang should autovectorize a loop like this quite well. Gcc might need some coaxing to vectorize it, or use SIMD intrinsics. It won't autovectorize unless it can prove that there is no aliasing (i.e. that _out[0] is not part of _Lin or _Rin arrays [which should be renamed to not use _X pattern too]).

Thanks for the suggestion, this modification is better:
| test case | original | this modification |
|---|---|---|
| complex double type with conjugate, len= $10^5$ |0.65 ms | 0.496 ms |
| complex double type with conjugate len= $10^7$ | 65 ms | 49.6 ms |
| complex double type without conjugate, len= $10^5$ |0.65 ms | 0.496 ms |
| complex double type without conjugate len= $10^7$ | 64.8 ms | 50.4 ms |

**Comment by hunghaoti** [comment]
> that is still slower than I'd hoped. If you really want to avoid using `zdotc`/`zdotu` it might be worth spending some time getting a faster C++ version, maybe use compiler explorer to see what code it is generating, probably some compiler flags to force it to autovectorize would help. Writing a version using AVX intrinsics is also possible.

Actually in this branch `601-use-mcl_lp64-for-default-binary-build` have compatible efficiency (both use C++ version rather than call zdotc):
| test case | master branch (C++)| branch:601-use-mcl_lp64-for-default-binary-build (C++)| master branch call blas |
|---|---|---|---|
| complex double type with conjugate, len= $10^5$ | 0.496 ms | 0.109 ms | 0.119 ms |
| complex double type with conjugate len= $10^7$ |  49.6 ms | 15.4 ms | 15.2 ms |
| complex double type without conjugate, len= $10^5$ | 0.496 ms | 0.099 ms |0.119 ms |
| complex double type without conjugate len= $10^7$ |  50.4 ms | 15.1 ms |15.2 ms |

**Comment by ianmccul** [comment]
> Actually in this branch `601-use-mcl_lp64-for-default-binary-build` have compatible efficiency (both use C++ version rather than call zdotc):

why the difference?

I tried some test code of my own, a version using C++ `std::complex`, a version using C++ but splitting into real/imag arrays, and a version using std::experimental::simd.  The simd vectorized version is faster for small arrays, but for large arrays it is probably memory bandwidth limited so all versions give basically the same timing. With `N = 10'000'000`:
```
$ ./zdotc
naive std::complex: 16.733092700000000264 ms, result = (1487.6978129023566453,1786.8840666525102279)
split scalar: 14.830114000000000019 ms, result = (1487.6978129023566453,1786.8840666524979497)
split SIMD: 14.890398999999996832 ms, result = (1487.6978129026383613,1786.8840666529004011)
```

With `N = 10'000`:
```
$ ./zdotc
naive std::complex: 0.021824244000000009996 ms, result = (-44.102521477578463305,31.734692406445805801)
split scalar: 0.012260067999999979885 ms, result = (-44.102521477578463305,31.734692406445766721)
split SIMD: 0.0032028689999999970794 ms, result = (-44.102521477578591202,31.734692406445841328)
```

So SIMD is probably useful just for the smaller array sizes that are in cache.  I haven't tried a SIMD version using `std::complex`.  It is curious that splitting the real and imag arrays gives a pretty good improvement for smaller arrays though, I am not sure what to make of that, I wasn't expecting it.  Maybe reading from 4 different arrays uses memory bandwidth better than reading from 2 arrays?  Actually it is probably my compiler; my local machine is gcc 12.3.0, on Godbolt gcc 15.1 gives:
```
naive std::complex: 0.018023076000000064906 ms, result = (-44.102521477578463305,31.734692406445805801)
split scalar: 0.01544372400000001283 ms, result = (-44.102521477578463305,31.734692406445805801)
split SIMD: 0.0086486810000000018117 ms, result = (-44.102521477578591202,31.734692406445841328)
```

Different hardware so not direcly comparable, but the `std::complex` and the split versions are pretty close to the same speed, and only 2x slower than SIMD, rather than ~ 4 - 6x slower.  (Godbolt code is at https://godbolt.org/z/sPP4qx6oz)

**Comment by yingjerkao** [comment]
But the current implementation  of calling `zdotc` is passing the output variable as the first parameter

https://github.com/Cytnx-dev/Cytnx/blob/9debc5feae93a3f375bc5c2b8173b1862fad9e31/src/backend/linalg_internal_cpu/Vectordot_internal.cpp#L11-L50

which is what Intel suggested in this example

https://www.intel.com/content/www/us/en/docs/onemkl/developer-guide-linux/2025-0/call-blas-funcs-return-complex-values-in-c-code.html


> If you want to use `zdotc` you need to have a configuration option to switch between intel and gcc style. Eg in the MPToolkit I have:
> 
> ```
> #if defined(FORTRAN_COMPLEX_RETURN_FIRST_ARG)
> void F77NAME(zdotu)(std::complex<double>* result, const integer *n,
>                     const std::complex<double> *dx, const integer *incx,
>                     const std::complex<double> *dy, const integer *incy);
> 
> void F77NAME(zdotc)(std::complex<double>* result, const integer *n,
>                     const std::complex<double> *dx, const integer *incx,
>                     const std::complex<double> *dy, const integer *incy);
> #elif defined(FORTRAN_COMPLEX_RETURN_IN_REGISTER)
> Fortran::complex
> F77NAME(zdotu)(const integer *n,
>                const std::complex<double> *dx, const integer *incx,
>                const std::complex<double> *dy, const integer *incy);
> Fortran::complex
> F77NAME(zdotc)(const integer *n,
>                const std::complex<double> *dx, const integer *incx,
>                const std::complex<double> *dy, const integer *incy);
> #endif
> ```
> 
> with an autoconf check to see which version works!

Thanks. 

However, we use `UNI_MKL` to switch between MKL and other BLAS version

https://github.com/Cytnx-dev/Cytnx/blob/18670bdeb83cd57f2a1a3389370e55d8279e6e7a/include/Type.hpp#L14-L30

So should work as in your case

**Comment by yingjerkao** [comment]
@yingjerkao  @pcchen  can you help me confirm that the isolated code is correct?

**Comment by hunghaoti** [comment]
> @yingjerkao @pcchen can you help me confirm that the isolated code is correct?

Hi @yingjerkao, I have just found that if you add the following code 
under line 34 in the file `CytnxBKNDCMakeLists.cmake`
```Cmake
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -lmkl_intel_lp64 -lmkl_gnu_thread -lmkl_core -liomp5")
```
https://github.com/Cytnx-dev/Cytnx/blob/9debc5feae93a3f375bc5c2b8173b1862fad9e31/CytnxBKNDCMakeLists.cmake#L31-L35
The the result will be correct. I am writing the unit test and benchmark code for `Vectordot` and then I will commit them.

IvanaGyro:
> > @yingjerkao @pcchen can you help me confirm that the isolated code is correct?
> 
> Hi @yingjerkao, I have just found that if you add the following code under line 34 in the file `CytnxBKNDCMakeLists.cmake`
> 
> ```cmake
> set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -lmkl_intel_lp64 -lmkl_gnu_thread -lmkl_core -liomp5")
> ```
> 
> https://github.com/Cytnx-dev/Cytnx/blob/9debc5feae93a3f375bc5c2b8173b1862fad9e31/CytnxBKNDCMakeLists.cmake#L31-L35
> 
> The the result will be correct. I am writing the unit test and benchmark code for `Vectordot` and then I will commit them.

Those compiler flags seem link lp64 MKL lib manually. Doesn't [FindBLAS](https://cmake.org/cmake/help/latest/module/FindBLAS.html#input-variables) CMake method do that for us? Are the flags included in the `BLAS_LINKER_FLAGS` variable after calling FindBLAS?

hunghaoti:
> > ```cmake
> > set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -lmkl_intel_lp64 -lmkl_gnu_thread -lmkl_core -liomp5")
> > ```
> 
> This links to the dynamical library , but I believe we are linking to single dynamical runtime library `mkl_rt`.

We don't need to add this line in `CytnxBKNDCMakeLists.cmake`. I discovered the code works by simply export `MKL_INTERFACE_LAYER` is either empty or set to LP64, rather than `MKL_INTERFACE_LAYER=LP64,GNU` value as default `cytnx` environment.

yingjerkao:
> > > ```cmake
> > > set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -lmkl_intel_lp64 -lmkl_gnu_thread -lmkl_core -liomp5")
> > > ```
> > 
> > 
> > This links to the dynamical library , but I believe we are linking to single dynamical runtime library `mkl_rt`.
> 
> We don't need to add this line in `CytnxBKNDCMakeLists.cmake`. I discovered the code works by simply export `MKL_INTERFACE_LAYER` is either empty or set to LP64, rather than `MKL_INTERFACE_LAYER=LP64,GNU` value as default `cytnx` environment.

Where is this environment set? 

Also, `MKL_INTERFACE_LAYER` is set in `CytnxBKNDCMakeLists.cmake` line 17

hunghaoti:
> > We don't need to add this line in `CytnxBKNDCMakeLists.cmake`. I discovered the code works by simply export `MKL_INTERFACE_LAYER` is either empty or set to LP64, rather than `MKL_INTERFACE_LAYER=LP64,GNU` value as default `cytnx` environment.
> 
> Where is this environment set?
> 
> Also, `MKL_INTERFACE_LAYER` is set in `CytnxBKNDCMakeLists.cmake` line 17

In my case, there is a script file in 
```
/home/hunghaoti/anaconda3/envs/cytnx/etc/conda/activate.d
```
The file content:
```bash
export CONDA_MKL_INTERFACE_LAYER_BACKUP=${MKL_INTERFACE_LAYER-}
export MKL_INTERFACE_LAYER=LP64,GNU

```
So the environment variable always be set if I activate the cytnx environment. Moreover, this file will be created after I install mkl using
```bash
conda install libblas=*=*mkl
```
If I just set `MKL_INTERFACE_LAYER` using
```bash
export=MKL_INTERFACE_LAYER=64
```
before running the test, then it work correctly.

yingjerkao:
Here is the relevant link 
https://www.intel.com/content/www/us/en/docs/onemkl/developer-guide-linux/2023-0/dynamic-select-the-interface-and-threading-layer.html

I find it odd that we can not do this in CMake. Intel One API provides FindMKL cmake files which might provides a better control over how to choose the right MKL.

cytnx/__init__.py
```diff
@@ -10,12 +10,21 @@
...
+    # Detect and set MKL interface
+    mkl_interface = detect_mkl_interface()
+    print(f"MKL interface: {mkl_interface}")
+    if mkl_interface == 2:  # ILP64
+        if ('numpy' in sys.modules) or ('scipy' in sys.modules):
+            warnings.warn("numpy and/or scipy are imported before cytnx. Please make sure it supports ILP64.")
+
+        set_mkl_ilp64()
+    elif mkl_interface == 1:  # LP64
+        set_mkl_lp64()
```
IvanaGyro, line 24: I am not sure what is the root problem now. Can the problem be solved if we always call `set_mkl_lp64()`?

IvanaGyro: > Here is the relevant link
https://www.intel.com/content/www/us/en/docs/onemkl/developer-guide-linux/2023-0/dynamic-select-the-interface-and-threading-layer.html

I think we call those function exactly in this PR. Am I right?

yingjerkao: I didn't apply this to the C++ code. That wrapper functions were intended for Python.

yingjerkao:
Somehow the tests passed under Rocky Linux 9.5 and g++ (GCC) 11.5.0

kaihsin:
The reason for mkl_rt is for the compatibility with numpy. The original intention was that ilp64 support 64bit int while lp64 default with 32bit for the dimension size. 

as you probably already discovered. numpy only support lp64. This make it limited. So the solution to make C++ side using ilp64 while keeping it inter play with numpy well, is to link with mkl_rt, so the library can selectional link to either ilp64 or lp64 depending on whether numpy is loaded or not in python 

If there is no need to support int64 linalg API even on the C++ side, then there is no need to deal with mkl_rt hassle

hunghaoti (approved):

PR #614, opened 2025-06-13
Fixed Gesvd and Svd_truncate with fermions

manuschneider:
`Gesvd` was buggy if used with fermions, and `Svd_truncate` was disabled for fermions. Both works now. I had to add the method `signflip_()` to `UniTensor` to allow for deletion of signflip entries if blocks get erased in `Svd_truncate`

### DIFF
--- a/include/UniTensor.hpp
+++ b/include/UniTensor.hpp
@@ -226,6 +226,7 @@ namespace cytnx {
                                 const cytnx_int64 &rowrank = -1, const std::string &name = "");
     virtual std::vector<cytnx_uint64> shape() const;
     virtual std::vector<bool> signflip() const;
+    virtual std::vector<bool> &signflip_();
     virtual bool is_blockform() const;
     virtual bool is_contiguous() const;
     virtual void to_(const int &device);
@@ -1872,9 +1873,8 @@ namespace cytnx {
       return this->_blocks.size();
     };
 
-    std::vector<bool> signflip() const;
-    // std::vector<bool> &signflip_reference() { return this->_signflip; };
-    // const std::vector<bool> &signflip_reference() const { return this->_signflip; };
+    std::vector<bool> signflip() const override { return this->_signflip; };
+    std::vector<bool> &signflip_() override { return this->_signflip; };
 
     void to_(const int &device) {
       //[21 Aug 2024] This is a copy from BlockUniTensor;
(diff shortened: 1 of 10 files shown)


pybind/unitensor_py.cpp
```diff
@@ -559,6 +559,7 @@ void unitensor_binding(py::module &m) {
     .def("bond", [](UniTensor &self, const std::string &label){return self.bond(label);} ,py::arg("label"))
     .def("shape", &UniTensor::shape)
     .def("signflip", &UniTensor::signflip)
+    .def("signflip_", &UniTensor::signflip_)
```
IvanaGyro, line 562: Tailing with "_" means internal method in Python.

IvanaGyro:
Is the new feature (`fermion_twist`) apart of  this bug fix?

manuschneider:
> Is the new feature (`fermion_twist`) apart of this bug fix? If not, could you create another PR for the new feature? That will be easier for reviewing.

`fermion_twist` is a new feature, but it has only a few lines of code. I do not know how to create a new PR now, can I undo the push of the changes?

IvanaGyro:
Yes, you can move the branch tag backward to the latest commit related to this PR, and use [rebase onto](https://git-scm.com/docs/git-rebase) to move the commit related to the new features to a new branch.

include/UniTensor.hpp
```diff
@@ -3116,6 +3116,16 @@ namespace cytnx {
...
 
+    /**
+     * @brief Get reference to the sign information of a fermionic UniTensor.
+     * @details Length is the number of blocks in the UniTensor. If the return is true, the sign of
+     * the elements of the corresponding block needs to be flipped.
+     * @warning This is an inline version which returns a reference. Changes to the reference affect
+     * the tensor!
+     * @return std::vector<bool> &
+     */
+    std::vector<bool> &signflip_() { return this->_impl->signflip_(); }
```
IvanaGyro, line 3127: Where is the definition of signflip_() for ferminoin unitensor?

pybind/unitensor_py.cpp
IvanaGyro, line 562: Do we really want the Python users to change the tensor by changing signflip? Is the operation safe (keep the tensor valid)? Is implementing a setter a better option?

include/UniTensor.hpp
manuschneider: https://github.com/Cytnx-dev/Cytnx/blob/aa9d656e0991713212b201c6301bc0dbb3f81a5a/include/UniTensor.hpp#L1877 This was implemented previously already

pybind/unitensor_py.cpp
manuschneider: I remember why it did not work to only have the implementation know about `signflip_()`: when calling `UniTensor._impl->signflip_()`, the `_impl` is a pointer to type `UniTensor_base`. Therefor, it would call the method of that class, which is not the fermionic implementation. Unless we want to deal with pointer casting, we should make `signflip_()` a method of `UniTensor` as well.
I will simply remove the pybind to it. The API documention on the C++ side contains a warning already.

IvanaGyro (approved):

Issue #607, opened 2025-06-03
Determine one license


IvanaGyro:
There are two licenses declared in this project. In `LICENSE` file, we declared the license as Apache 2.0; in `setup.py` we declared the licnese as LGPL. Which one is the correct license?

pcchen:
I would sugget to use Apache 2.0.

yingjerkao:
I think that was an old license from Uni10. We should all use Apache 2.0.

yingjerkao:
@IvanaGyro This is resolved?

IvanaGyro:
Do we need setup_macos.py?

yingjerkao:
Are we completely moving to condo_build? If so, we can remove `setup_macos.py`

IvanaGyro:
We are using scikit-build-core. As https://github.com/Cytnx-dev/Cytnx/issues/629#issuecomment-3087227199 mentioned, users can just call `pip install .` to install Cytnx on Mac.

pcchen:
I think the current `'pip install .' is quite good.

Issue #608, opened 2025-06-03
`license` argument in setup.py is deprecated.


IvanaGyro:
Here is the message from pip 25.1.1

```
SetuptoolsDeprecationWarning: License classifiers are deprecated.

          ********************************************************************************
          Please consider removing the following classifiers in favor of a SPDX license expression:

          License :: OSI Approved :: GNU Lesser General Public License v3 or later (LGPLv3+)

          See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
          ********************************************************************************
```

As mentioned in #607, we seem accidentally adopt two licenses, here are the available license identifiers from the [SPDX page](https://spdx.org/licenses/) for the two licenses.

For LGPL, there are five available identifiers

- LGPL-2.0-or-later
- LGPL-2.1-only
- LGPL-2.1-or-later
- LGPL-3.0-only
- LGPL-3.0-or-later

For Apache 2.0, the available identifier is `Apache-2.0`.

IvanaGyro:
The deprecation warning is emitted by `setup_macos.py` — line 152 has `license="GNU LGPL"` (the deprecated string-arg form) and line 156 has the `License :: OSI Approved :: GNU Lesser General Public License v3 or later (LGPLv3+)` classifier (the exact one named in pip 25.1.1's warning). `pyproject.toml` is already in compliance: it uses the SPDX form `license = "Apache-2.0"` with no `License ::` classifier.

`setup_macos.py` is the only remaining source of any LGPL / deprecated-license string on master, and it isn't referenced by any CI workflow, build script, or doc. #802 deletes it, which closes this issue together with #607.

PR #616, opened 2025-06-18
Migrates the build system from setuptools to scikit-build-core

IvanaGyro:
This change allows for a more modern and flexible build process,
especially regarding the handling of CMake projects.

The migration involves replacing `setup.py` with `pyproject.toml`
and updating the build script to use scikit-build-core's configuration
options. Custom CMake definitions are replaced with the CMake preset in
the `pyproject.toml`

The default BLAS provider is changed to OpenBLAS from MKL, which is more
widely supported accross different operating systems.


IvanaGyro:
This PR also fix #607 and #608.

kaihsin:
Have you tested against the mkl and CUDA from pypi? Would love to move away from conda if toml can work

IvanaGyro:
We haven't published Cytnx to PyPI. Additionally, pre-built NumPy and SciPy on PyPI link against OpenBLAS instead of MKL. To be consistent with the PyPI ecosystem, I don't think we will deliver Cytnx linked against MKL to PyPI by default. Delivering some variants is possible.

About CUDA, I suggest not depend on any CUDA related package on PyPI. I referred to CuPy. CuPy doesn't put any CUDA related Python package as their dependencies, which means that CuPy assumes users prepare the software required for building CuPy on their own. The other main reason that to not depend on any C/C++ or CUDA related package on PyPI is that is not easy and not clean to integrate those package into our build process.

Regarding Conda, we have automatically published Cytnx on Conda. The publish process doesn't take too much time. It is still worth to publish Cytnx to Conda to benefit the user who is still using Conda.

Developers don't have to reply on Conda to setup their build environment, even if there is a conda build recipe in our repo.

kaihsin:
are you suggesting don't release wheel at all and let client side compilation? or release wheel with statically linked mkl for selected distro?

pcchen:
> However, the whole build process on Mac's M series chip is still expected to be failed because the building problem of HPTT has not been solved. If you want to disable HPTT, maybe, you can try
> 
> ```shell
> pip install . -vv \
>   --config-settings skbuild.cmake.args=--preset=openblas-cpu \
>   --config-settings "skbuild.cmake.args=-G Unix Makefiles"
>   --config-settings "skbuild.cmake.args=-DUSE_HPTT=OFF"
> ```
> 
> But I an note sure if the compile definition passed via command argument will override the definition in the preset.

This is to report that the build still fail at the HPTT on Mac's M series.

Issue #443, opened 2024-08-01
Add the function for random SVD.


hunghaoti:
Add the function for random SVD based on the paper:
https://arxiv.org/pdf/2403.00562

manuschneider:
> FYI I believe the svd we used on CUDA side is via rsvd

How could one make the implementation of #619 faster on the GPU side by using what CUDA provides directly?

hunghaoti:
Implemented in #619 .

PR #619, opened 2025-06-24
Randomized SVD

manuschneider:
I implemented a randomized SVD, as suggested in #443. It uses the standard method, see for example `Algorithm 4.4` in https://arxiv.org/abs/0909.4061
All methods work for `Tensor` and dense `UniTensor`, with `BlockUniTensor` and `BlockFermionicUniTensor` still missing.

`Rsvd` applies the matrix several times to random Gaussian vectors. Then a `QR` decomposition is used to obtain a unitary transformation that projects onto the significant directions to find the largest singular values. Then an SVD is done in the reduced space.

`Rsvd_truncate` uses this with an additional truncation step. Several oversampling methods are available: a prefactor to the dimension to keep or a constant factor. I thought having both might be helpful in the implementation with symmetries, but not sure anymore. We can change it once symmetries are implemented.

To understand the code: I basically copied `Gesvd` and `Gesvd_truncate` and adopted them, so a diff to these methods shows the real changes.

Additionally, I wrote some unit tests for these cases and also for `Svd_truncate` and `Gesvd_truncate`. All truncated versions now check if the input tensor is diagonal.


pybind/linalg_py.cpp
```diff
@@ -45,6 +58,67 @@ void linalg_binding(py::module &m) {
...
+    },
+    py::arg("Tin"), py::arg("keepdim"), py::arg("is_U") = true, py::arg("is_vT") = true,
+    py::arg("power_iteration") = 2, py::arg("seed") = -1);
+  m_linalg.def(
+    "Rsvd",
+    [](const cytnx::UniTensor &Tin, const cytnx_uint64 &keepdim, const bool &is_U,
+       const bool &is_vT, const cytnx_uint64 &power_iteration = 0, int64_t & seed) {
+      if (seed == -1) {
+        // If user doesn't specify seed argument
+        seed = cytnx::random::__static_random_device();
```
IvanaGyro, line 79: Is generating random numbers from the `std::random_device` directly every function call a good idea? We are not doing encryption. Is there a more efficient way to generate the pseudo random seed?

src/linalg/Rsvd.cpp
```diff
@@ -0,0 +1,236 @@
...
+    #ifdef UNI_CUQUANTUM
+      #include "backend/linalg_internal_gpu/cuQuantumGeSvd_internal.hpp"
+    #endif
+  #endif
+
+namespace cytnx {
+  namespace linalg {
+    typedef Accessor ac;
+    std::vector<Tensor> Rsvd(const Tensor &Tin, const cytnx_uint64 &keepdim, const bool &is_U,
+                             const bool &is_vT, const cytnx_uint64 &power_iteration,
```
IvanaGyro, line 25: `bool` is built-in type. Just pass it by value. The same as `int`s.

src/linalg/Rsvd.cpp
```diff
@@ -0,0 +1,236 @@
...
+        return out;
+  #else
+        cytnx_error_msg(true, "[Rsvd] fatal error,%s",
+                        "try to call the gpu section without CUDA support.\n");
+        return std::vector<Tensor>();
+  #endif
+      }
+    }  // Rsvd(Tensor)
+
+    void _Rsvd_Dense_UT(std::vector<cytnx::UniTensor> &outCyT, const cytnx::UniTensor &Tin,
```
IvanaGyro, line 101: In C++ there is no convention for treating the method prepended with "_" as an internal function. C++ use accessibility control to handle the same idea. According to the coding style we applied, the function name should starts with the capital character.

https://google.github.io/styleguide/cppguide.html#Function_Names

src/linalg/Rsvd_truncate.cpp
```diff
@@ -0,0 +1,292 @@
...
+                                 const cytnx_uint64 &power_iteration, const unsigned int &seed) {
+      // DenseUniTensor:
+      cytnx_uint64 keep_dim = keepdim;
+
+      Tensor tmp = Tin.get_block_().contiguous();
+      // if(Tin.is_contiguous()) tmp = Tin.get_block_();
+      // else{ tmp = Tin.get_block(); tmp.contiguous_();}
+
+      std::vector<cytnx_uint64> tmps = tmp.shape();
+      std::vector<cytnx_int64> oldshape(tmps.begin(), tmps.end());
```
IvanaGyro, line 156: Why to copy the vector twice?

tests/linalg_test/Svd_truncate_test.cpp
```diff
@@ -0,0 +1,281 @@
...
+    int size = 1;
+    std::vector<Bond> bonds = {Bond(size), Bond(size), Bond(size)};
+    int rowrank = 1;
+    bool is_diag = false;
+    auto labels = std::vector<std::string>();
+    auto T = UniTensor(bonds, labels, rowrank, cytnx::Type.Double, cytnx::Device.cpu, is_diag);
+    random::Make_uniform(T, -10, 0, 0);
+    std::vector<UniTensor> Svds = linalg::Svd_truncate(T, 1);
+    EXPECT_TRUE(CheckLabels(T, Svds)) << fail_msg.TraceFailMsgs();
+    EXPECT_TRUE(ReComposeCheck(T, Svds)) << fail_msg.TraceFailMsgs();
```
IvanaGyro, line 42: Checking a method returning bool will let a error message a little useless when the test fails. Do you consider using [::testing::AssertionResult](https://google.github.io/googletest/advanced.html#using-a-function-that-returns-an-assertionresult) or [custom matcher](https://google.github.io/googletest/reference/matchers.html)?

Here is an example of using `AssertionResult`: https://github.com/Cytnx-dev/Cytnx/blob/18670bdeb83cd57f2a1a3389370e55d8279e6e7a/tests/test_tools.h#L90-L109

IvanaGyro: By using the built-in mechanism, we don't have to maintain `fail_msg` manually.

tests/utils/rsvd_reference_tensors.ipynb
```diff
@@ -0,0 +1,150 @@
+{
```
IvanaGyro, line 1: Why to put a notebook here? This folder only contains C++ code before.

pybind/linalg_py.cpp
ianmccul: maybe just increment a global std::atomic<uint64_t> for the seed?

pybind/linalg_py.cpp
```diff
@@ -23,7 +23,20 @@ using namespace cytnx;
...
+    "Rand_isometry",
+    [](const Tensor &Tin, const cytnx_uint64 &keepdim, const cytnx_uint64 &power_iteration,
+       int64_t &seed) {
+      if (seed == -1) {
+        // If user doesn't specify seed argument
+        seed = cytnx::random::__static_random_device();
+      }
+      return cytnx::linalg::Rand_isometry(Tin, keepdim, power_iteration, seed);
+    },
+    py::arg("Tin"), py::arg("keepdim"), py::arg("power_iteration") = 2, py::arg("seed") = -1);
```
ianmccul, line 38: Where is the oversampling parameter set?  Default power_iteration is 2, where does this come from?

src/linalg/Rsvd.cpp
ianmccul: And underscore followed by capital letter is reserved; user code must not use such identifiers.

src/linalg/Rsvd.cpp
```diff
@@ -0,0 +1,236 @@
...
+                      "\n");
+      cytnx_error_msg(keepdim < 1, "[ERROR][Rsvd] Keepdim must be > 0, but is %d.\n", keepdim);
+
+      Tensor in = Tin.contiguous();
+      if (Tin.dtype() > Type.Float) in = in.astype(Type.Double);
+
+      // form isometry Q[0] and apply Q.Dagger * in
+      Tensor Q = Rand_isometry(Tin, keepdim, power_iteration, seed);
+      in = Matmul(Q.Conj().permute_({1, 0}), in);
+
```
ianmccul, line 38: ahh, there is no provision for oversampling?  Oversampling is more important than using the power method.  It is also cheaper. A reasonable default might be p=10 oversampling, with q=0.  Without oversampling, the singular values at the tail of the spectrum will be very poor, even with q > 0

src/linalg/Rsvd_truncate.cpp
```diff
@@ -0,0 +1,292 @@
...
+  namespace linalg {
+    typedef Accessor ac;
+    std::vector<Tensor> Rsvd_truncate(const Tensor &Tin, const cytnx_uint64 &keepdim,
+                                      const double &err, const bool &is_U, const bool &is_vT,
+                                      const unsigned int &return_err, const cytnx_uint64 &mindim,
+                                      const cytnx_uint64 &oversampling_summand,
+                                      const double &oversampling_factor,
+                                      const cytnx_uint64 &power_iteration,
+                                      const unsigned int &seed) {
+      // check input arguments
```
ianmccul, line 31: Ahh, now we have some oversampling, great!  The version with no oversampling probably shouldn't exist then, you get same same effect by simply setting the oversampling to 0.

pybind/linalg_py.cpp
manuschneider: The oversampling is only used for `Rsvd_truncate`. The usual `Rsvd` keeps all values, and `Rand_isometry` does not do an SVD (so oversampling makes no sense here).

I decided to set `power_iteration` to 2, because I did some numerical checks `power_iteration`  set to 0 and the singular values were not that accurate (note that no oversampling is used here!). An inexperienced user would see this as a bug, and an experienced user can set it to a different value if faster execution or more accuracy are needed. Certainly a conservative choice here if the singular values decay quick.

pybind/linalg_py.cpp
manuschneider: This is currently the standard behavior for all functions using random numbers. One could implement the suggestion of @ianmccul , but it would affect several places and the pybind part.

src/linalg/Rand_isometry.cpp
```diff
@@ -0,0 +1,42 @@
+#include <iostream>
+#include <string>
+#include <vector>
+
+#include "Tensor.hpp"
+#include "UniTensor.hpp"
+#include "algo.hpp"
+#include "linalg.hpp"
+#include "random.hpp"
+using namespace std;
```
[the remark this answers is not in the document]
manuschneider: Removed it here, but did not touch the 88 other occurances in .cpp files and .cu files

src/linalg/Rsvd.cpp
manuschneider: Copied this behavior from other SVD methods, probably this has to be changed in many places. So should we pass `cytnx_uint64` by value as well?

src/linalg/Rsvd.cpp
```diff
@@ -0,0 +1,236 @@
...
+  #ifdef UNI_GPU
+    #ifdef UNI_CUQUANTUM
+      #include "backend/linalg_internal_gpu/cuQuantumGeSvd_internal.hpp"
+    #endif
+  #endif
+
+namespace cytnx {
+  namespace linalg {
+    typedef Accessor ac;
```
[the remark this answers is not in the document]
manuschneider: On the other hand, there are 22 further occurances of this typedef. Maybe better keep one convention?

src/linalg/Rsvd.cpp
ianmccul: The entire codebase of cytnx is filled with pass-by-reference where pass-by-value would be more appropriate.  For new code, pass everything by value unless there is a reason otherwise.

tests/utils/rsvd_reference_tensors.ipynb
manuschneider: Because I used a notebook to create the test data, it was the fastest option. I think there is a `.py` file already in the folder as well.

src/linalg/Rsvd_truncate.cpp
manuschneider: I set `power_iteration=0` as a default for `Rsvd_truncate` now and use oversampling by default instead. The versions without oversampling use `power_iteration=2` still.

tests/linalg_test/Svd_truncate_test.cpp
IvanaGyro: I see. Maybe you can just put the new testcases in the existing test files. So that we will not have to modify extra file if we want to refactor the tests in the future.

src/BlockFermionicUniTensor.cpp
```diff
@@ -2760,7 +2760,7 @@ namespace cytnx {
     this->combineBonds(idx_mapper, force);
   }
 
-  void _BKF_from_DN(BlockFermionicUniTensor *ths, DenseUniTensor *rhs, const bool &force) {
+  void _bkf_from_dn(BlockFermionicUniTensor *ths, DenseUniTensor *rhs, const bool &force) {
```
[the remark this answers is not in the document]
ianmccul: The google coding conventions implicitly assume that you don't use leading underscores, because for most purposes they are banned by the C++ standard:

>    All identifiers that begin with an underscore followed by an uppercase letter, or that contain a double underscore, are reserved to the implementation for any use.

>    All identifiers that begin with an underscore are reserved in the global namespace.

If you are going to use a leading underscore, it **must** be followed by a *lowercase letter*, and it **must** be in some namespace or function scope, it cannot be in the global namespace.

In practice it will probably just work, compilers don't check identifier names, but don't be surprised if there is some conflict.

tests/utils/rsvd_reference_tensors.ipynb
manuschneider: The tensors that are created here are part of the repository anyways, these are just helper functions in case something needs to be changed in the future. I changed them to `.py`.

tests/linalg_test/Svd_truncate_test.cpp
manuschneider: If we use one test file for `Svd`, `Gesvd` and `Rsvd`, the following functions could be reused:
```
bool ReComposeCheck(const UniTensor& Tin, const std::vector<UniTensor>& Tout, double tol);
bool CheckLabels(const UniTensor& Tin, const std::vector<UniTensor>& Tout);
bool SingularValsCorrect(const UniTensor& res, const UniTensor& ans, double tol);
```
But these are only small test functions anyways.

On the other hand, we can easily switch test cases on and off by removing files from `tests/CMakeLists.txt`, and doing a diff between the files is also very easy to see differences if something needs to be changed.

I think combining the tests only makes sense if we automatize the process even more, and write test functions that loop over the three versions and test all of them in one test function (however, `Rsvd` works a bit differently and I am not sure if we should integrate it). We could think about it if #623 is implemented, and I would discuss it there.

I keep it for now as it is to keep consistency between `Rsvd_test` and `GeSvd_test`/`Svd_test`.

PR #621, opened 2025-06-26
avoid function names starting with `_`

manuschneider:
Function names starting with `-` are reserved in C++. We use them to mimick the Python convention, but this could possibly cause conflicts with certain compilers (in the future).

Here is the previous discussion on this:

@IvanaGyro 
Following the coding style we applied. Function names should be upper camel case. C++ doesn't need to imply internal on the method name because all functions declared in .cpp (source) files and in internal namespaces are internal in the language semantics. In contrast Python doesn't have access control, but developers feel they need to indicate a method is "private" as other languages like C++. As a result, there is the underscore convention.

See: https://google.github.io/styleguide/cppguide.html#Function_Names

@ianmccul  ianmccul [Jun 26, 2025](https://github.com/Cytnx-dev/Cytnx/pull/619#discussion_r2168002462)

The google coding conventions implicitly assume that you don't use leading underscores, because for most purposes they are banned by the C++ standard:

    All identifiers that begin with an underscore followed by an uppercase letter, or that contain a double underscore, are reserved to the implementation for any use.

    All identifiers that begin with an underscore are reserved in the global namespace.

If you are going to use a leading underscore, it must be followed by a lowercase letter, and it must be in some namespace or function scope, it cannot be in the global namespace.

In practice it will probably just work, compilers don't check identifier names, but don't be surprised if there is some conflict.Following the coding style we applied. Function names should be upper camel case. C++ doesn't need to imply internal on the method name because all functions declared in .cpp (source) files and in internal namespaces are internal in the language semantics. In contrast Python doesn't have access control, but developers feel they need to indicate a method is "private" as other languages like C++. As a result, there is the underscore convention.

See: https://google.github.io/styleguide/cppguide.html#Function_Names

Additionally, could you cherry-pick this commit to another branch and open a new PR? It seems that this commit is not related to this PR. This PR is growing and the workload on reviewing this PR become heavier 😂.
Member
@[ianmccul](https://github.com/ianmccul) ianmccul [Jun 26, 2025](https://github.com/Cytnx-dev/Cytnx/pull/619#discussion_r2168002462)

The google coding conventions implicitly assume that you don't use leading underscores, because for most purposes they are banned by the C++ standard:

    All identifiers that begin with an underscore followed by an uppercase letter, or that contain a double underscore, are reserved to the implementation for any use.

    All identifiers that begin with an underscore are reserved in the global namespace.

If you are going to use a leading underscore, it must be followed by a lowercase letter, and it must be in some namespace or function scope, it cannot be in the global namespace.

In practice it will probably just work, compilers don't check identifier names, but don't be surprised if there is some conflict.

### DIFF
--- a/src/BlockFermionicUniTensor.cpp
+++ b/src/BlockFermionicUniTensor.cpp
@@ -2760,7 +2760,7 @@ namespace cytnx {
     this->combineBonds(idx_mapper, force);
   }
 
-  void _BKF_from_DN(BlockFermionicUniTensor *ths, DenseUniTensor *rhs, const bool &force) {
+  void _bkf_from_dn(BlockFermionicUniTensor *ths, DenseUniTensor *rhs, const bool &force) {
     if (!force) {
       // more checking:
       if (int(rhs->bond_(0).type()) != bondType::BD_NONE) {
@@ -2799,11 +2799,11 @@ namespace cytnx {
     }
   }
 
-  void _BKF_from_BK(BlockFermionicUniTensor *ths, BlockUniTensor *rhs, const bool &force) {
+  void _bkf_from_bk(BlockFermionicUniTensor *ths, BlockUniTensor *rhs, const bool &force) {
     cytnx_error_msg(true, "[ERROR] BlockFermionicUT-> BlockUT not implemented.%s", "\n");
   }
 
-  void _BKF_from_BKF(BlockFermionicUniTensor *ths, BlockFermionicUniTensor *rhs,
+  void _bkf_from_bkF(BlockFermionicUniTensor *ths, BlockFermionicUniTensor *rhs,
                      const bool &force) {
     cytnx_error_msg(true, "[ERROR] BlockFermionicUT-> BlockFermionicUT not implemented.%s", "\n");
   }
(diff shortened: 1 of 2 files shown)


manuschneider:
Just to clarify: this commit only avoids some `_` followed by upper case letters, which is reserved in C++. Other occurrences of functions and variable starting with `_` are used excessively for internal implementations in the current implementation.

IvanaGyro:
Thank you for your effort to create a new PR for reviewing.

I know it's frustrated and annoyed when being corrected the coding style. I am sorry about my rudeness. Being strict is not my purpose. My habit brought from Google. At Google, we cannot merge any code that violates the coding style guide. The number of rules in the internal style guide is almost twice than the public Google C++ style guide, so the developers must be very careful to not violate the guide. Reviewers almost always help to point out the code that doesn't follow the guide and let the contributor to correct it.

I think this is a good moment for us to discuss if we have to strictly follow the picked style guide for new files? Personally, I feel following the existing style in the existing files is fine. If the answer of this question is no, I will ignore all style problems in the future and just try to review the logic. If the answer of the question is yes, than go back to this issue. Is there any reason that the function name cannot be the format required by the style guide, `BkfFromDn`?

yingjerkao (approved):

PR #625, opened 2025-07-04
Add CUDA Sort tests and handle edge case for zero-dimension sorting

jysh1214:


tests/gpu/algo_test/Sort_test.cpp
```diff
@@ -0,0 +1,60 @@
...
+namespace SortTest {
+
+  void CheckResult(const Tensor& sorted, const Tensor& original);
+
+  void TestSortAllTypes(const std::vector<cytnx_uint64>& shape, const std::string& test_name) {
+    for (auto dtype : dtype_list) {
+      if (dtype == Type.Bool) {
+        continue;
+      }
+      SCOPED_TRACE("Testing " + test_name + " with dtype: " + std::to_string(dtype));
```
IvanaGyro, line 20: Here are some formatting issues:
1. The order of include statements should follow [the guide](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes).
2. `<gtest/gtest.h>` should be `"gtest/gtest.h"` according to the guide.
3. > All of a project's header files should be listed as descendants of the project's source directory without use of UNIX directory aliases . (the current directory) or .. (the parent directory).
4. [Do not use using-directives (e.g., using namespace foo)](https://google.github.io/styleguide/cppguide.html#Namespaces) Although the risk of using using-directive is small in source files.

Additionally, do you consider to use [AssertionResult](https://google.github.io/googletest/advanced.html#using-a-function-that-returns-an-assertionresult) to rewrite `CheckResult`? And would you consider to use [value-parameterized tests](https://google.github.io/googletest/advanced.html#value-parameterized-tests)? These two are optional.

jysh1214: Hi @IvanaGyro ,
1. It's fixed.
2. It's fixed.
3. I will prepare another PR to fix it.
4. It's fixed.

Rewrote ` CheckResult` using `AssertionResult`. 
Used `TEST_T` to implement tests with different tensor sizes.

yingjerkao:
@jysh1214 Please refer to https://github.com/Cytnx-dev/Cytnx/tree/master/developer_tools and install pre-commit for automatic Clang format linting

jysh1214:
> @jysh1214 Please refer to https://github.com/Cytnx-dev/Cytnx/tree/master/developer_tools and install pre-commit for automatic Clang format linting

Got it. Thanks.

tests/gpu/algo_test/Sort_test.cpp
IvanaGyro: > I will prepare another PR to fix it.

Haha, I'm sorry that I just found the `tests` folder is not added as the include path in the cmake files.

IvanaGyro (approved):
LGTM

jysh1214:
> Actually, I am thinking if replacing the for-loop of dtype_list with the value-parameterized tests is a good idea or not. What do you think?

Going through all `dtypes` using for-loop is clear enough for me, with the test name postfixed by `_all_types`.
If I have a more graceful implementation, I’ll submit another PR for that.

Issue #601, opened 2025-05-28
Use mkl_lp64 for default binary build


yingjerkao:
> I suggest we ship binaries with mkl_lp64 instead of mkl_ilp64 by default. Users can decide to compile the ilp64 version if that is real necessary.  

 _Originally posted by @yingjerkao in [#595](https://github.com/Cytnx-dev/Cytnx/issues/595#issuecomment-2906520849)_

yingjerkao:
> How does MKL dynamic runtime decide to link to mkl_lp64 or mkl_ilp64? I think we shipped Cytnx with mkl_rt to Conda.

Although it is dynamic, we still have to choose a version

https://github.com/Cytnx-dev/Cytnx/blame/0e9794d7f1dfef233ee86c94ea32e5bc7c4b5cfb/CytnxBKNDCMakeLists.cmake#L1-L9

IvanaGyro:
I found this. We may also switch to ILP64 at runtime.

https://github.com/Cytnx-dev/Cytnx/blob/0e9794d7f1dfef233ee86c94ea32e5bc7c4b5cfb/src/backend/linalg_internal_interface.cpp#L15-L34

PR #626, opened 2025-07-10
Integrate ARPACK to Cytnx.

hunghaoti:
## Integrate ARPACK to Cytnx.
1. This branch is branched from branch #601.
2. Change the internal implementation for Arnoldi algorithm using `ARPACK` package and support for Block UniTensor.
3. Add the new API for Lanczos algorithm using `ARPACK` package. Still leave the original API there.
4. Add the corresponding unit test.
5. Add the corresponding API documentation and also modify the user guide using arpack dependency.
6. Add the arpack dependency in workflow and build action.
7. Fix linalg::Matvec bug in gpu for dtype=Type.Float.
8. For `ptr_as` and `gpu_ptr_as` in 'include/linalg.hpp', using `reinterpret_cast` rather than `static_cast`.
9. Remove redundant files.

### DIFF
--- a/pybind/cytnx.cpp
+++ b/pybind/cytnx.cpp
@@ -84,7 +84,6 @@ PYBIND11_MODULE(cytnx, m) {
   // m.def("detect_mkl_interface", &cytnx::detect_mkl_interface);
   // m.def("get_mkl_code", &cytnx::get_mkl_code);
 
-
   // global vars
   // m.attr("cytnxdevice") = cytnx::cytnxdevice;
   // m.attr("Type")   = py::cast(cytnx::Type);


hunghaoti:
I benchmark this implementation and the original one. For the case calculating the eigenvector of the transfer matrix of uMPS with bond dimension D with crit=1.0e-9 using Arnoldi algorithm:
Run on (8 X 3315.83 MHz CPU s)
| dtype| D| k | original  | current implementation|
|--------|---|-----|------------|----------------------------------|
|C128  |10|1    |64.3 (ms)|      3.13 (ms)                  |
|C128  |20|1    |659 (ms)|        18.1 (ms)                 |
|C128  |40|1    |7311 (ms)|      78.9 (ms)                 |
|C128  |10|4    |109 (ms)|        4.74 (ms)                 |
|C128  |20|4    |891 (ms)|        18.3 (ms)                 |
|C128  |40|4    |9305 (ms)|       79.6 (ms)                |

For the case using Lanczos algorithm to simulate the ground state of the one-site effective Hamiltonian:
| dtype| D|  original $\epsilon_e=10^{-12}$ | current (which='SA') $\epsilon_r=10^{-12}$|current (which='SA') $\epsilon_r=10^{-6}$
|--------|---|------------|----------------------------------|--------------------|
|F64  |20|11.1 (ms)|      7.96 (ms)                  | 4.74 (ms)   |
|F64  |50|33.7 (ms)|        41.3 (ms)                 | 24.4 (ms)  |
|F64  |100|183 (ms)|      248 (ms)                 | 139 (ms)     |

ianmccul:
Its interesting that $k=1$ is faster than $k=4$ here, but both values are so small, I am not sure it is representative. For ARPACK, a common default for $k$ is `max(20, 2*nev+1)`.

For a DMRG solver, the number of iterations per step is small enough that ARPACK is not so effective here, you typically want better control over the number of iterations directly.  But with using different convergence criteria, it is not so easily comparable. The ARPACK criterion should be much more robust.

hunghaoti:
> Its interesting that k = 1 is faster than k = 4 here, but both values are so small, I am not sure it is representative. For ARPACK, a common default for k is `max(20, 2*nev+1)`.
> 
> For a DMRG solver, the number of iterations per step is small enough that ARPACK is not so effective here, you typically want better control over the number of iterations directly. But with using different convergence criteria, it is not so easily comparable. The ARPACK criterion should be much more robust.

The parameter `k` is the number of eigenvalues and k=nev. Yes, in `scipy.sparse.linalg.eigsh`, the default value of `ncv` is max(20, 2*nev+1). We also provide this input argument `ncv` for users to choose this value.

IvanaGyro:
> Typically, $\epsilon_r\approx \epsilon_e^2$.

I am familiar with these algorithm. However, based on the relation of $\epsilon_r$ and $\epsilon_e$ you mentioned, why you tested with $\epsilon_r = 10^{-6}$ instead of $\epsilon_r = 10^{-24}$?

hunghaoti:
> I am familiar with these algorithm. However, based on the relation of ϵ r and ϵ e you mentioned, why you tested with ϵ r = 10 − 6 instead of ϵ r = 10 − 24 ?

Sorry I am wrong. Actually is $\epsilon_r^2\\approx\epsilon_e$

yingjerkao (approved):

PR #628, opened 2025-07-12
Refactor sort internal implementation

jysh1214:
- Use macros to eliminate duplicated code
- Fix include order according to google c++ style guide
- Replace NULL with `nullptr` for modern C++ compliance
- Replace function pointers with `std::function` for type safety

### DIFF
--- a/src/backend/algo_internal_interface.cpp
+++ b/src/backend/algo_internal_interface.cpp
@@ -1,6 +1,22 @@
 #include "algo_internal_interface.hpp"
 
-using namespace std;
+// Macro to assign a device-specific sort function to a function table.
+#define CYTNX_ASSIGN_SORT_FUNC(Device, Table, DataType) \
+  Table[Type.DataType] = Device##Sort_internal_##DataType
+
+// Macro to assign all supported data type sort functions for a specific device.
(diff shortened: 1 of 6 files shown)


jysh1214:
Yeah. Let's follow the precedent.

src/backend/algo_internal_interface.hpp
```diff
@@ -3,6 +3,7 @@
 
 #include <iostream>
 #include <vector>
+#include <functional>
```
IvanaGyro, line 6: Was this added unintentionally?

src/backend/algo_internal_interface.cpp
```diff
@@ -1,44 +1,11 @@
 #include "algo_internal_interface.hpp"
 
-using namespace std;
-
 namespace cytnx {
   namespace algo_internal {
 
     algo_internal_interface aii;
```
IvanaGyro, line 6: Is this variable used in any place? Can the whole file be removed?

jysh1214: I'll prepare another PR to remove the unnecessary files.

IvanaGyro (approved):

PR #645, opened 2025-08-18
Remove CUTT dependency

yingjerkao:
CUTT is integrated into cuTensor so this PR removes the dependencies of CUTT and use cuTensor APIs to perform permutation.


hunghaoti (approved):
I have checked all unit test including gpu test passed.

PR #647, opened 2025-08-20
Add unittests for cuAdd

jysh1214:


tests/gpu/linalg_test/Add_test.cpp
```diff
@@ -0,0 +1,196 @@
...
+      return ::testing::AssertionFailure()
+             << "Add result mismatch: CUDA Add result differs from CPU Add result. "
+             << "Left dtype: " << left_tensor.dtype() << ", Right dtype: " << right_tensor.dtype()
+             << ", tolerance used: " << tolerance;
+    }
+
+    return ::testing::AssertionSuccess();
+  }
+
+  ::testing::AssertionResult CheckAddScalarResult(const cytnx::Tensor& gpu_result,
```
yingjerkao, line 152: Is it a common practice to test against  CPU results for GPU?

jysh1214: I think comparing the cpu and cuda code is sufficient for element-wise functions.
I mean if we write scalar code for testing, it would be almost the same as the cpu version.
However, if scalar code for testing is required, I can do that.

yingjerkao: Ok. Thanks. We will follow this practice of comparing CPU/GPU results.

yingjerkao (approved):

PR #648, opened 2025-08-23
Refine Add CPU implement

jysh1214:
## Motivation

The Cytnx Add implementation contained lots duplicated code across different type combinations, which caused:

## Solution

Refactored the Add internal implementation using templates:
- Abstract common arithmetic logic into reusable template functions
- Use type priority to handle `<lower_type>_<higher_type>` case


yingjerkao:
For education purpose, could you summarize the idea behind this implementation?

ianmccul:
It looks like this implements another mechanism to do type promotion, based on `include/utils/type_priority.hpp`, which is not ideal - it it isn't extensible and has some drawbacks. In https://github.com/Cytnx-dev/Cytnx/pull/514/commits/710c98c73d24dba02bdbecad54dcfc8dcdf5b6fc I refactored the `type_promote` mechanism, which also isn't ideal but it must match exactly the promotion flow that is implicitly contained in the function tables for binary operations that assume that result type. In the same PR I added a better dispatch mechanism -- see `Kron` function for an example.

PR #651, opened 2025-08-26
Refine Add CPU implement

jysh1214:
The Cytnx Add implementation contained lots duplicated code across different type combinations, which caused:

This PR refines the Add CPU implementation.
Our goal is to generalize all arithmetic operations using templates, and eventually refactor the multiple-dispatch mechanism for arithmetic.

```c++
template <typename TLin, typename TRin>
void ArithmeticInternalImpl(...) {
  if (type == 0)
    cytnx::linalg_internal::AddInternalImpl<TLin, TRin>(
      out, Lin, Rin, len, shape, invmapper_L, invmapper_R);
  else if (type == 1)
    cytnx::linalg_internal::MulInternalImpl<TLin, TRin>(
      out, Lin, Rin, len, shape, invmapper_L, invmapper_R);
  else if (type == 2)
    // ...
  else
    // ...
}
```

Next, we plan to replace the `linalg_internal::lii.Ari_ii` usage in arithmetic .cpp files with the `std::variant` + `std::visit` pattern, similar to what we did in commit [710c98c](https://github.com/Cytnx-dev/Cytnx/commit/710c98c73d24dba02bdbecad54dcfc8dcdf5b6fc).


src/backend/linalg_internal_cpu/Add_internal.hpp
```diff
@@ -2,754 +2,83 @@
...
 
 #include <assert.h>
-#include <iostream>
-#include <iomanip>
 #include <vector>
 #include "backend/Storage.hpp"
 #include "Type.hpp"
+#include "utils/utils.hpp"
+
+namespace {
```
IvanaGyro, line 10: The style guide doesn't allow using unnamed namespace in the header files. See: https://google.github.io/styleguide/cppguide.html#Internal_Linkage

I think the reason is that this will cause every transition unit own its copy of the namespace and make problem when applying [Include What You Use](https://google.github.io/styleguide/cppguide.html#Include_What_You_Use) rule.

Can the implementation of `AddInternalCore` be put in `AddInternalImpl` function directly?

IvanaGyro (approved):

Issue #587, opened 2025-03-20
Merge repo Cytnx_doc into this repo


IvanaGyro:
I suggest keeping the document as close as the code so that the developers can keep the document aligned with the code easily. A case shows that putting the document and the code in the same repo benefits. Considering we are going to change dependency A to dependency B and also want to keep the document correct enough at any moment, we have to do three steps when the document and the code are located in different repos:

1. Added B as a dependency in the document (A is still the dependency, so we cannot remove A from the document here.)
2. Switch dependency A to dependency B in the codebase
3. Removed dependency A from the document

It's worth to mention it is not possible to make the document aligned with the codebase at each moment when dependency A conflicts with dependency B in this case.

By merging the document repo into this repo, we can easily make the document aligned with the code over time. We only need one step, even one commit, to finish the job that needs three steps above while the document and the code are located in different repos.

Here is the steps for merging two git repos: https://stackoverflow.com/questions/1425892/how-do-you-merge-two-git-repositories

kaihsin:
I believe this is the goal. The problem is the build process takes too long, so updating just user guides takes forever. The User guides were more designed as "blog" style posts instead of API documentation.

IvanaGyro:
We spend most of time on build HPTT. There are some ways to reuse the previous build to reduce the build time in GitHub actions.

yingjerkao:
@hunghaoti I wonder if there is a benchmark for with and without HPTT?

hunghaoti:
> [@hunghaoti](https://github.com/hunghaoti) I wonder if there is a benchmark for with and without HPTT?

I benchmark `UniTensor::contiguous_` function, which used `HPTT`.
Run on (8 X 4200 MHz CPU s)
cpu info:
```
CPU Caches:
  L1 Data 48 KiB (x4)
  L1 Instruction 32 KiB (x4)
  L2 Unified 1280 KiB (x4)
  L3 Unified 8192 KiB (x1)
Load Average: 2.30, 3.18, 3.78
```
with `HPTT`
```
---------------------------------------------------------------------
Benchmark                           Time             CPU   Iterations
---------------------------------------------------------------------
BM_Tensor_contiguous/100        0.020 ms        0.020 ms        33972
BM_Tensor_contiguous/1000        2.30 ms         2.29 ms          308
BM_Tensor_contiguous/10000        179 ms          178 ms            4   
```
without `HPTT`
```
---------------------------------------------------------------------
[...]

PR #657, opened 2025-09-02
Apply .gitattributes configurations

ultimatile:
Although commit 9adad37 introduced .gitattributes to enforce LF, many files still retain CRLF line endings, resulting in diffs every time the repository is checked out. They should be converted to LF as specified by .gitattributes. Note that because only the line endings would changed, the differences may not be obvious in the "Files changed" view.


IvanaGyro:
This seems an issue about merging cytnx_doc repo into this repo. Link #587 for reference.

Thanks for your contribution!

PR #658, opened 2025-09-02
Refine Mul CPU implementation

jysh1214:
The Cytnx arithmetic implementation contained lots duplicated code across different type combinations, which caused:

This PR refines the Mul CPU implementation, similar to https://github.com/Cytnx-dev/Cytnx/pull/651.
Our goal is to generalize all arithmetic operations using templates, and eventually refactor the multiple-dispatch mechanism for arithmetic.

```c++
template <typename TLin, typename TRin>
void ArithmeticInternalImpl(...) {
  if (type == 0)
    cytnx::linalg_internal::AddInternalImpl<TLin, TRin>(
      out, Lin, Rin, len, shape, invmapper_L, invmapper_R);
  else if (type == 1)
    cytnx::linalg_internal::MulInternalImpl<TLin, TRin>(
      out, Lin, Rin, len, shape, invmapper_L, invmapper_R);
  else if (type == 2)
    // ...
  else
    // ...
}
```

Next, we plan to replace the `linalg_internal::lii.Ari_ii` usage in arithmetic .cpp files with the `std::variant` + `std::visit` pattern, similar to what we did in commit [710c98c](https://github.com/Cytnx-dev/Cytnx/commit/710c98c73d24dba02bdbecad54dcfc8dcdf5b6fc).


src/backend/linalg_internal_cpu/Mul_internal.hpp
```diff
@@ -1,755 +1,70 @@
...
-    void Mul_internal_btb(boost::intrusive_ptr<Storage_base> &out,
-                          boost::intrusive_ptr<Storage_base> &Lin,
-                          boost::intrusive_ptr<Storage_base> &Rin, const unsigned long long &len,
-                          const std::vector<cytnx_uint64> &shape,
-                          const std::vector<cytnx_uint64> &invmapper_L,
-                          const std::vector<cytnx_uint64> &invmapper_R);
-
+    template <typename TLin, typename TRin>
+    inline void MulInternalImpl(boost::intrusive_ptr<Storage_base> &out,
+                                boost::intrusive_ptr<Storage_base> &Lin,
```
IvanaGyro, line 13: I know this function is the templatized version of the original function. I just wonder `Lin` and `Rin` can be passed by value of instrusive_ptr instead of be passed by reference of instrusive_ptr, right? Or maybe `const&` is also a better choice than reference.

It is totally fine to not make a decision in this PR.

IvanaGyro (approved):
You may want to correct the PR message before merging :)

> The Cytnx Add implementation contained lots...

I think "Add" should be "Mul".

PR #678, opened 2025-09-19
Change intrusive_ptr_add_ref/release signatures so ADL can find them

ultimatile:
I encountered a build failure in my environment (macOS on Apple Silicon with Homebrew).
The functions `intrusive_ptr_add_ref` and `intrusive_ptr_release` did not match Boost’s expected signatures (see the [reference](https://www.boost.org/doc/libs/latest/boost/smart_ptr/intrusive_ptr.hpp)), so the unqualified calls were not found by ADL. 
They now follow

```C++
void intrusive_ptr_add_ref(T* p);
void intrusive_ptr_release(T* p);
```

A forward declaration of `Storage_base` is also added so these functions are found by ADL. 
This resolves the build in my environment.

The coupling the `intrusive_ptr_base` code to the concrete `Storage_base` type isn't ideal. 
If you have a cleaner way to decouple this, I'm happy to revise.

### DIFF
--- a/include/intrusive_ptr_base.hpp
+++ b/include/intrusive_ptr_base.hpp
@@ -9,6 +9,11 @@
 
 namespace cytnx {
   /// @cond
+  class Storage_base;  // forward, defined in backend/Storage.hpp
+  // The following two declarations are necessary for ADL.
+  void intrusive_ptr_add_ref(Storage_base *);
+  void intrusive_ptr_release(Storage_base *);
+
   template <class T>
   class intrusive_ptr_base {
    public:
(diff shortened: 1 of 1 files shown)


IvanaGyro:
What and which version of compiler did you use? Could you also provide the error messages from the compiler?

ultimatile:
My build commands are `module load llvm libomp openblas && uv build` (for LLVM Clang v21.1.1) and `module load libomp openblas && uv build` (for Apple Clang v17.0.0.17000013)

`module` commands just set `CPPFLAGS` and `LDFLAGS` like `CPPFLAGS="-I/opt/homebrew/opt/openblas/include -I/opt/homebrew/opt/libomp/include"`

I attached log messages for both cases:
[llvmclang_log.txt](https://github.com/user-attachments/files/22422704/llvmclang_log.txt)
[appleclang_log.txt](https://github.com/user-attachments/files/22422706/appleclang_log.txt)

IvanaGyro:
Maybe we also don't need to change the signature in intrusive_ptr_base.hpp? Is the final solution of this problem moving class Storage_base to a new header file?

IvanaGyro (approved):

PR #680, opened 2025-09-20
Inv: bug fix and implementation for UniTensor

manuschneider:
-fixed a bug in Inv, where the clipping was applied to the value instead of the absolute part of it for real valued Tensors.
-implemented Inv for UniTensor

### DIFF
--- a/include/UniTensor.hpp
+++ b/include/UniTensor.hpp
@@ -360,6 +360,9 @@ namespace cytnx {
     virtual boost::intrusive_ptr<UniTensor_base> normalize();
     virtual void normalize_();
 
+    virtual boost::intrusive_ptr<UniTensor_base> Inv(double clip);
+    virtual void Inv_(double clip);
+
     virtual boost::intrusive_ptr<UniTensor_base> Conj();
     virtual void Conj_();
 
@@ -777,6 +780,14 @@ namespace cytnx {
     void Div_(const boost::intrusive_ptr<UniTensor_base> &rhs);
     void Div_(const Scalar &rhs);
     void lDiv_(const Scalar &lhs);
+
+    void Inv_(double clip) { this->_block.Inv_(clip); };
+
+    boost::intrusive_ptr<UniTensor_base> Inv(double clip) {
+      boost::intrusive_ptr<UniTensor_base> out = this->clone();
+      out->Inv_(clip);
+      return out;
+    };
 
     void Conj_() { this->_block.Conj_(); };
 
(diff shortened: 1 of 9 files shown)


manuschneider:
The TDVP check fails. It seems that the Lanczos does not converge. However, the algorithm should not be affected by any of the changes in this branch. Could it be that TDVP is sensitive to the random seed?
On my machine the TDVP check works fine (Linux). Any suggestions?

pybind/linalg_py.cpp
```diff
@@ -574,16 +585,16 @@ void linalg_binding(py::module &m) {
 
   m_linalg.def(
     "Pow", [](const UniTensor &Tin, const double &p) { return cytnx::linalg::Pow(Tin, p); },
-    py::arg("Tn"), py::arg("p"));
+    py::arg("Tin"), py::arg("p"));
```
IvanaGyro, line 588: These are breaking changes of Python API.

src/linalg/Inv.cpp
```diff
@@ -32,6 +33,32 @@ namespace cytnx {
       }
       return out;
     }
+
+  }  // namespace linalg
+}  // namespace cytnx
+
+namespace cytnx {
+  namespace linalg {
+    cytnx::UniTensor Inv(const cytnx::UniTensor &Tin, double clip) {
```
IvanaGyro, line 42: These can be put together with the namespace above.

src/linalg/Inv_.cpp
```diff
@@ -1,34 +1,57 @@
 #include "linalg.hpp"
 #include "Tensor.hpp"
+#include "UniTensor.hpp"
 
 #ifdef BACKEND_TORCH
 #else
 
-  #include "backend/linalg_internal_interface.hpp"
+  #include "../backend/linalg_internal_interface.hpp"
```
IvanaGyro, line 8: I think the change of include path is not needed.

IvanaGyro (commented):
I didn't carefully review this PR. Just find some points that need to be modified with a glance.

pybind/linalg_py.cpp
manuschneider: see #690

src/linalg/Inv_.cpp
IvanaGyro: This is what I said.

hunghaoti (approved):

manuschneider:
#690 is still open, I would like to fix it before merging, since some breaking changes are introduced here.

Issue #684, opened 2025-09-25
Bug in Pow for complex types on GPU


manuschneider:
On the GPU, the `Pow` function gives other results on the GPU than the CPU. An example is in the unittests in the branch [fix_Pow](https://github.com/Cytnx-dev/Cytnx/tree/fix_Pow)

Kernels that need to be fixed:
```
cuPow_internal_kernel_cd
cuPow_internal_kernel_cf
```

PR #686, opened 2025-09-25
Fix bug in Pow on GPU with complex data type

manuschneider:
Fixes https://github.com/Cytnx-dev/Cytnx/issues/684Fixes https://github.com/Cytnx-dev/Cytnx/issues/684


yingjerkao:
@manuschneider I wonder if the failed test can be fixed?

manuschneider:
> @manuschneider I wonder if the failed test can be fixed?

I do not know what is going on here. I only modified the GPU part, but the TDVP test does not use GPU at all, right.

My feeling is that the additional unit tests change the status of the random number generator, and the TDVP test seems sensitive to it.

@hunghaoti could you have a look into it?

PR #694, opened 2025-10-24
Fix cuMod and add unittests for cuMod

jysh1214:


src/backend/linalg_internal_gpu/cuMod_internal.cu
```diff
@@ -1243,7 +1267,13 @@ namespace cytnx {
       if (len % 512) NBlocks += 1;
 
       if (Lin->size() == 1) {
-        cufMod_lconst_kernel<<<NBlocks, 512>>>(_out, _Lin[0], len, _Rin);
+        // TODO: Investigate why _Rin is a host pointer instead of device pointer
+        // This causes CUDA Error 700 (illegal memory access) when kernel tries to access host
```
IvanaGyro: "This" in the comment is a little bit confused because the code below works.

IvanaGyro (approved):

Issue #692, opened 2025-10-08
Unexpected dtype returned when operating with Float32 / ComplexFloat32 tensors and numpy scalars


Chan-Sing-Hong:
**Description**
Unexpected dtype are returned when performing operations involving `Float`, `ComplexFloat` Tensors and `numpy.float32`, `numpy.complex64`.

**Expected Behavior**
`cytnx.Type.Float`                 /   `numpy.float32`         ->   `cytnx.Float`
`cytnx.Type.ComplexFloat`   /   `numpy.float32`         ->   `cytnx.Type.ComplexFloat`
`cytnx.Type.ComplexFloat`   /   `numpy.complex64`   ->   `cytnx.Type.ComplexFloat`

**Actual Behavior**
However, all of the above cases currently return `ComplexDouble (Complex Float64)`.

A notebook demonstrating the issue is attached: 
[Float32_issue.ipynb](https://github.com/user-attachments/files/22757875/Float32_issue.ipynb)

IvanaGyro:
The following code is copied from the notebook for easier reference.

```py
A = cytnx.ones([2,2,2,2]).astype(cytnx.Type.ComplexFloat)
B = numpy.float32(1)
C = A / B

print(A.dtype_str())
print(B.dtype)
print(C.dtype_str())
```

output:

```
Complex Float (Complex Float32)
float32
Complex Double (Complex Float64)
```

IvanaGyro:
Is operating with numpy scalar a feature we want to support?

pcchen:
## Status: core fixed on master; one documented remainder

The dtype-loss core of this issue is fixed by the merged 2026-07 work:

- **#982** — `Type.type_promote` now promotes across the real/complex boundary by precision (`ComplexFloat × Double → ComplexDouble` instead of silently keeping the lower enum index).
- **#991** — the pybind operator keep-sets preserve numpy scalar dtypes on Tensor/UniTensor arithmetic and `__setitem__` (`np.float32`/`np.complex64`/all integer numpy dtypes no longer collapse to Int64/Double paths), pinned by `pytests/binding_dtype_test.py`.
- **#984** — the remaining linalg entry points (`Dot`, `Matmul(_dg)`, `Axpy`, `Gemm(_Batch)`, `Ger`, `Mod`, `Outer`, `Tensordot`, …) route dtype selection through `Type.type_promote`.

[...]

PR #699, opened 2025-11-01
Add numpy scalar bindings and pybind for cytnx.Scalar

manuschneider:
Previously, the numpy classes were converted into Python specific types before handed to the C++ side of Cytnx. Therefore, many type conversions did not work as expected. For example, cytnx.Float + numpy.Float = cytnx.Double

I added explicit function calls for numpy scalars to avoid this. However, the reverse functions like numpy + cytnx do not work yet because they either try to call numpy routines or still do the conversion to python specific types first.

Moreover, I added pybindings for the cytnx.Scalar class so people can convert numpy scalars to cytnx scalars explicitly to avoid the problems.

This addresses #692

### DIFF
--- a/pybind/linalg_py.cpp
+++ b/pybind/linalg_py.cpp
@@ -270,8 +270,6 @@ void linalg_binding(py::module &m) {
       return cytnx::linalg::ExpH(Tin, a, b);
     },
     py::arg("Tin"), py::arg("a"), py::arg("b") = cytnx_int16(0));
-  // m_linalg.def("ExpH", [](const UniTensor &Tin, const cytnx::Scalar &a, const cytnx::Scalar &b) {
-  // return cytnx::linalg::ExpH(Tin, a, b); }, py::arg("Tin"), py::arg("a"), py::arg("b") = 0);
   m_linalg.def(
     "ExpH",
     [](const UniTensor &Tin, const py::numpy_scalar<std::complex<double>> &a,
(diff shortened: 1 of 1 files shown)


pybind/linalg_py.cpp
```diff
@@ -270,6 +270,79 @@ void linalg_binding(py::module &m) {
       return cytnx::linalg::ExpH(Tin, a, b);
     },
     py::arg("Tin"), py::arg("a"), py::arg("b") = cytnx_int16(0));
+  // m_linalg.def("ExpH", [](const UniTensor &Tin, const cytnx::Scalar &a, const cytnx::Scalar &b) {
```
IvanaGyro: Unused code?

pybind/linalg_py.cpp
```diff
@@ -270,6 +270,79 @@ void linalg_binding(py::module &m) {
...
+  // return cytnx::linalg::ExpH(Tin, a, b); }, py::arg("Tin"), py::arg("a"), py::arg("b") = 0);
+  m_linalg.def(
+    "ExpH",
+    [](const UniTensor &Tin, const py::numpy_scalar<std::complex<double>> &a,
+       const py::numpy_scalar<std::complex<double>> &b) {
+      return cytnx::linalg::ExpH(Tin, (cytnx::cytnx_complex128)a);
```
IvanaGyro: Is c-style cast a proper way to cast numpy_scalar? Or it should be `a.cast<T>()`? If c-style cast works , I think static_cast is a better option for C++.

PR #697, opened 2025-10-31
Revert "Refine Div CPU implement"

IvanaGyro:
Reverts Cytnx-dev/Cytnx#683

The PR being reverted seems to break the TDVP test in Python.

PR #705, opened 2025-11-09
Fix pytest tdvp.

hunghaoti:
Fix pytest tdvp.

### DIFF
--- a/pytests/example/TDVP/tdvp1_dense_test.py
+++ b/pytests/example/TDVP/tdvp1_dense_test.py
@@ -17,8 +17,8 @@ def test_tdvp1_dense():
     Jz = 0.0
     hx = 0.0
     hz = -1.0
-    tau = -1.0j
-    time_step = 12
+    tau = -0.5j
+    time_step = 20
     # prepare up state
     As, Es = tdvp1_XXZmodel_dense(J, Jz, hx, hz, MPS_rand, chi, tau, time_step)
     error = np.abs(Es[-1]-(-1.0*Nsites))


IvanaGyro:
Is


```py
tau = -1.0j
time_step = 12
```

not an accepted use case?

The test should pass if the use case is acceptable. We should fix the code that fails the test instead of modifying the test to satisfy the potentially problematic code. Please consider to review #697.

ianmccul:
The terminology here is confusing.  One might think that `time_step` is the Δt used in each iteration, but instead it appears to be the **number of iterations**, with each iteration being a time step `tau`. 

In most circumstances a timestep of 1 (in units where the excitation velocity is O(1)) is already quite large and will not converge to the physical solution.  In this case however, the model tested is rather trivial and is only testing an on-side magnetic field `hz = -1`. In principle this is exactly solvable with TDVP for arbitrary timestep, but it depends on the details of how the exponential is calculated. A Lanczos exponential will have trouble, since it is basically calculating a Taylor series expansion of `exp(j)` about zero, which requires at least a few iterations.  A quick look at https://github.com/Cytnx-dev/Cytnx/blob/master/src/linalg/Lanczos_Exp.cpp, its likely to have numerical stability problems.  But it is also unrealistic for real calculations (that would have a timestep << the excitation velocity). Basically, if you need more than a few iterations of the Lanczos for the exponential to converge then the timestep is too big. Since there is no reorthogonalization or other techniqes to stabilize the lanzcos exponential, setting `max_iter` to maybe 10 or so is probably a good idea.

As an aside, this code in https://github.com/Cytnx-dev/Cytnx/blob/bc1e5f93d30c02971423441f5265b578ee73a438/example/TDVP/tdvp1_dense.py#L91-L93 is a bit suspicious:

```c++
    sx = cytnx.physics.pauli('x').real()
    sy = cytnx.physics.pauli('y')
    sz = cytnx.physics.pauli('z').real()
 ```
 
 Normally, for spin-1/2 model, the spin operators are _half_ of the Pauli matrices, i.e. `sz = diag(-0.5/0.5)`.  That isn't done here, so actually the Hamiltonian parameters are twice (for on-site) and 4x (for two-site) the values one might expect. And hence it is calculating `exp(j)`, whereas for timestep `tau=1` and on-site field `hz=1` you would expect it to be calculating `exp(0.5j)`. SciPy `expm_multiply` uses 
>  A. H. Al-Mohy and N. J. Higham,
> “Computing the Action of the Matrix Exponential, with an Application to Exponential Integrators,”
> SIAM J. Sci. Comput. 33 (2), 488–511 (2011).
> https://doi.org/10.1137/100788860

TLDR: the test is faulty. But it is a (very special and physically unrealistic) case that TDVP would probably be able to handle with a robust matrix exponential algorithm. Fixing the timestep to 0.5 is OK, 0.1 would be better (and more realistic).

yingjerkao (approved):
This PR looks fine.

PR #710, opened 2025-11-20
Use TARGET_DIRECTORY for Kron.cpp CUDA property

ultimatile:
Building `src/linalg/Kron.cpp` for GPU should be handled by `nvcc` (or another CUDA-family compiler).
However, under the current configuration, the `LANGUAGE CUDA` property does not propagate to the top-level CMake configuration, causing the file to be compiled incorrectly with a C++ compiler such as `g++`.
This PR corrects the issue by using `TARGET_DIRECTORY`, enabling `Kron.cpp` to be built properly with `nvcc`.
This fix may also resolve the ordering issue mentioned in the comment.


IvanaGyro:
I wonder why CMAKE_SOURCE_DIR doesn't cover the file.

PR #714, opened 2025-11-24
changed get() and slicing 

### LINKED ISSUE #397: rowrank is modified after slicing
Example code

````
UT = cytnx.UniTensor.ones([2,3,4]).set_rowrank(2)
UT.print_diagram()
print(UT.rowrank())
UT[:,:,2:4].print_diagram()
print(UT[1,:,:].rowrank())

-----------------------
tensor Name : 
tensor Rank : 3
block_form  : False
is_diag     : False
on device   : cytnx device: CPU
          ---------     
         /         \    
   0 ____| 2     4 |____ 2
         |         |    
   1 ____| 3       |        
         \         /    
          ---------     
2
-----------------------
tensor Name : 
tensor Rank : 3
block_form  : False
is_diag     : False
on device   : cytnx device: CPU
         --------     
        /        \    
        |      2 |____ 0
        |        |    
        |      3 |____ 1
        |        |    
        |      2 |____ 2
        \        /    
         --------     
0
````

We need to have a consistent way to do this.

One possibility is 
* Don't modify rank.
* Reduce the dim in shape, but dim>=1.
* (D1, D2, D3) -->slcing--> (Dp1, Dp2, Dp3) where Di >= Dpi >=1.

### PR DESCRIPTION
**Pull request by manuschneider**
UniTensors keep their metadata (possibly with bonds removed for constant access of one element in a given dimension).
Diagonal UniTensors can be accessed by two-dimensional slicing as well now.

This fixes #397

### CODE REVIEW

**Review by pcchen — COMMENTED** [review]
# PR Review: changed get() and slicing

**Posted by Claude Code on behalf of @pcchen**

## Summary

This PR extends `get()` and slicing for `UniTensor`:
- Diagonal UniTensors can now be sliced with two-dimensional accessors (converting to dense)
- `UniTensor::operator[]` overloads are added
- `Accessor::operator==` is added
- Bond metadata (labels, rowrank) is now preserved through slicing

The approach is sound and well-tested. There are a few bugs and style issues to address before merging.

---

## Critical Issues

### 1. Memory leak in `DenseUniTensor::get()` — diagonal two-accessor path

**`src/DenseUniTensor.cpp`**, in the `else` branch for 2 accessors:

```cpp
DenseUniTensor *out = this->clone_meta();  // allocates raw pointer
...
} else {  // convert to non-diagonal UniTensor
  out->_block = this->_block;
  out->to_dense_();
  return out->get(accessors);  // 'out' is NEVER freed — memory leak
}
```

`out->get(accessors)` returns a **new** `boost::intrusive_ptr<UniTensor_base>` wrapping a different object. The `out` raw pointer is never placed in an `intrusive_ptr`, so it leaks on this path.

**Fix:** wrap `out` before the recursive call:
```cpp
boost::intrusive_ptr<DenseUniTensor> out_ptr(out);
out_ptr->_block = this->_block;
out_ptr->to_dense_();
return out_ptr->get(accessors);
```

---

### 2. Wrong error tag in `DenseUniTensor::set()`

**`src/DenseUniTensor.cpp`**, `set()` method, two-accessor diagonal path:

```cpp
"[ERROR][DenseUniTensor][get] For diagonal UniTensors, only one or two "
```

The tag says `[get]` but this is the `set()` function. Should be `[set]`.

---

## Important Issues

### 3. Type inconsistency: `cytnx_int32` vs `cytnx_int64` for `removed`

**`src/backend/Tensor_impl.cpp`**, `set()` path:

```cpp
-      std::vector<cytnx_int32> remove_id;
+      std::vector<cytnx_int32> removed;   // still int32!
```

The rename was done but the type stayed `cytnx_int32`, inconsistent with `get()` which uses `cytnx_int64`. While this is a local variable with no external impact, it should be consistent.

---

### 4. `_is_braket_form` update is one-sided

**`src/DenseUniTensor.cpp`**, non-diagonal path:

```cpp
if (out->is_tag() && !out->_is_braket_form) {
  out->_is_braket_form = out->_update_braket();
}
```

This only promotes `_is_braket_form` from `false` to `true`, never the reverse. Slicing can also invalidate a previously valid braket form (e.g., if bonds are removed unevenly). Should be:

```cpp
if (out->is_tag()) {
  out->_is_braket_form = out->_update_braket();
}
```

---

## Minor Issues

### 5. Commented-out dead code in `pybind/unitensor_py.cpp`

The old error path is left as commented code in three places:

```cpp
// cytnx_error_msg(true,
// "[ERROR] cannot get element using [tuple] on is_diag=True UniTensor since the block is rank-1, consider [int] or [int:int] instead.%s", "\n");
```

Please remove these.

### 6. Unused variable `cnt` in `pybind/unitensor_py.cpp`

In all three copies of the new diagonal slicing code block:
```cpp
cytnx_uint64 cnt = 0;
for (...) {
    cnt++;  // incremented but never read
```

`cnt` should be removed.

### 7. Triplicated code in `pybind/unitensor_py.cpp`

The diagonal two-accessor handling block is copy-pasted verbatim three times (for `__getitem__`, a second `__getitem__` variant, and `__setitem__`). Consider extracting a helper lambda or function to reduce maintenance burden.

---

## Positive Observations

- The `removed` out-parameter approach for propagating squeezed dimensions from `Tensor_impl` up to `DenseUniTensor` is clean and doesn't break the public API.
- Test coverage is thorough: single/double accessor cases, diagonal/non-diagonal, get/set/operator[], and immutability of source.
- The `AreEqUniTensorMeta` helper is a useful addition.
- Fixing `get/set_block` → `get/put_block` in error messages is a good cleanup.
- The `Accessor::operator==` implementation is correct and well-tested.

---

## Recommended Actions

1. **Fix the memory leak** (item 1) — critical, causes heap corruption in the diagonal two-accessor `get()` path.
2. **Fix the wrong `[get]` tag in `set()`** (item 2).
3. **Fix `_is_braket_form` update** (item 4).
4. Remove commented-out code and unused `cnt` (items 5–6).
5. Consider extracting the duplicated pybind diagonal slice code (item 7).

manuschneider:
1. [x] 1. fixed, create `intrusive_ptr` before return
2. [x] 2. fixed
3. [x] 3. fixed
~~[ ]~~ 4. This does not need to be fixed: a tensor in braket form can never become a non-braket by slicing, even if bonds are removed. 
5. [x] 5. fixed
6. [x] 6. fixed
7. [x] 7. fixed, using Lambda now to avoid repetitive code

pcchen (approved):

PR #722, opened 2025-11-29
Add back getOptimalOrder to python API.

hunghaoti:


IvanaGyro:
> Hi @chiamin and all. I found that this API `getOptimalOrder()` has been comment at the commit [4660bfd](https://github.com/Cytnx-dev/Cytnx/commit/4660bfdf9ffd744607d09f62a43db3d5fdc09002). We add the API `getOrder` and `setOrder` at that time, so we can use `setOrder(optimal = Ture)` directly and then use `getOrder()`, rather than using `getOptimalOrder()`. So I think we can close this PR without any change.

We can remove the commented code if it has been replaced with other APIs.

pcchen:
I will close this without change. Use 'getOrder' with 'setOrder(optimal = Ture)'

PR #723, opened 2025-12-01
Refactor arithmetic ops to use std::variant + std::visit

jysh1214:
Remove Arithmetic_internal.hpp/cpp and Ari_ii from linalg_internal_interface


jysh1214:
> It is also obvious that there are no unit tests covering tricky cases involving type promotions and mixed signed/unsigned etc.

Yeah, you’re right, it’s a critical problem.
I’ve sent [PR#730](https://github.com/Cytnx-dev/Cytnx/pull/730) for the mixed-type unit tests. Please help review it. Thx.

I will solve others issues later.

IvanaGyro (approved):

PR #725, opened 2025-12-05
Transpose exchanges roles of left and right indices

### LINKED ISSUE #367: Inconsistent behavior of Transpose()
While `Tranpose()` permutes the indices for dense UniTensor, such that the matrix representation is actually the transpose of the original matrix representation, this is not the case for symmetric UniTensors (or possibly all UniTensors with directed bonds). For the latter, only the direction  of the bonds is inverted, but no permutation is done.
The permutation should be consistent.

The same happens when using `Dagger()`. 
Btw, the name `.Dagger` for conjugate transpose is very natural for us physicists, but I have not seen it in programming languages/libraries. I know `.Ctranspose()` or `.H()` (numpy). Maybe we should stick to a more common convention like numpy.

manuschneider:
This fixes #367.

New behavior:
1) The order of the indices is inverted.
2) The rowrank is adapted accodingly.
3) Incoming bonds become outgoing ones and vice versa (redirect).

Unit tests are adapted accordingly. Some corrections in the unit tests were also done where I saw them (replacing `size_t` by `cytnx_int64` etc.

### DIFF
--- a/include/Bond.hpp
+++ b/include/Bond.hpp
@@ -799,7 +799,7 @@ namespace cytnx {
 
     /**
     @brief Group the duplicated quantum number and return the new instance
-            of the Bond ojbect.
+            of the Bond object.
         @details This function will group the duplicated quantum number and return
             the new instance of the Bond object. It will also the \p mapper, where
                 \p mapper is about the new index from old index via\n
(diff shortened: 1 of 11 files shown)


manuschneider:
The index order is now inverted for all UniTensor types, as discussed in #367.

PR #730, opened 2025-12-06
Add mixed-type GPU arith ops tests and fix Cpr/Mod dispatcher

jysh1214:
- Skip Int64 vs Int16 Cpr test due to unresolved issue

### DIFF
--- a/tests/gpu/linalg_test/Add_test.cpp
+++ b/tests/gpu/linalg_test/Add_test.cpp
@@ -126,6 +126,33 @@ namespace AddTest {
     }
   }
 
+  // Test tensor-to-tensor addition with mixed types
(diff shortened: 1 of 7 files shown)


tests/gpu/linalg_test/Add_test.cpp
```diff
@@ -126,6 +126,33 @@ namespace AddTest {
     }
   }
 
+  // Test tensor-to-tensor addition with mixed types
+  TEST_P(AddTestAllShapes, gpu_tensor_add_tensor_mixed_types) {
```
IvanaGyro: It seems that this test already covers `TEST_P(AddTestAllShapes, gpu_tensor_add_tensor_all_types)`. Is it fine the remove the following one?

jysh1214:
Let me modify each unit tests to use mioxed types one by one.

PR #733, opened 2025-12-11
added pybind for UniTensor.apply() and apply_()

manuschneider:


pcchen (commented):

**Overall:** Correct and minimal. The bindings themselves are fine. A few things worth raising:

---

### Issue — `apply_()` returns by value, inconsistent with other inplace methods

The public C++ `UniTensor::apply_()` (line 3725 of `UniTensor.hpp`) returns `UniTensor` by value, while every other inplace method (`contiguous_()`, `permute_()`, etc.) returns `UniTensor&`. This means in Python, chaining `x.apply_().something()` would operate on the returned copy, not `x`. This is a pre-existing C++ issue but the pybind exposure makes it user-visible. It should be fixed in `UniTensor.hpp` to return `UniTensor&`, consistent with the rest.

---

### Missing — Python tests

The C++ tests in `BlockFermionicUniTensor_test.cpp` cover `apply()`/`apply_()` on the C++ side, but there are no Python-side tests verifying the new bindings work. A basic smoke test with a `BlockFermionicUniTensor` in the Python test suite would catch any binding-level issues.

---

### Minor — No docstrings in pybind

Other method bindings pass a docstring as the third argument (or use `py::doc()`). The two new `.def` calls lack these, so `help(ut.apply)` in Python gives no documentation despite the C++ Doxygen being comprehensive.

---

### Minor — `signflip_` still commented out

Line 590 of `unitensor_py.cpp` has:
```cpp
// .def("signflip_", &UniTensor::signflip_)
```
`signflip_` is the complement to `apply`. If the intent is to expose the fermionic signflip API, this should be unblocked at the same time.

---

### Positive

- The two bindings are correct — no argument wrappers needed since neither method takes parameters.
- `apply()` on non-fermionic tensors is a no-op (returns self), so there are no correctness risks for `DenseUniTensor` or `BlockUniTensor`.

PR #711, opened 2025-11-21
Release to PyPI

IvanaGyro:
## Summary

Stand up a PyPI release pipeline (linux + macOS) via cibuildwheel and bring
in the supporting CI / build-system changes.

### Build / packaging
- **`build: rework cibuildwheel helpers and pyproject configuration`**
  - `pyproject.toml`: drop the unused `setuptools` build-system dep, share
    one `build/` directory across wheel tags via
    `[tool.cibuildwheel.config-settings]`, and overhaul Linux / macOS
    `environment` and `environment-pass` to forward `CCACHE_*` and
    `CMAKE_*_COMPILER_LAUNCHER` into the manylinux container and pin a
    stable compiler / `MACOSX_DEPLOYMENT_TARGET` on macOS.
  - `tools/cibuildwheel_before_all.sh`: install build deps via `dnf` |
    `apk` for manylinux / musllinux.
  - `tools/cibuildwheel_before_all_macos.sh` *(new)*: probe Homebrew
    dylibs (including transitive deps) for their `minos` and persist the
    highest one as `MACOSX_DEPLOYMENT_TARGET` so wheels don't silently
    encode a newer minimum than `delocate` can satisfy.
  - `tools/cibuildwheel_before_build.py`: env dump + `ccache --show-config`
    for diagnostics.
  - `CMakeLists.txt`: log `CMAKE_BINARY_DIR` / `PROJECT_BINARY_DIR` to help
    diagnose ccache `base_dir` mismatches under cibuildwheel.
  - `.gitignore`: ignore the local `.ccache/` directory used by helpers.

### Workflows
- **`ci: revamp release_pypi workflow with push triggers, ccache mounting, and TestPyPI publish`**
  Trigger on push to master and `v*` tags in addition to PRs / dispatch.
  Pin runners (`ubuntu-24.04`, `ubuntu-24.04-arm`, `macos-14`,
  `macos-15-intel`). Bind-mount the host ccache into the manylinux
  container via `CIBW_CONTAINER_ENGINE` and forward `CCACHE_DIR` /
  `CCACHE_BASEDIR` / `CCACHE_COMPILERCHECK`. Add a `ReleaseTestPyPI` job
  that downloads the wheels and publishes to TestPyPI for PRs and tag
  pushes. New `tools/validate_ccache_stats.py` is wired up as
  `CIBW_TEST_COMMAND` to assert ccache hits per build.
- **`ci: enable PR runs and bump conda_build runner to macos-15-intel`**
[...]

PR #741, opened 2026-01-31
Fix: expand ~ in CCACHE_CONFIGPATH to actual home directory

b95702041:
- Use os.path.expanduser() to properly expand ~ in ccache config path
- Fixes FileNotFoundError on macOS builds in cibuildwheel
- Resolves CI failure in PR #711


pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["scikit-build-core >=0.11", "pybind11 >=3.0"]
+requires = ["scikit-build-core >=0.11", "pybind11 >=3.0", "setuptools"]
```
IvanaGyro: Why setuptools is needed?

tools/cibuildwheel_before_all.sh
```diff
@@ -1,18 +1,36 @@
...
-# yum install -y arpack-devel boost-devel ccache lapack-devel libomp-devel openblas-devel
-yum install -y arpack-devel boost-devel ccache libomp-devel openblas-devel
-# On Red Hat/CentOS systems (which manylinux images are based on), the lapack-devel
-# package typically provides the Fortran bindings and miss the C-interface header, so
-# use lapacke.h shipped with openblas instead.
-ls /usr/include
-ln -s /usr/include/openblas/lapacke.h /usr/include/lapacke.h
-ln -s /usr/include/openblas/lapack.h /usr/include/lapack.h
-ln -s /usr/include/openblas/lapacke_mangling.h /usr/include/lapacke_mangling.h
+# Install ccache
```
IvanaGyro: Why not to install ccache by the package manager?

IvanaGyro (commented):
Great work! I have some questions.

pyproject.toml
```diff
@@ -38,10 +38,12 @@ wheel.packages = ["cytnx"]
 cmake.args = [
     # Use the default preset defined in CMakePresets.json
     "--preset=openblas-cpu",
+    "-DBoost_NO_BOOST_CMAKE=ON",
     # TODO: Remove setting of generator to support builuding on Windows.
     # scikit-build uses Ninja as the default cmake generator. However, Ninja
     # conflicts against the build process of HPTT.
     "-G Unix Makefiles",
+    "-DBoost_NO_BOOST_CMAKE=ON",
```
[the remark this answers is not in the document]
b95702041: removed the duplicates and modified the cibuildwheel env.

pyproject.toml
IvanaGyro: Can you try to figure out which package provide the outdated `FindPythonLibsNew.cmake`? pybind11 has fixed the issue about `distutils` few years ago: https://github.com/pybind/pybind11/issues/3677

b95702041: The file `./cmake/Modules/FindPythonLibsNew.cmake` was copied from pybind11 on January 6, 2020 and has never been updated since.

b95702041: I checked the git history:

`gitlog --follow ./cmake/Modules/FindPythonLibsNew.cmake`

`commit 4af776705ff03bf6ef32a2de32b6ef58e053a214 Author: kaihsin [kaihsinwu@gmail.com](mailto:kaihsinwu@gmail.com) Date: Mon Jan 6 01:22:33 2020 -0500 update, merge complete, clean container base
`

This file was manually added by kaihsin in 2020, not installed from the pybind11 package.

IvanaGyro: I think its fine to remove it because we have bound with pybind11 now.

pcchen (commented):
# PR #741 Review — Fix: expand ~ in CCACHE_CONFIGPATH

## Critical Issues (1)

**1. `expanduser` called before null check — `TypeError` if env var is unset**
`tools/cibuildwheel_before_build.py`

```python
ccache_config_path = os.getenv('CCACHE_CONFIGPATH')

# Expand ~ to actual home directory
ccache_config_path = os.path.expanduser(ccache_config_path)  # ← called with None if var unset
if not ccache_config_path:
    raise RuntimeError('The CCACHE_CONFIGPATH environment variable must be set.')
```

`os.getenv()` returns `None` if the variable is not set. `os.path.expanduser(None)` raises `TypeError` in Python 3, masking the real error message. The null check must come first:

```python
ccache_config_path = os.getenv('CCACHE_CONFIGPATH')
if not ccache_config_path:
    raise RuntimeError('The CCACHE_CONFIGPATH environment variable must be set.')
ccache_config_path = os.path.expanduser(ccache_config_path)
```

---

## Important Issues (3)

**2. `libomp-devel` dropped from `dnf`/`yum` install**
`tools/cibuildwheel_before_all.sh`

The old script installed `libomp-devel` (OpenMP). The new `dnf` and `yum` branches omit it. If the C++ build uses OpenMP (`-fopenmp`), this will cause silent link failures. Only the `apk` branch doesn't need it because Alpine ships OpenMP with GCC. Add `libomp-devel` back to the `dnf`/`yum` branches.

**3. `.bak` file committed to git**
`cmake/Modules/FindPythonLibsNew.cmake.bak`

Renaming `FindPythonLibsNew.cmake` → `FindPythonLibsNew.cmake.bak` commits a backup artifact to the repo. Either delete the file if it's no longer needed, or keep the original name. A `.bak` file in `cmake/Modules` will confuse future developers.

**4. `Boost_NO_BOOST_CMAKE=ON` only applied to Linux, not macOS**
`pyproject.toml`

`CMAKE_ARGS = "-DBoost_NO_BOOST_CMAKE=ON"` is set in `[tool.cibuildwheel.linux]` environment but not `[tool.cibuildwheel.macos]`. Combined with the `CMakeLists.txt` change removing `CONFIG` from `find_package(Boost REQUIRED)`, this may behave inconsistently: Linux forces the find-module path, macOS uses whatever CMake defaults to with Homebrew Boost. If macOS already works, add a comment explaining why the flag is not needed there.

---

## Minor Issues (2)

**5. `~` in `CCACHE_CONFIGPATH` is fragile**
`pyproject.toml`: `CCACHE_CONFIGPATH = "~/ccache.conf"` — cibuildwheel passes this as a literal string; shell expansion does not apply. The Python fix makes it work, but `$HOME/ccache.conf` would be more conventional and self-documenting.

**6. `setuptools` added without version pin**
`pyproject.toml`: `"setuptools"` added to `build-system.requires` with no version constraint. This may cause non-reproducible builds as setuptools makes breaking changes. Consider `"setuptools>=68"` or similar.

---

## Strengths

- The `execinfo.h` / `HAS_EXECINFO` guard in `cytnx_error.hpp` is correct and necessary for non-glibc platforms (macOS, musl/Alpine).
- Multi-distro `apk`/`dnf`/`yum` dispatch in `cibuildwheel_before_all.sh` is clean and extensible.
- `ln -sf` instead of `ln -s` prevents errors on re-runs.
- `manylinux_2_28` downgrade is a reasonable compatibility improvement.

tools/cibuildwheel_before_all.sh
```diff
@@ -1,18 +1,24 @@
...
-ls /usr/include
-ln -s /usr/include/openblas/lapacke.h /usr/include/lapacke.h
-ln -s /usr/include/openblas/lapack.h /usr/include/lapack.h
-ln -s /usr/include/openblas/lapacke_mangling.h /usr/include/lapacke_mangling.h
+# Install required packages based on distro
+if command -v apk &> /dev/null; then
+    # musllinux (Alpine)
+    apk add --no-cache boost-dev openblas-dev arpack-dev ccache
+elif command -v dnf &> /dev/null; then
+    # manylinux_2_28+ (AlmaLinux/RHEL)
```
IvanaGyro, line 8: @codex we will only support/use manylinux_2_28+ just leave dnf

pyproject.toml
IvanaGyro: @codex remove "setuptools" and cmake/Modules/FindPythonLibsNew.cmake.bak to resolve the problem in this thread

IvanaGyro:
@b95702041 I will take over this PR.

pyproject.toml
IvanaGyro: @codex remove "setuptools" and cmake/Modules/FindPythonLibsNew.cmake.bak to resolve the problem in this thread

tools/cibuildwheel_before_all.sh
IvanaGyro: @codex resolve the problem in this thread

PR #749, opened 2026-03-18
implemented Trace for GPU

### LINKED ISSUE #735: Missing GPU implementation for Trace
With commit c5dd9c4be2bc4d935434e048002f772e602b24cb

When I try to perform trace for a UniTensor, I got RuntimeError
````
RuntimeError: 
# Cytnx error occur at cytnx::Tensor cytnx::linalg::Trace(const cytnx::Tensor&, const cytnx::cytnx_uint64&, const cytnx::cytnx_uint64&)
# error: [ERROR][Trace] GPU is under developing.

# file : /home/pcchen/github/Cytnx-dev/Cytnx/src/linalg/Trace.cpp (91)
````

### PR DESCRIPTION
**Pull request by manuschneider**
Currently, Tensor.Trace corresponds to creating a UniTensor and contracting it with a (diagonal) identity UniTensor;
removed Nomp argument in internal Trace functions;

This fixes #735

### CODE REVIEW

**Comment by ianmccul** [comment]
I think there is no reason to ever allow trace over a non-square tensor.  By definition that means that the left right indices represent different spaces, so the trace is meaningless.  In the special case where one space is a diagonal projection $P$ of the other space, it might make sense to calculate something like $\text{Tr} PM$ and optimize that if necessary.

**Comment by manuschneider** [comment]
An error message is now thrown in all cases where the index dimensions of the traced indices do not match.

File: src/linalg/Trace.cpp [hunk 1]
```diff
@@ -55,23 +59,30 @@ namespace cytnx {
       vector<cytnx_uint64> accu;
       shape.erase(shape.begin() + ax2);
       shape.erase(shape.begin() + ax1);
-      // 2) get out put elementsize.
+      // 2) get output element size.
       cytnx_uint64 Nelem = 1;
       for (int i = 0; i < shape.size(); i++) Nelem *= shape[i];
-      // 3) get diagonal element numbers:
-      cytnx_uint64 Ndiag = Tn.shape()[ax1] < Tn.shape()[ax2] ? Tn.shape()[ax1] : Tn.shape()[ax2];
```
IvanaGyro, line 62: This is a breaking change from the view of Python API, we may have to deprecate it before removing it.

pcchen:
I will land this PR #749  as-is to unblock GPU users. Then we do the follow up refactor #834.

pcchen (commented):
## PR #749 Review

### Overview

This PR adds GPU support for `Tensor::Trace` by routing both 2D and n-D cases through `UniTensor::eye + Contract`, mirroring the existing CPU n-D approach. It also removes the unused `Nomp` parameter from all `Trace_internal` signatures, adds a missing axis-dimension mismatch check, and unskips the GPU trace tests.

---

### Strengths

- **Old `_trace_2d_gpu` was silently wrong**: The previous implementation dereferenced GPU device memory from the host (`rawdata[i * Ldim + i]` on a GPU tensor), which is UB. The `UniTensor::eye + Contract` approach runs entirely on-device via cuBLAS. This is a correctness fix, not just a feature.
- **Old `_trace_nd_gpu` used `zeros` instead of `eye`**: The old code created a zero-filled identity tensor (`UniTensor(zeros(...))`) — which would produce all-zero results for the n-D trace. The new `UniTensor::eye` is correct.
- **New axis-mismatch validation** (`src/linalg/Trace.cpp:618`): Previously `Ndiag = min(shape[ax1], shape[ax2])` silently accepted non-square traces. The explicit error message is the right behavior.
- **`Ndiag` simplification**: After the shape-match guard, `Tn.shape()[ax1]` is used directly — correct and simpler.
- **`cuTrace_ii` dispatch table** is fully registered for all 11 types.

---

### Issues

**Medium priority:**

1. **Spurious blank lines in both `.hpp` files** — A formatting artifact from removing `const int &Nomp,`. The line break that previously followed `Nomp` now creates an empty line between `Ndiag` and `Nelem` in 9 out of 11 declarations in both `Trace_internal.hpp` and `cuTrace_internal.hpp`. Example from `Trace_internal.hpp:157`:
   ```cpp
   // current (odd-looking)
   void Trace_internal_d(const bool &is_2d, Tensor &out, const Tensor &Tn,
                         const cytnx_uint64 &Ndiag,

                         const cytnx_uint64 &Nelem, ...

   // should be
   void Trace_internal_d(const bool &is_2d, Tensor &out, const Tensor &Tn,
                         const cytnx_uint64 &Ndiag, const cytnx_uint64 &Nelem, ...
   ```
   Affects `_d`, `_f`, `_u64`, `_i64`, `_u32`, `_i32`, `_u16`, `_i16`, `_b` in both headers.

2. **`_trace_2d_gpu` uses `UniTensor::eye + Contract` for a 2D trace** (`cuTrace_internal.cu:253`). For a simple `tr(A)`, allocating an identity matrix and running a full contraction is heavier than a diagonal reduction kernel (e.g. `cublasDdot` with stride `N+1`). The old CPU path used a direct loop. This is a performance regression in the 2D case in exchange for simplicity and correctness. Fine as a first pass, but worth a follow-up issue noting the optimization opportunity.

3. **`set_labels` in GPU path vs `relabel_` in CPU path** (`cuTrace_internal.cu:263` vs `Trace_internal.cpp:28`). Both set labels on a freshly constructed `I_UT` so the effect is identical, but using different APIs for equivalent operations in symmetric code paths is confusing. Should use one consistently — `relabel_` matches the CPU.

**Low priority:**

4. **No test for the new axis-mismatch error** (`Trace.cpp:618`). A test like:
   ```cpp
   Tensor A = Tensor({3, 4}, Type.Double);
   EXPECT_THROW(cytnx::linalg::Trace(A, 0, 1), std::runtime_error);
   ```
   would pin the new guard against regression.

5. **Dead `return out` after `cytnx_error_msg(true, ...)`** (`Trace.cpp:657`, `682`). `cytnx_error_msg` with a `true` condition always throws, so the `return out` on the next line is unreachable. Harmless, but slightly misleading.

6. **No GPU numerical correctness test**. The unskipped tests verify the GPU trace runs without crashing, but no test explicitly checks `GPU_result ≈ CPU_result` for a known tensor. A simple `EXPECT_NEAR` comparison between CPU and GPU trace of the same tensor would catch any numerical divergence.

---

### Minor Nits

- `Trace.cpp:630`: Comment fixed from "elementsize" → "element size". Good.
- `cuTrace_internal.cu:273`: Comment "// currently identical to CPU version" is accurate and useful.
- The `#ifdef UNI_GPU` guard structure in `Trace.cpp` is correct — CPU-only builds hit the `cytnx_error_msg` branch before returning.

---

### Summary

The core implementation is correct and fixes two pre-existing bugs (host dereference of GPU memory, zero tensor used as identity). The main items worth addressing before merge are the `.hpp` formatting artifacts (item 1, easy cleanup) and the missing test for the new error condition (item 4). The performance note (item 2) is better tracked as a follow-up issue.

pcchen (approved):

Issue #763, opened 2026-03-31
linalg.Svd produces wrong (negative) auxiliary bond qnums


chiamin:
This is a similar bug to #761.

linalg.Svd on T=s,U,Vd produces auxiliary bonds with wrong negative qnums when the bond directions are [OUT, IN, OUT] with rowrank=2. In the resulting U, no sector satisfies the QN conservation.

I attached an example to demonstrate the bug.
[cytnx_svd_wrong_qnums.py](https://github.com/user-attachments/files/26377631/cytnx_svd_wrong_qnums.py)

Environment: cytnx 1.0.0, Python 3.9, Linux

manuschneider:
Fixed in #764.

PR #764, opened 2026-04-01
fixed reversed qnums infixed reversed qnums in linalg functions if combined incoming bond BdLeft is outgoing instead of incoming (#761 #763) linalg functions

### LINKED ISSUE #761: linalg.Qr produces wrong (negative) auxiliary bond qnums for BlockUniTensor
linalg.Qr on a BlockUniTensor, T=QR, produces an auxiliary bond with wrong negative qnums when the bond directions are [OUT, IN, OUT] with rowrank=2. In the resulting R, **no sector pair satisfies QN conservation**, yet cytnx stores 2 nonzero blocks without raising an error.

I attached an example to demonstrate the bug.
[cytnx_qr_wrong_qnums.py](https://github.com/user-attachments/files/26362245/cytnx_qr_wrong_qnums.py)

I suggest that any QN tensor needs to pass a QN check whenever it is created.

Environment: cytnx 1.0.0, Python 3.9, Linux

manuschneider:
This happens if the combined incoming bond `BdLeft` is outgoing instead of incoming.
Fixes #761 and #763


pcchen (commented):
# PR Review: Cytnx-dev/Cytnx#764

**Fix assessment:** Correct — using `calc_reverse_qnums()` when `BdLeft.type() == BD_OUT` is the right approach and addresses #761 / #763.

---

## Important Issues

### 1. `calc_reverse_qnums()` called inside the loop — O(Nblocks) recomputation
**Files:** All 9 patched locations (`Eig.cpp:191`, `Eigh.cpp:184`, `ExpH.cpp:158`, `ExpM.cpp:144`, `Gesvd.cpp:240/466`, `Qr.cpp:221`, `Svd.cpp:249/474`)

In the `else` branch, `calc_reverse_qnums()` is called on every loop iteration, recomputing the full reversed vector each time. It should be hoisted out:

```cpp
const auto left_qnums = (BdLeft.type() == bondType::BD_IN)
                         ? BdLeft.qnums()
                         : BdLeft.calc_reverse_qnums();
for (cytnx_uint64 b = 0; b < Tin.Nblocks(); b++) {
  mgrp[left_qnums[new_itoi[b][0]]].push_back(b);
}
```

Note: `qnums()` likely returns a const ref while `calc_reverse_qnums()` returns by value, so use `auto` (not `const auto&`) to avoid a dangling reference.

### 2. Missing regression tests for 6 of 7 patched functions
**Affects:** Eig, Eigh, ExpH, ExpM, Gesvd, Qr

Only `Svd` got a regression test. The same `BD_OUT BdLeft` bug class exists in all other patched functions and is entirely untested. This is especially risky for:
- **Gesvd** and **Qr** — used heavily in DMRG/TDVP; wrong qnums silently corrupt the subspace and propagate through the simulation
- **ExpH** and **ExpM** — no Block UniTensor tests exist at all in those test files; this is a good opportunity to add them

The SVD test pattern is a good template — each new test should verify both `qnums()` and `getDegeneracies()` on the auxiliary bond (not just `ReComposeCheck`).

---

## Suggestions

### 3. Bare `else` — no guard for unexpected bond types

The `else` branch silently handles any non-`BD_IN` type. An explicit guard would be more defensive:

```cpp
} else if (BdLeft.type() == bondType::BD_OUT) {
  // ...
} else {
  cytnx_error_msg(true, "[ERROR] unexpected BdLeft bond type in block categorization", "\n");
}
```

### 4. SVD test uses a degenerate single-sector tensor
**File:** `tests/linalg_test/Svd_test.cpp`

The regression test has only one qnum sector (`{4}` with degeneracy 1). A multi-sector test (e.g. `Bond(BD_OUT, {{0},{1},{2}}, {2,3,2}, {Symmetry::U1()})`) would give higher confidence that `calc_reverse_qnums()` indexing is correct across all sectors.

---

## Strengths

- The fix logic is correct and well-understood.
- The SVD regression test explicitly checks both qnum values and degeneracies on the auxiliary bond — stronger than a pure round-trip recompose check.
- The fix is applied consistently across all 9 locations.

---

## Recommended Action

1. **Before merge:** Hoist `calc_reverse_qnums()` out of the loop (applies to all 9 sites).
2. **Before merge:** Add regression tests for at minimum Gesvd and Qr (highest real-world impact).
3. **Can follow up:** Tests for Eig, Eigh, ExpH, ExpM in a follow-on PR.
4. **Optional:** Add defensive `else if` guard.

---

manuschneider:
1. Resolved
2. & 4. I implemented the case in #763 and more unit tests.
3. We currently only support types `BD_IN` or `BD_OUT` for a `BlockUniTensor`, so need for a check here.

Ready for a final review and merge.

tests/linalg_test/GeSvd_test.cpp
```diff
@@ -44,6 +44,28 @@ namespace GesvdTest {
...
+  describe:Multi-sector regression for the case where the combinded bond BdLeft is outgoing.
+  rowrank=2 with a leading BD_OUT bond makes the combined left bond BD_OUT, and several qcharge
+  sectors survive in the auxiliary bond.
+  ====================*/
+  TEST(Gesvd, U1_sym_regression_multi_sector_bd_out_left) {
+    fail_msg.Init(UnitTest::GetInstance()->current_test_info()->name());
+    UniTensor T({Bond(BD_OUT, {{0}, {1}}, {2, 2}, {Symmetry::U1()}),
+                 Bond(BD_IN, {{4}, {5}}, {2, 2}, {Symmetry::U1()}),
+                 Bond(BD_OUT, {{4}, {5}}, {2, 2}, {Symmetry::U1()})},
+                {"a", "b", "c"}, 2, Type.Double, Device.cpu, false);
```
IvanaGyro, line 57: Will `rowrank` work properly under the OUT-IN-OUT setting?

pcchen:
## Issue Status Check

Here is the current status of all raised issues:

---

### Important Issues (Before Merge)

| Issue | Status |
|-------|--------|
| Hoist `calc_reverse_qnums()` out of the loop (all 9 sites) | ✅ Confirmed fixed — `const auto left_qnums` is computed before the loop in all patched locations |
| Add regression tests for Gesvd and Qr | ✅ Done — `U1_sym_regression_multi_sector_bd_out_left` for Gesvd, `BkUt_Qr_reversed_qnums` for Qr; both check qnums, degeneracies, and recompose |
| Upgrade SVD test from single-sector to multi-sector | ✅ Done — two new multi-sector tests added |

### Can Follow Up (Not Blocking)

| Issue | Status |
|-------|--------|
| Add BD_OUT regression tests for Eig, Eigh, ExpH, ExpM | ⏩ Still missing — deferred to a follow-up PR as agreed |

### Suggestions

| Issue | Status |
|-------|--------|
| Add defensive `else if` guard for unexpected bond types | 🚫 Won't fix — only `BD_IN`/`BD_OUT` are supported types for `BlockUniTensor` |

---

### Open Question

@IvanaGyro asked whether `rowrank` works correctly under the `OUT-IN-OUT` bond layout used in the new `GeSvd` regression test (`GeSvd_test.cpp:52–67`). This has not been answered yet.

@manuschneider — could you confirm whether `rowrank=2` is handled correctly with the `BD_OUT, BD_IN, BD_OUT` structure in that test?

---

**Bottom line:** The two blocking requirements are resolved. This PR is ready to merge once the merge conflict is cleared and the `rowrank`/OUT-IN-OUT question above is addressed.

tests/linalg_test/GeSvd_test.cpp
manuschneider: It should, and the test confirms it.
The bug fixed in this PR occurred whenever the first bond was of type `BD_OUT`; this is now fixed. Mixed directions for inputs and outputs are supported and important for actual algorithms, so this is a meaningful test. If two bonds of different directions are combined, one of them has their qnums reversed.

manuschneider:
I added unit tests for all linalg functions in this PR. Everything is resolved and covered now.

pcchen (approved):

PR #768, opened 2026-04-03
Normalize ZnSymmetry outputs into valid qnums

### LINKED ISSUE #766: Z2 combine_rule can produce -1
I am not sure if this is a bug or is on purpose, but the current `combine_rule` for Z2 symmetry can produce 0, 1, or -1, while -1 quantum number is now allowed in creating a bond. To me, it is more natural to restrict Z2 qn to only {0,1}.

I attached an example code as follows

```
import cytnx

sym = cytnx.Symmetry.Zn(2)

# combine_rule can give 0, 1 or -1
print("=== Z2 combine_rule ===")
for a in range(-1, 3):
    for b in range(-1, 3):
        result = sym.combine_rule(a, b)
        print("  combine(", a, ",", b, ") =", result)

# However, Bond rejects qnums -1 ---
b_bad = cytnx.Bond(cytnx.BD_IN, [[-1], [0]], [1, 1], [sym]) # raise error
```

yingjerkao:
Fixes #766.

## Summary
- normalize `ZnSymmetry` scalar and vector `combine_rule` outputs into `[0, n)`
- normalize `reverse_rule` output so public results satisfy `check_qnum`
- remove the signed-modulo behavior that could produce `-1` for `Z2` and other invalid values on reverse
- add a public API regression for `Symmetry::Zn(2)` covering both combine and reverse paths

## Verification
- `./build/openblas-cpu/tests/test_main --gtest_filter="Bond.ZnSymmetryNormalizesPublicRules"`
- `./build/openblas-cpu/tests/test_main --gtest_filter="Bond.InitWithQnum_v2:Bond.CombindBondSymm_v2:Bond.ZnSymmetryNormalizesPublicRules"`


ianmccul:
I can't see anything obviously wrong with the PR, but just a question of the semantic meaning of a Zn quantum number.  The current code allows numbers outside the range `[0,n)` but normalizes outputs to that range with the modulo function.  If I was designing this I would make numbers outside the range `[0,n)` erroneous and an error if it is passed into a function expecting a Zn quantum number.

IvanaGyro (approved):

PR #777, opened 2026-04-09
removed docs/uv.lock from repo

manuschneider:
docs/uv.lock is a temporary file that gets created by `uv sync`. It should not be in the repo so I removed it.


pcchen (approved):

IvanaGyro:
It's not a temporary file. For packages which want to lock the versions of all dependencies, uv.lock should be committed. The package under docs is not published or shared by the use of Cytnx.

manuschneider:
Should it be recovered then? There were some conflicts of the versions in certain branches.

IvanaGyro:
If different version of dependencies will generate different version of files which are committed in git, then uv.lock should be committed. I am not sure how docs are compiled.

PR #799, opened 2026-05-06
cleaned up error messages

manuschneider:


src/backend/StorageImplementation.cpp
```diff
@@ -222,8 +223,8 @@ namespace cytnx {
         }
 #else
         cytnx_error_msg(
-          1, "%s",
-          "[ERROR][Internal] Storage.to_. the Storage is as GPU but without CUDA support.");
+          true, "[ERROR][Internal][Storage.to_] The Storage is on GPU but without CUDA support.%s",
+          "n");
```
IvanaGyro: \n, same as other places.

src/RegularGncon.cpp
```diff
@@ -389,7 +389,7 @@ namespace cytnx {
         for (int i = 0; i < TN_names.size(); i++) {
           cout << "        " << TN_names[i] << endl;
         }
-        cytnx_error_msg(true, "%s", "\n");
+        cytnx_error_msg(true, "[ERROR] TNs appeared in ORDER line, but are not defined.%s", "\n");
```
IvanaGyro: The error message here is duplicated. same as src/RegularNetwork.cpp

manuschneider: The networks print out further information to cout, before cytnx_error_msg throws an error and writes to cerr.

src/backend/StorageImplementation.cpp
manuschneider: Thanks, fixed.

src/RegularGncon.cpp
IvanaGyro: Yes, that is what I meant. The current change will duplicate the message was printed to stdout. There are two options for cleaning this up.

option 1: compose the error message before outputting
```c++
if (!TN_names.empty()) {
  std::ostringstream oss;

  oss << "Following TNs appeared in ORDER line, but are not defined.\n";
  for (const auto& name : TN_names) {
    oss << "        " << name << '\n';
  }

  const std::string msg = oss.str();

  // Works with the current macro, but unsafe if msg.size() >= about 512.
  cytnx_error_msg(true, "%s", msg.c_str());
}
```
However, notice that the length of message will be limited by the current implementation of `cytnx_error_msg`. The max length of the composed string is 512. See: https://github.com/Cytnx-dev/Cytnx/blob/14ef89a7b4051dfd65e407ae79dfd7d6acfd8ead/include/cytnx_error.hpp#L38

option2: just output the whole message to stderr and let `cytnx_error_msg` pass "\n" only before updating the implementation of `cytnx_error_msg`
```c++
        cerr << "[ERROR] Following TNs appeared in ORDER line, but is not defined." << endl;
        for (const auto& name : TN_names) {
          cerr << "        " << name  << endl;
        }
```

manuschneider: I went for option 2.

IvanaGyro (approved):

PR #806, opened 2026-05-14
implemented to_dense for Block(Fermionic)UniTensor

manuschneider:
also works for all types if the UniTensor is already non-diagonal


pcchen:
## PR #806 Review: `to_dense` / `to_dense_` for Block(Fermionic)UniTensor

### Overview

This PR adds `to_dense()` and `to_dense_()` to `BlockUniTensor` and `BlockFermionicUniTensor` (previously unimplemented), and relaxes the `DenseUniTensor` variants from throwing on non-diagonal input to being no-ops. The logic is uniform across all three types: if `is_diag`, expand each block via `linalg::Diag`; otherwise return a clone / do nothing.

---

### Correctness

**No issues** with the core logic. The block-by-block `Diag` expansion is correct, `_is_diag` is cleared in both paths, and the `clone_meta(true, true)` call in the diag path correctly preserves bonds, labels, rowrank, and `_signflip` for the fermionic variant.

The fermionic test explicitly verifies that `_signflip` flags survive `to_dense()` unchanged — good.

---

### Issues

#### 1. Behavioral breaking change for `DenseUniTensor` (minor, intentional)

`DenseUniTensor::to_dense()` and `to_dense_()` previously threw `std::logic_error` on non-diagonal input. They now silently return a clone / no-op. This is the right direction for generic code paths, and the tests have been updated, but it is a **user-visible API change**. Any downstream code that used the exception as a guard (`try { t.to_dense(); } catch (...) { /* it's not diagonal */ }`) will now silently pass. Worth noting in the PR description or a changelog comment.

#### 2. Missing complex dtype test for Block types

All new Block tests use `Type.Double`. There is no test for `Type.ComplexDouble`. Since `linalg::Diag` dispatches on dtype, a `ComplexDouble` diagonal block should expand correctly, but it's untested. A single `astype(Type.ComplexDouble)` variant of the existing `to_dense_diag` tests would close this gap.

#### 3. `clone()` in the non-diag path — always a full data copy

When `is_diag == false`, both `BlockUniTensor::to_dense()` and `BlockFermionicUniTensor::to_dense()` call `this->clone()`, which copies all block data. For large tensors this is O(total elements). This is safe and correct, but for the case where the caller just wants to ensure a non-diag representation (e.g. `T = T.to_dense()`), they silently pay the copy cost. Consider documenting this in a function-level comment, or at minimum noting it in the PR description so users are not surprised.

---

### Suggestions

#### 4. Code duplication between `BlockUniTensor` and `BlockFermionicUniTensor`

The two implementations of `to_dense()` and `to_dense_()` are byte-for-byte identical except for the class name. This is unavoidable given the inheritance structure (neither inherits from the other), so no action required — just noting for awareness.

#### 5. `to_dense_()` non-diag path — silently does nothing

`BlockUniTensor::to_dense_()` and `BlockFermionicUniTensor::to_dense_()` do nothing when `is_diag == false`, consistent with the updated `DenseUniTensor`. This is the correct behavior for an in-place call on an already-dense tensor, and the tests verify it.

---

### Test Coverage

| Case | `BlockUniTensor` | `BlockFermionicUniTensor` | `DenseUniTensor` |
|---|---|---|---|
| `to_dense()` on diag | ✅ | ✅ | (existing) |
| `to_dense_()` on diag | ✅ | ✅ | (existing) |
| `to_dense()` on non-diag | ✅ | ✅ | ✅ updated |
| `to_dense_()` on non-diag | ✅ | ✅ | ✅ updated |
| `_signflip` preserved | — | ✅ | — |
| ComplexDouble dtype | ⚠️ missing | ⚠️ missing | — |
| Multi-sector bonds | ✅ | ✅ | — |

---

### Strengths

- Implementation is compact and consistent across all three types.
- The fermionic test is thorough: it verifies both the block values and the `_signflip` flags after expansion, and covers both `to_dense()` and `to_dense_()`.
- Updated `DenseUniTensor` tests correctly reflect the new no-op semantics instead of deleted `EXPECT_THROW` lines.

---

### Recommendation

**Approve after addressing:**
1. Add a `ComplexDouble` variant of the `to_dense_diag` test for at least one of the Block types.
2. (Optional) Note the DenseUniTensor behavior change in the PR description so it's visible in the git history.

manuschneider:
> Posted by Claude Code on behalf of @pcchen

1. Intended
2. Added
3. I double checked, and chose the behavior of `contiguous`, `to` and `as_type`: returns itself if the UniTensor is already non-diagonal instead of returning a clone.
4. See #844
5. Intended

All issues addressed.

pcchen:
## Final Review: PR #806 — `to_dense` / `to_dense_` for Block(Fermionic)UniTensor

All issues raised in the prior review have been addressed:

| Issue | Status |
|---|---|
| `DenseUniTensor` behavior change (intentional) | ✅ Acknowledged |
| Missing `ComplexDouble` dtype test for Block types | ✅ Added — both Block tests now loop over `{Type.ComplexDouble, Type.Double}` |
| Non-diag path returns `this` (not a clone) | ✅ Consistent with `contiguous()`, `to()`, `astype()` — my original review was incorrect on this point |
| Code duplication between Block types | ✅ Tracked separately in #844 |
| `to_dense_()` non-diag no-op | ✅ Intentional and correct |

The codex bot's concern about `to_dense__non_diag` calling `to_dense()` instead of `to_dense_()` is also resolved — the current diff correctly calls `ut_to_diag.to_dense_()`.

No blocking issues remain. The implementation is correct, consistent across all three tensor types, and well-tested. Ready to merge.

pcchen:
## Codecov Failure Analysis

This is a **Codecov project coverage** failure, not a test failure — all tests pass.

### Root Cause: Missing Coverage Upload

The key signal is buried in the report:
> HEAD has 1 upload less than BASE

The base commit (`9f0e8fb`) had more coverage report uploads than this PR's HEAD. One CI job — almost certainly the GPU test suite — didn't upload its coverage data for this run. That alone explains the apparent -8.40% project coverage drop.

### Confirming It's a False Alarm

The dramatic stats reveal the artifact:
- Files: 213 → 241 (+28 files not in this PR)
- Lines: 31,032 → 35,321 (+4,289 lines not in this PR)

These aren't from this PR's changes. The branch is **52 commits behind master**, so Codecov is comparing the PR's partial CI results against a base that includes coverage from all those extra commits' test runs.

### The Actual Patch Coverage

The PR's own new code sits at 74.07% with 7 **partial** lines (not missing lines) across three files — almost certainly branch-coverage partials from `for` loop exit conditions and the `if (_is_diag)` guard's untaken branch in some configurations:

| File | Patch % |
|---|---|
| `BlockFermionicUniTensor.cpp` | 70% (3 partials) |
| `BlockUniTensor.cpp` | 72.72% (3 partials) |
| `DenseUniTensor.cpp` | 83.33% (1 partial) |

### What To Do

1. **Rebase the branch on master** — closing the 52-commit gap will fix the misleading comparison.
2. **Re-run the missing CI job** — so all coverage uploads are present before Codecov checks.

pcchen:
I will approve the PR and Rebase the branch on master.

pcchen (approved):

PR #767, opened 2026-04-03
Fix block Svd_truncate and Gesvd_truncate return_err output

yingjerkao:
Fixes #762.

## Summary
- make block `Svd_truncate` and `Gesvd_truncate` return discarded singular values in the `return_err` path
- make `return_err == 1` return the first discarded singular value instead of the smallest kept value
- keep block output ordering consistent with the dense truncation path
- add regressions for both block APIs and the `return_err == 1` edge case

## Verification
- `./build/openblas-cpu/tests/test_main --gtest_filter="linalg_Test.BkUt_Svd_truncate_return_err_returns_discarded_values:linalg_Test.BkUt_Gesvd_truncate_return_err_returns_discarded_values:linalg_Test.BkUt_Svd_truncate_return_err_one_returns_first_discarded_value:linalg_Test.BkUt_Gesvd_truncate_return_err_one_returns_first_discarded_value"`
- `./build/openblas-cpu/tests/test_main --gtest_filter="Svd_truncate.*:Gesvd_truncate.*:linalg_Test.BkUt_Svd_truncate1:linalg_Test.BkUt_Svd_truncate2:DenseUt_Svd_truncate:DenseUt_Gesvd_truncate"`

PR #808, opened 2026-05-14
Svd_truncate and Gesvd_truncate return the error as expected

### LINKED ISSUE #762: Svd_truncate/Gesvd_truncate return wrong discarded singular values for symmetric UniTensor
A bug spotted by my student Jerry Chen:

## Summary

`cytnx.linalg.Svd_truncate` on symmetric (`UTenType.Block`) `UniTensor` appears to return the wrong value in the `return_err` path.

For block/symmetric tensors, the "discarded singular values" output can overlap with or match the kept singular values. This is inconsistent with the dense tensor path and appears to come from
selecting the kept tail of the sorted singular spectrum instead of the discarded prefix.

The same issue appears in both:

- `linalg::Svd_truncate(const UniTensor&, ...)`
- `linalg::Gesvd_truncate(const UniTensor&, ...)`

## Versions

This was diagnosed looking at current `master`. However the bug was spotted using version 1.0.1. The bug has existed since the `return_err` mechanism was added in commit c8a65dc5.

## Actual behavior

For symmetric/block `UniTensor`:

- `return_err > 1` returns `Sall.get({ac::tilend(smidx)})`, which is the kept tail of the sorted singular values, not the discarded values.
- `return_err == 1` returns `Smin`, which is the smallest kept singular value.

This behavior is incorrect in both paths.

## Expected behavior

The symmetric/block `UniTensor` path should match the dense tensor implementation:

- `return_err > 1` should return the discarded singular values only.
- `return_err == 1` should return the first discarded singular value, not the smallest kept one. This is documented in arxiv:2401.01921 as returning "the largest of the truncated singular values".

## Relevant source locations

Block path in `Svd_truncate`:

- `src/linalg/Svd_truncate.cpp:312-316`

Block path in `Gesvd_truncate`:

- `src/linalg/Gesvd_truncate.cpp:354-358`

Dense/tensor reference behavior:

- `src/backend/linalg_internal_cpu/memcpyTruncation.cpp:197-207`

## Root cause

Both block implementations sort all singular values ascending and compute a truncation threshold `Smin` with index `smidx`, then do:

```cpp
outCyT.push_back(UniTensor(Sall.get({ac::tilend(smidx)})));
```

That slice starts at smidx and runs to the end, which corresponds to the kept values, not the discarded ones.

Similarly, for return_err == 1, the code returns Smin, which is the truncation threshold / smallest kept value, while the dense path returns the first discarded value.

## Additional issue noticed while tracing

There also seems to be a separate mindim inconsistency in the block implementations:

- src/linalg/Svd_truncate.cpp:193-198 ignores mindim in one branch
- src/linalg/Gesvd_truncate.cpp:246-250 ignores mindim in the opposite branch

This looks independent of the discarded-spectrum bug, but likely belongs in the same area.

manuschneider:
fix for Block(Fermionic)UniTensors: same behavior as DenseUniTensor and as documented

fixes #762

### DIFF
--- a/src/linalg/Gesvd_truncate.cpp
+++ b/src/linalg/Gesvd_truncate.cpp
@@ -344,10 +344,15 @@ namespace cytnx {
 
       // handle return_err!
       if (return_err == 1) {
-        outCyT.push_back(UniTensor(Tensor({1}, Smin.dtype())));
-        outCyT.back().get_block_().storage().at(0) = Smin;
+        Tensor out = zeros({1}, Smin.dtype());
+        if (smidx > 0) out(0) = Sall(smidx - 1);
+        outCyT.push_back(UniTensor(out));
       } else if (return_err) {
-        outCyT.push_back(UniTensor(Sall.get({Accessor::tilend(smidx)})));
+        if (smidx > 0) {
+          outCyT.push_back(UniTensor(Sall.get({Accessor::range(0, smidx, 1)})));
+        } else {
+          outCyT.push_back(UniTensor::zeros({1}, {}, Tin.dtype()));
+        }
       }
     }  // _gesvd_truncate_Block_UT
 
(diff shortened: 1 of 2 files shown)


manuschneider:
Will not be developed further, #767 addresses it.

Issue #712, opened 2025-11-24
Installation process changes


manuschneider:
The user guide should be updated to reflect changes in the installation process.

Which parts changed? We want to move from conda to uv, and the installation script that the advanced installation script refer to were removed as well.

Shall we add the installation process on Mac as well? #48 #638 (and not sure if #449 and #663 were resolved or need documentation)

These issues might be related: #149 #473 #661 #668 #654

manuschneider:
The advanced install part is very confusing and it is not clear which installation path we recommend. I would suggest to restructure it:
1) A step-by-step guide from cloning the project to running your first C++/Python program. Uses only our recommended compile path + tools, and the default options
2) Alternative tools (for installing dependencies, compiling, ...)
3) All supported compile options, and how they affect all the above steps.

Currently, the script, which we do not use anymore, is still mentioned; `uv` is not mentioned; the API doc generation is outdated; the user guide generation is not present

PR #818, opened 2026-05-18
Modernize Codecov: dual C++/Python flags, portable coverage, preset-driven CI

IvanaGyro:
## Summary

Rebuilds the Tests-and-Codecov CI so it actually reports both C++ and Python coverage, on top of the PEP 517 / scikit-build-core install path. Composed of five atomic commits:

1. **Make coverage instrumentation portable and target-scoped.** Replaces the global `CMAKE_CXX_FLAGS` mutation with `target_compile_options` / `target_link_options` on the `cytnx` target, wrapped in `$<BUILD_INTERFACE:--coverage>` so downstream `find_package(Cytnx)` consumers aren't dragged into coverage. Branches on `CMAKE_CXX_COMPILER_ID` for GCC / Clang / AppleClang (`--coverage`), MSVC (status message — OpenCppCoverage is the Windows path, no compile flags needed), and unknown compilers (warning, configure still succeeds).
2. **Rewrite codecov.yml for dual C++/Python coverage.** Two upload flags (`cpp` from gcovr over `src/`+`include/`+`pybind/`, `python` from pytest-cov over `cytnx/`) with `paths:` filters and `carryforward: true`. `component_management` splits the C++ side into "C++ backend" and "Python bindings" components for the PR comment. `target: auto` / `threshold: 1%` on project + patch status, down from `100%`. `codecov.notify.after_n_builds: 2` coalesces the two uploads of a single CI run into one status update.
3. **Declare `[project.optional-dependencies]` for test/coverage/dev tooling.** Adds `test` (pytest, pytest-cov), `coverage` (gcovr), and `dev` extras. `pip install --editable .[dev]` is now the single entry point for contributors and CI.
4. **Rebuild Tests-and-Codecov CI on PEP 517 editable install.** Replaces the hand-rolled `cmake configure → build → install → set PYTHONPATH` flow with `pip install --editable '.[dev]' --config-settings=build-dir=build --config-settings=cmake.define.CMAKE_BUILD_TYPE=Debug --config-settings=cmake.define.RUN_TESTS=ON`. Coverage XML uploaded twice (`cpp` and `python` flags) via `codecov-action@v6`; JUnit XMLs shipped to Codecov Test Analytics via `test-results-action@v1`. Order: editable install → import sanity check → explicit `cmake --build . --target test_main` safety net → ctest → pytest with `--cov=cytnx --cov-branch` → gcovr → uploads.
5. **Drop redundant `install(DIRECTORY cytnx)`; pip is the Python install path.** Originally landed as #819, consolidated here. With `wheel.packages = ["cytnx"]` in pyproject.toml, the CMake install of the Python sources duplicates what scikit-build-core already does and prevents editable mode from redirecting `import cytnx` back to the source tree. Removing it lets coverage measurement see actual hits on `cytnx/*.py`. Docs (`adv_install.rst`) get a new "Option C: pip install" section and an explicit note that bare CMake installs no longer ship the Python sources.

## Test plan

- [x] Workflow run on this PR completes; Codecov posts a single PR comment with both `cpp` and `python` flags.
- [x] Python coverage on the comment is **non-zero** (the previous run showed 0% because of the install duplication this branch now fixes).
- [x] PR comment renders three components: *C++ backend*, *Python bindings*, *Python package*.
- [x] `gcovr` report includes files under `pybind/` (no longer excluded).
- [x] Status checks `codecov/project/cpp`, `codecov/project/python`, and `codecov/patch` appear and use the 1% threshold.


CMakeLists.txt
```diff
@@ -292,8 +292,35 @@ if(RUN_TESTS)
...
+  #     Produces .gcno/.gcda files that gcovr and lcov can consume.
+  #
+  #   - MSVC: no native gcov-style instrumentation. On Windows, coverage is
+  #     typically collected with OpenCppCoverage (post-build binary
+  #     instrumentation), which needs no CMake flags. Skip silently except
+  #     for a status message so configure still succeeds.
+  if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR
+     CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    target_compile_options(cytnx PUBLIC --coverage)
+    target_link_options(cytnx PUBLIC --coverage)
```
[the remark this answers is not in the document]
IvanaGyro: (Reply written by Claude on behalf of @IvanaGyro.)

Good catch — applied in 40053f8f. Wrapping the flags in `$<BUILD_INTERFACE:--coverage>` keeps the propagation to `test_main` during this project's build (the gtest binary still gets instrumented when it links `cytnx`) while stripping `--coverage` from the exported `CytnxTargets.cmake`, so external projects that `find_package(Cytnx)` against the installed library aren't dragged into coverage instrumentation.

.github/workflows/ci-cmake_tests.yml
```diff
@@ -10,6 +10,10 @@ jobs:
     defaults:
       run:
         shell: bash -l {0}
+    env:
+      # Pinned via --config-settings=build-dir=build below so ctest and
+      # gcovr know where the build tree (and .gcno / .gcda files) live.
+      BUILD_DIR: ${{ github.workspace }}/build
```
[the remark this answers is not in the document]
IvanaGyro: (Reply written by Claude on behalf of @IvanaGyro.)

The premise is right that `pyproject.toml` pins `--preset=openblas-cpu` and that preset's `binaryDir` is `build/openblas-cpu`, but the install command in this workflow overrides that with `--config-settings=build-dir=build`. scikit-build-core's `build-dir` setting passes `-B build` to cmake, and CMake's `-B` argument takes precedence over the preset's `binaryDir`, so the actual build tree ends up at `${{ github.workspace }}/build/` -- matching the `BUILD_DIR` env var here.

Verified on the run for commit `eb521352`: configure / install logs show `Build files have been written to: /home/runner/work/Cytnx/Cytnx/build/`, ctest reports `100% tests passed, 0 tests failed out of 935`, and the gcovr step uploads `/home/runner/work/Cytnx/Cytnx/build/coverage-cpp.xml`. If `BUILD_DIR` had pointed at the wrong place, ctest would have errored with a missing `CTestTestfile.cmake` -- which it doesn't.

So the override is doing what `BUILD_DIR` assumes. No code change needed; leaving the env var as-is.

IvanaGyro:
The coverage drops with the new settings, which changes the threshold from 100% to 1%, so the GitHub action failed. This is not a regression.

manuschneider:
The advanced install part of the user guide becomes increasingly obscure. I addressed this in #712 as well. It seems the recommended install path is Option C, hidden after A and B.

docs/source/adv_install.rst
```diff
@@ -255,21 +265,51 @@ Additional options for HPTT if -DUSE_HPTT=on:
     $make install
 
 
+**Option C. Install the Python API via pip (recommended for Python users)**
```
manuschneider, line 268: The recommended Option C is compiling everything from scratch (including hptt), even if there were no changes. Is this intended?

IvanaGyro: Yes, if the user has never built with the default preset before or other presets have covered the build directory. BTW, this document needs to be rewritten.

pcchen (commented):
## PR #818 Review

### Overview

This PR rebuilds the Tests-and-Codecov CI workflow from scratch. It replaces a hand-rolled `cmake configure → build → install → PYTHONPATH` flow with a PEP 517 editable install, fixes the C++ coverage flags to be target-scoped rather than globally mutating `CMAKE_CXX_FLAGS`, adds dual C++/Python Codecov reporting with components, and drops the redundant `install(DIRECTORY cytnx)` from CMake that was breaking Python coverage (causing it to report 0%). The five commits are logically ordered and each does one coherent thing.

---

### Strengths

- **Root cause fix is correct**: The `install(DIRECTORY cytnx)` removal is the right solution to the 0% Python coverage bug — the installed copy was shadowing the editable source tree, so pytest-cov was measuring the wrong files.
- **`$<BUILD_INTERFACE:--coverage>` in CMakeLists.txt** is the correct idiom: coverage flags propagate in-tree (to `test_main` via `PUBLIC`) but are stripped from exported `CytnxTargets.cmake` so downstream `find_package(Cytnx)` consumers are not dragged in.
- **`CMAKE_CXX_COMPILER_ID MATCHES "Clang"`** correctly catches `Clang`, `AppleClang`, and `IntelLLVM` via regex, avoiding the need for an explicit `AppleClang` branch.
- **`carryforward: true`** in `codecov.yml` is essential: without it, a Python-only PR would silently drop C++ coverage to 0%.
- **`after_n_builds: 2`** correctly coalesces the two flag uploads into one Codecov notification, avoiding confusing double-updates for reviewers.
- **`if: ${{ !cancelled() }}`** on the test-results upload ensures JUnit XML reaches Codecov even when tests fail, so Test Analytics shows the failing tests.
- **`--cov=cytnx` fix**: the old code had `--cov=com`, which is clearly wrong. The new `--cov=cytnx` is correct.
- **`ignore` glob patterns** updated from bare directory names (`"tests"`) to proper patterns (`"tests/**"`) — the old patterns likely did not work as intended.

---

### Issues

**Medium priority:**

1. **`codecov/codecov-action@v6` is not pinned to a commit SHA** (all three upload steps in the workflow). The project correctly pins `peaceiris/actions-gh-pages` to a commit hash in PR #839; the same practice should apply here. A tag like `@v6` can be silently force-pushed by the action maintainer. Replace with the current SHA for `v6`:
   ```yaml
   uses: codecov/codecov-action@13afa45c3c9ec450f4e85a20c5c36e4b9f27b6d3  # v6
   ```

2. **`after_n_builds: 2` has no timeout behaviour on partial upload failure**. If the C++ coverage upload step succeeds but the Python step fails (network error, token issue, etc.), Codecov waits indefinitely for the second build and never posts a PR comment or status. This is a known Codecov limitation and not directly fixable in `codecov.yml`, but worth documenting in a comment — or accepting that `fail_ci_if_error: true` on each upload step will fail the CI run and leave the PR visibly broken.

**Low priority:**

3. **`--doctest-modules` in the pytest command** is risky in CI. If any `cytnx/*.py` module has a doctest that imports an optional dependency or has side effects, the entire pytest run fails. Consider whether all existing module-level doctests are reliable enough for CI, or scope it to `pytests/` only (drop `--doctest-modules` or add `--ignore=cytnx/`).

4. **`dev` extra duplicates `test` + `coverage` instead of composing them** (`pyproject.toml`). The more maintainable pattern avoids the duplication:
   ```toml
   dev = ["cytnx[test,coverage]"]
   ```
   As written, adding a new package to `test` requires updating `dev` separately or it silently falls out of sync.

5. **gcovr `--exclude` regexes and `codecov.yml` `ignore` globs overlap but are not identical**. gcovr uses `'.*_deps.*'` (regex, matches `_deps` anywhere in path) while codecov.yml uses `"**/_deps/**"` (glob, matches only `_deps` as a path component). In practice they cover the same files, but divergence between the two lists can become confusing when adding new excluded directories. A comment in the workflow step pointing to the codecov.yml ignore list would help.

---

### Minor Nits

- `ci-cmake_tests.yml`: the inline comments explaining each step are thorough and useful for a CI file that future contributors will need to debug.
- `adv_install.rst` Option C: the instruction "Re-run `pip install --editable .` from the repository root to make the Python API importable; the already-compiled artifacts will be reused" is slightly inaccurate if the user did a bare CMake install to a different prefix — the `.so` from that cmake build won't automatically be reused by pip's editable install. A minor documentation edge case.
- The `BUILD_DIR` env var is defined as an absolute path (`${{ github.workspace }}/build`) while `--config-settings=build-dir=build` passes a relative path. Both resolve to the same directory, but the inconsistency is slightly confusing on first read.

---

### Summary

The design is sound and addresses the real root causes (0% Python coverage, global CMake flags, broken install path). The primary actionable item before merge is pinning `codecov/codecov-action` to a commit SHA (item 1) to match the project's existing supply-chain practice. Item 4 (`dev` extra duplication) is a minor style issue. Everything else is low-priority.

docs/source/adv_install.rst
pcchen: We will re-write the document after we fix the docs deployment.

pcchen (approved):

IvanaGyro:
@pcchen Setting `informational: false` is intentional. It force the PR in the future to add tests while adding new features.

PR #828, opened 2026-05-25
Consolidate documentation dependencies into root pyproject.toml

### LINKED ISSUE #827: docs: merge docs/pyproject.toml into root pyproject.toml
## Background

The documentation toolchain currently lives in `docs/pyproject.toml`, a completely separate project managed by `uv`. This is unusual: NumPy, SciPy, PyTorch, JAX, and QuTiP all keep doc dependencies in their root `pyproject.toml`, either as an optional dependency group or in a plain `docs/requirements.txt`. A separate `pyproject.toml` means two separate virtual environments to maintain, two lock files, and no shared tooling between library and doc builds.

## Proposed change

Move the doc dependencies from `docs/pyproject.toml` into the root `pyproject.toml` as an optional group:

```toml
[project.optional-dependencies]
docs = [
    "sphinx",
    "sphinxcontrib-bibtex>=2.7.0",
    "sphinxcontrib-jquery>=4.1",
    "sphinxbootstrap4theme>=0.6.0",
    "sphinx-multiversion>=0.2.4",
    "furo",
]
```

Delete `docs/pyproject.toml`, `docs/uv.lock`, and `docs/src/` after the migration.

## Building docs with the new layout

After the migration, install and build with either of the following:

**With uv (recommended):**
```bash
uv sync --extra docs
uv run sphinx-build -M html docs/source docs/build
```

**With pip:**
```bash
pip install -e ".[docs]"
sphinx-build -M html docs/source docs/build
```

The CI workflow would change from `working-directory: docs / uv sync` to a single `uv sync --extra docs` at the repo root, and the `sphinx-build` invocation would add `docs/` prefixes to the source and build paths.

## Notes

- `doxygen` and `graphviz` are installed via `apt-get` in CI and are not Python packages; they stay as-is.
- The `docs/api_doc_script.sh` script and `docs/api_docs/` tree are unaffected by this change.
- This requires checking that the root `requires-python` is broad enough to cover all doc dependencies (currently `>=3.10` on the doc side).

IvanaGyro:
## Summary

Move the Sphinx documentation toolchain from a standalone
`docs/pyproject.toml` (managed as its own `uv` project) into the root
`pyproject.toml` as a `docs` optional-dependencies group. This brings the
layout in line with NumPy, SciPy, JAX, and QuTiP, which all ship docs
deps from the root manifest, and eliminates the second virtualenv and
lock file the docs job had to maintain. Closes #827.

Alongside the dependency consolidation, this PR makes two small workflow
adjustments that keep `.github/workflows/docs.yml` in a safe state while
the publish pipeline is refactored separately:

- Run the docs build on every PR and on manual dispatch, not only on
  pushes to `master`, so changes to the docs pipeline can be exercised
  before they reach the published site.
- Gate `docs/api_doc_script.sh`, the follow-up move step, and the
  gh-pages Deploy step with `if: false` until a follow-up PR can
  refactor them. The script's `git checkout` restore logic does not
  cope with the detached-HEAD state of `pull_request` runs, and
  publishing in the current shape would force-replace `Cytnx_doc` with
  a build that is missing the API reference.

## Changes

- **`pyproject.toml`** — add `[project.optional-dependencies] docs = [
  sphinx, sphinxcontrib-bibtex, sphinxcontrib-jquery,
  sphinxbootstrap4theme, sphinx-multiversion, furo ]`. Lower bounds are
  chosen so the resolver can satisfy the existing root
  `requires-python = ">=3.9"`: `sphinx>=7.4.7` (last 7.x release with
  Python 3.9 support — 8.x requires `>=3.10`) and
  `sphinxcontrib-bibtex>=2.6.2` (2.7.0 raises its floor to `>=3.10`).
  On Python 3.10+, `uv` resolves to newer versions (sphinx 8.x,
  bibtex 2.7.x).
- **Delete `docs/pyproject.toml` and `docs/uv.lock`** — superseded by the
  root manifest. Doc deps are installed via `uv sync --extra docs` or
  `pip install -e ".[docs]"`.
- **`.github/workflows/docs.yml`** — drop `working-directory: docs` from
  the dep-install and Sphinx-build steps and run them from the repo
  root. Use `uv sync --python 3.10 --extra docs --no-install-project` so
  the cytnx C++ extension is not built for the docs job (it isn't needed
  to render the user guide and would pull in CMake/BLAS/etc.). The
  api-docs and Sphinx steps run under `uv run --no-sync`, which reuses
  the already-prepared virtualenv without retriggering an install and
  puts `sphinx-build` on `PATH` for both the direct invocation and the
  `make` call inside `api_doc_script.sh`. The setup-uv default
  cache-dependency-glob already covers `pyproject.toml`, so no override
  is set.

  Add `pull_request:` and `workflow_dispatch:` to the `on:` triggers so
  PRs and manual runs exercise the Sphinx build (the existing `push` to
  `master` trigger is kept). Gate the `Build api documentation`,
  `Move api_build into html output`, and `Deploy` steps with `if: false`
  plus TODO comments describing the follow-up refactor — re-enable
  `api_doc_script.sh` in the new uv-based layout, switch the publish to
  per-version `destination_dir: <version>/` with a matching `stable/`
  alias, send `master` pushes into `dev/`, drop `force_orphan: true` so
  the gh-pages branch can accumulate versions, and serve a
  `switcher.json` from the docs site root so the theme's version
  dropdown lists every published release.
- **`docs/README.md`** — rewrite build instructions for the new
  `--extra docs` layout with both `uv` and `pip` variants, with
  `docs/source` and `docs/build` paths since commands now run from the
  repo root. Add an optimization note covering `--no-install-project`
  for contributors who only render the user guide.

## Test plan

- [x] `uv sync --python 3.10 --extra docs --no-install-project` resolves
      cleanly on Python 3.10; spot-checked Python 3.9 resolves too.
- [x] `uv run --no-sync sphinx-build -M html docs/source docs/build`
      succeeds. Build finishes with 25 warnings, all pre-existing
      content issues in `.rst` sources (duplicate `autosectionlabel`
      labels, one missing include, two pages outside the toctree) —
      none introduced by this change.
- [x] Walked 5419 internal `href`/`src` references in the generated
      HTML; only `api_build/...` targets (produced by
      `docs/api_doc_script.sh` in CI, not in a local sphinx-only build)
      and `py-modindex.html` (which Sphinx omits without `:py:module:`)
      are absent. No sphinx-rendered internal link broken.


.github/workflows/docs.yml
```diff
@@ -19,20 +19,20 @@ jobs:
         with:
           version: "0.8.17"
           enable-cache: true
-          cache-dependency-glob: "docs/uv.lock"
+          cache-dependency-glob: "pyproject.toml"
```
IvanaGyro: Just remove cache-dependency-glob, the default values cover pyproject.toml

.github/workflows/docs.yml
```diff
@@ -19,20 +19,20 @@ jobs:
...
-          uv sync --python 3.10
+          uv sync --python 3.10 --extra docs --no-install-project
           sudo apt-get update && sudo apt-get install -y doxygen graphviz
       - name: Build api documentation
         working-directory: docs
+        env:
+          PATH: ${{ github.workspace }}/.venv/bin:${{ env.PATH }}
         run: |
-          uv run bash api_doc_script.sh
+          bash api_doc_script.sh
```
IvanaGyro: also consider uv run --no-sync

docs/README.md
```diff
@@ -3,23 +3,45 @@ The documentation of the Cytnx tensor network library
...
 # Build locally for preview
-Build the HTML user guide into `build/html/`:
+Build the HTML user guide into `docs/build/html/`. Run these commands from the
+repository root.
+
+With `uv` (recommended):
+
+```bash
+uv sync --extra docs
+uv run sphinx-build -M html docs/source docs/build
```
IvanaGyro: also uv run --no-sync

.github/workflows/docs.yml
```diff
@@ -19,20 +19,20 @@ jobs:
...
       - name: Sphinx build
-        working-directory: docs
         run: |
-          uv run sphinx-build -M html source build
+          .venv/bin/sphinx-build -M html docs/source docs/build
```
IvanaGyro: consider uv run --no-sync for all the commands you added/modified

docs/README.md
```diff
@@ -3,23 +3,45 @@ The documentation of the Cytnx tensor network library
...
+
+The commands above also build the cytnx C++ extension, which is not actually
+required to render the user guide. To install only the doc dependencies and
+skip the C++ build, pass `--no-install-project` to `uv sync` (or set up a
+separate `pip` virtual environment) and invoke `sphinx-build` from the
+resulting environment directly:
+
+```bash
+uv sync --extra docs --no-install-project
+.venv/bin/sphinx-build -M html docs/source docs/build
```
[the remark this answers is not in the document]
IvanaGyro: accept

manuschneider (commented):
Suggestions with AI help:

1) The changed workflow is never exercised by this PR's CI. `docs.yml` triggers only on: push: branches: [master]. The docs pipeline runs only after merge, so a failure to build the docs would first occur during merge, breaking the published Cytnx_doc site. Validation rests entirely on the author's local run. I'd suggest a one-off workflow_dispatch (or a temporary pull_request trigger) to prove the pipeline before merging.

2) The 3.9 / Sphinx-7.x path is resolved but never built. Both CI (--python 3.10) and the author build on 3.10 → Sphinx 8.x. The ≥3.9 lower-bound combo (Sphinx 7.4.7) was only checked for resolution, not an actual render. Low risk, but a local pip user on 3.9 is on an untested rendering path.

3) Minor, pre-existing: `api_doc_script.sh` does `mkdir build` (no `-p`) — fails if `docs/build` already exists. Unchanged by this PR and fine on a fresh CI checkout; just noting it's load-bearing on cleanliness.

.github/workflows/docs.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - master
+  pull_request:
```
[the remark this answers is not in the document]
IvanaGyro: Disable running docs/api_doc_script.sh in this PR. We will refactor the whole deploy workflow in the other PR.

.github/workflows/docs.yml
```diff
@@ -19,25 +21,27 @@ jobs:
...
-          cache-dependency-glob: "docs/uv.lock"
       - name: Install dependencies
-        working-directory: docs
         run: |
-          uv sync --python 3.10
+          uv sync --python 3.10 --extra docs --no-install-project
           sudo apt-get update && sudo apt-get install -y doxygen graphviz
+      # TODO: re-enable once api_doc_script.sh is fixed in a follow-up commit.
       - name: Build api documentation
+        if: false
```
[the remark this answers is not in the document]
IvanaGyro: Disable deployment in this PR. We will renew the whole deploy workflow in the other PR because the change is so big.

IvanaGyro:
> Suggestions with AI help:
> 
>     1. The changed workflow is never exercised by this PR's CI. `docs.yml` triggers only on: push: branches: [master]. The docs pipeline runs only after merge, so a failure to build the docs would first occur during merge, breaking the published Cytnx_doc site. Validation rests entirely on the author's local run. I'd suggest a one-off workflow_dispatch (or a temporary pull_request trigger) to prove the pipeline before merging.
> 
>     2. The 3.9 / Sphinx-7.x path is resolved but never built. Both CI (--python 3.10) and the author build on 3.10 → Sphinx 8.x. The ≥3.9 lower-bound combo (Sphinx 7.4.7) was only checked for resolution, not an actual render. Low risk, but a local pip user on 3.9 is on an untested rendering path.
> 
>     3. Minor, pre-existing: `api_doc_script.sh` does `mkdir build` (no `-p`) — fails if `docs/build` already exists. Unchanged by this PR and fine on a fresh CI checkout; just noting it's load-bearing on cleanliness.

 1. added
 2. Python 3.9 is end of life, I will bump the minimal support Python version to 3.10 in the future.
 3. disable running the script and deployment. The whole workflow need to be refactor to handle old versions easier. This is a big change so I will make change in the future PR.

manuschneider (approved):

PR #837, opened 2026-05-27
feat(symmetry): enforce Zn input-validation semantics (#771)

IvanaGyro:
Closes #771. Follow-up refactor tracked in #840.

## Summary

Pick a single contract for the Zn arithmetic API: valid quantum numbers are
canonical representatives in `[0, n)`, matching what `check_qnum` /
`check_qnums` already define and what `Bond` construction enforces.

The split policy used here is:

1. **Python binding** — keep accepting out-of-range Zn inputs, but emit a
   `FutureWarning` and normalize into `[0, n)` before forwarding to C++.
   The warning announces removal of the lenient behaviour in v2.0.0.
2. **C++ API** — reject out-of-range Zn inputs. `ZnSymmetry::combine_rule_`
   (scalar and vector overloads) and `reverse_rule_` validate inputs
   against `[0, n)` and throw `std::logic_error` via `cytnx_error_msg`
   on violation.

Tightening the C++ rule surfaced a pre-existing typo in
`UniTensor::_fx_get_total_fluxs`, fixed in the same PR so bisect does
not hit a regression window. See "Latent bug fixed in passing" below.

## Commits

1. `feat(symmetry): warn and normalize out-of-range Zn inputs in Python`
2. `fix(unitensor): use per-component symmetry in _fx_get_total_fluxs`
3. `feat(symmetry): reject out-of-range Zn qnums in C++ arithmetic API`

Ordering matters: (2) must land before (3) so `_fx_get_total_fluxs` is
already correct when the stricter Zn validation goes in.

### `src/Symmetry.cpp`
- Replace the defensive `NormalizeZnQnum` helper (introduced in #768) with
  `ValidateZnQnum` / `ValidateZnQnums` validation helpers.
- `ZnSymmetry::combine_rule_` (vector and scalar overloads) and
  `reverse_rule_` validate inputs up front; the bodies simplify to
  `(inL + inR) % n` for combine and `(n - in) % n` for reverse, since
  inputs are now constrained to `[0, n)`. The vector overload uses a
  straightforward double loop instead of a single flat index with
  integer division and modulo.

### `pybind/symmetry_py.cpp`
- Add a `NormalizeZnInput` helper in an anonymous namespace that returns
  the input unchanged for non-Zn symmetries, and for Zn:
  - checks whether the qnum is in `[0, n)`,
  - if not, emits a `FutureWarning` via `PyErr_WarnEx(PyExc_FutureWarning,
    ..., 2)` matching the existing `get_blocks_(slient=...)` pattern in
    `pybind/unitensor_py.cpp`; the message is assembled with
    `std::format` to match other diagnostics in the codebase,
  - returns the positive-modulo normalization of the input.
- Wrap the bound `combine_rule(qnL, qnR, is_reverse)` and
  `reverse_rule(qin)` lambdas to call `NormalizeZnInput` on each
  argument before forwarding to the C++ rule. The lambdas take
  `const Symmetry &self` to match the const-ness of the underlying
  methods.

### `include/UniTensor.hpp` — latent bug fixed in passing
- `_fx_get_total_fluxs` (and its duplicate copy on
  `BlockFermionicUniTensor`) seeded `total_qns[i]` on the BRA-initial
  branch with `syms[0].reverse_rule(...)`. Every other arithmetic step
  in the same function uses `syms[i]`, so this was a typo specific to
  that branch. On uniform symmetry lists (`{U1, U1}`, `{Zn, Zn}`) the
  bug is invisible because `syms[0]` and `syms[i]` are the same type;
  on mixed lists with a BRA first bond (e.g. `{U1, Zn(N)}`) it applied
  `U1::reverse_rule` to a Zn qnum and produced an out-of-range result
  that the rest of the loop then fed back into `Zn::combine_rule`.
  Silent before strict Zn validation, a hard `std::logic_error` after.
  Replace `syms[0]` with `syms[i]` in both copies.

### `tests/Bond_test.cpp`
- Remove `Bond.ZnSymmetryNormalizesPublicRules`, which pinned the old
  lenient behaviour from #768.
- Add `Bond.ZnSymmetryPublicRulesProduceCanonicalOutputs` covering Z2
  and Z3 canonical inputs for `combine_rule`,
  `combine_rule(..., /*is_reverse=*/true)`, and `reverse_rule`.
- Add `Bond.ZnSymmetryRejectsOutOfRangeInputs` covering negative and
  `>= n` inputs on every public arithmetic overload, including the
  vector form and the `is_reverse` branch.

### `tests/BlockUniTensor_test.cpp`
- Add `BlockUniTensorTest.FxGetTotalFluxsUsesPerComponentSymmetry`,
  which constructs `UniTensor({BD_BRA{U1, Zn(2)}, BD_KET{U1, Zn(2)}})`.
  Without the `_fx_get_total_fluxs` fix this constructor would throw
  `std::logic_error` from the strict Zn validation; with the fix it
  completes and reports the expected symmetry list.

### `pytests/symmetry_test.py` (new)
Six tests covering:
- canonical Zn inputs do not emit warnings,
- negative inputs to `combine_rule` warn with the documented message
  and return the canonical combined value,
- `>= n` inputs warn and normalize,
- the `is_reverse` branch warns and normalizes (exercised on Z3 so the
  reverse step is a non-trivial permutation),
- `reverse_rule` warns + normalizes,
- U1 still accepts arbitrary integers without warnings.

## Internal call sites

`src/Bond.cpp` and `src/SparseUniTensor.cpp` are the C++ internal callers
of `combine_rule_` / `reverse_rule_`. They feed qnums that have already
been validated by `Bond` construction (which calls `check_qnum` on every
entry), so they remain compatible with the stricter rule.

## Test plan

Run on both `mkl-cpu` and `openblas-cpu` presets:

```sh
cmake -S . -B build/mkl-cpu -DRUN_TESTS=ON -DUSE_MKL=ON
cmake --build build/mkl-cpu -- -j$(nproc)
ctest --test-dir build/mkl-cpu

cmake -S . -B build/openblas-cpu -DRUN_TESTS=ON
cmake --build build/openblas-cpu -- -j$(nproc)
ctest --test-dir build/openblas-cpu
```

Python tests:

```sh
pytest pytests/symmetry_test.py -v
```

### Results

- `pre-commit run` — clean.
- `ctest --test-dir build/openblas-cpu` — **100% tests passed**, 938 /
  938 (incl. the two new `Bond.ZnSymmetry*` tests and the new
  `BlockUniTensorTest.FxGetTotalFluxsUsesPerComponentSymmetry`
  regression test).
- `ctest --test-dir build/mkl-cpu` — **100% tests passed**, 938 / 938.
- `pytest pytests/symmetry_test.py` — **6 / 6 passed**.

### DIFF
--- a/pybind/symmetry_py.cpp
+++ b/pybind/symmetry_py.cpp
@@ -1,3 +1,4 @@
+#include <format>
 #include <vector>
 #include <map>
 #include <random>
@@ -16,6 +17,24 @@
 namespace py = pybind11;
 using namespace pybind11::literals;
 using namespace cytnx;
+
+namespace {
+  cytnx_int64 NormalizeZnInput(const Symmetry &sym, const cytnx_int64 qnum, const char *fn_name,
+                               const char *arg_name) {
+    if (sym.stype() != SymmetryType::Z) return qnum;
+    const cytnx_int64 n = sym.n();
+    if (qnum >= 0 && qnum < n) return qnum;
+    cytnx_int64 normalized = qnum % n;
+    if (normalized < 0) normalized += n;
+    const std::string message = std::format(
+      "Passing out-of-range Z{} qnum {} to '{}' (argument '{}') is deprecated and will be "
+      "rejected in v2.0.0; pass a canonical representative in [0, {}) instead. Normalizing to "
+      "{} for now.",
+      n, qnum, fn_name, arg_name, n, normalized);
+    if (PyErr_WarnEx(PyExc_FutureWarning, message.c_str(), 2) < 0) throw py::error_already_set();
+    return normalized;
+  }
+}  // namespace
 
 void symmetry_binding(py::module &m) {
   py::enum_<SymmetryType>(m, "SymType")
@@ -55,12 +74,19 @@ void symmetry_binding(py::module &m) {
     .def("check_qnums", &Symmetry::check_qnums, py::arg("qnums"))
     .def(
       "combine_rule",
-      [](Symmetry &self, const cytnx_int64 &inL, const cytnx_int64 &inR, const bool &is_reverse) {
-        return self.combine_rule(inL, inR, is_reverse);
+      [](const Symmetry &self, const cytnx_int64 &inL, const cytnx_int64 &inR,
+         const bool &is_reverse) {
+        const cytnx_int64 normL = NormalizeZnInput(self, inL, "combine_rule", "qnL");
+        const cytnx_int64 normR = NormalizeZnInput(self, inR, "combine_rule", "qnR");
+        return self.combine_rule(normL, normR, is_reverse);
       },
       py::arg("qnL"), py::arg("qnR"), py::arg("is_reverse") = false)
     .def(
-      "reverse_rule", [](Symmetry &self, const cytnx_int64 &qin) { return self.reverse_rule(qin); },
+      "reverse_rule",
+      [](const Symmetry &self, const cytnx_int64 &qin) {
+        const cytnx_int64 norm = NormalizeZnInput(self, qin, "reverse_rule", "qin");
+        return self.reverse_rule(norm);
+      },
       py::arg("qin"))
     .def("get_fermion_parity", &Symmetry::get_fermion_parity, py::arg("qnum"))
     .def("is_fermionic", &Symmetry::is_fermionic)
--- a/pytests/symmetry_test.py
+++ b/pytests/symmetry_test.py
@@ -1,0 +1,85 @@
+import warnings
+
+import pytest
+
+import cytnx
+
+
+def test_zn_combine_rule_canonical_inputs_no_warning():
+    """Canonical qnums in [0, n) must not trigger any deprecation warning."""
+    z3 = cytnx.Symmetry.Zn(3)
+    with warnings.catch_warnings():
+        warnings.simplefilter("error", FutureWarning)
+        assert z3.combine_rule(1, 2) == 0
+        assert z3.combine_rule(2, 2) == 1
+        assert z3.combine_rule(0, 0) == 0
+        assert z3.combine_rule(1, 1, is_reverse=True) == 1
+        assert z3.reverse_rule(0) == 0
+        assert z3.reverse_rule(1) == 2
+        assert z3.reverse_rule(2) == 1
+
+
+def test_zn_combine_rule_negative_input_warns_and_normalizes():
+    """Out-of-range Zn inputs to combine_rule emit FutureWarning and are normalized."""
+    z2 = cytnx.Symmetry.Zn(2)
+    with pytest.warns(
+            FutureWarning,
+            match=
+            "Passing out-of-range Z2 qnum -1 to 'combine_rule' \\(argument 'qnL'\\) is deprecated"):
+        assert z2.combine_rule(-1, 0) == 1
+
+    with pytest.warns(
+            FutureWarning,
+            match=
+            "Passing out-of-range Z2 qnum -1 to 'combine_rule' \\(argument 'qnR'\\) is deprecated"):
+        assert z2.combine_rule(0, -1) == 1
+
+
+def test_zn_combine_rule_above_range_input_warns_and_normalizes():
+    """qnum >= n is normalized via modulo."""
+    z3 = cytnx.Symmetry.Zn(3)
+    with pytest.warns(FutureWarning,
+                      match="Passing out-of-range Z3 qnum 5"):
+        assert z3.combine_rule(5, 0) == 2
+
+    with pytest.warns(FutureWarning,
+                      match="Passing out-of-range Z3 qnum 4"):
+        assert z3.combine_rule(1, 4) == 2
+
+
+def test_zn_combine_rule_with_is_reverse_warns_and_normalizes():
+    """is_reverse path also warns + normalizes.
+
+    Uses Z3 (N > 2) so the reverse step is non-trivial: in Z2 every qnum
+    is its own inverse, which would let a bug in the reverse path pass
+    silently.
+    """
+    z3 = cytnx.Symmetry.Zn(3)
+    with pytest.warns(FutureWarning, match="out-of-range Z3 qnum -1"):
+        # combine(-1, 0) normalizes to combine(2, 0) = 2; reverse(2) in Z3 = 1.
+        assert z3.combine_rule(-1, 0, is_reverse=True) == 1
+
+
+def test_zn_reverse_rule_out_of_range_warns_and_normalizes():
+    """reverse_rule warns + normalizes for out-of-range inputs."""
+    z3 = cytnx.Symmetry.Zn(3)
+    with pytest.warns(
+            FutureWarning,
+            match=
+            "Passing out-of-range Z3 qnum -1 to 'reverse_rule' \\(argument 'qin'\\) is deprecated"):
+        assert z3.reverse_rule(-1) == 1
+
+    with pytest.warns(FutureWarning,
+                      match="Passing out-of-range Z3 qnum 4"):
+        assert z3.reverse_rule(4) == 2
+
+
+def test_u1_does_not_normalize_or_warn():
+    """U1 accepts arbitrary integers; no warning, no modulo."""
+    u1 = cytnx.Symmetry.U1()
+    with warnings.catch_warnings():
+        warnings.simplefilter("error", FutureWarning)
+        assert u1.combine_rule(-3, 5) == 2
+        assert u1.combine_rule(7, -10) == -3
+        assert u1.reverse_rule(-4) == 4
+        assert u1.reverse_rule(9) == -9
--- a/src/Symmetry.cpp
+++ b/src/Symmetry.cpp
@@ -11,10 +11,15 @@ using namespace std;
 namespace cytnx {
 
   namespace {
-    cytnx_int64 NormalizeZnQnum(cytnx_int64 qnum, const cytnx_int64 n) {
-      qnum %= n;
-      if (qnum < 0) qnum += n;
-      return qnum;
+    void ValidateZnQnum(const cytnx_int64 qnum, const cytnx_int64 n) {
+      cytnx_error_msg(
+        (qnum < 0) || (qnum >= n),
+        "[ERROR][ZnSymmetry] qnum %lld is out of the valid range [0, %lld) for Z%lld.\n",
+        static_cast<long long>(qnum), static_cast<long long>(n), static_cast<long long>(n));
+    }
+
+    void ValidateZnQnums(const std::vector<cytnx_int64> &qnums, const cytnx_int64 n) {
+      for (const auto &q : qnums) ValidateZnQnum(q, n);
     }
   }  // namespace
 
@@ -123,21 +128,28 @@ namespace cytnx {
   void cytnx::ZnSymmetry::combine_rule_(std::vector<cytnx_int64> &out,
                                         const std::vector<cytnx_int64> &inL,
                                         const std::vector<cytnx_int64> &inR) {
+    ValidateZnQnums(inL, this->n);
+    ValidateZnQnums(inR, this->n);
     out.resize(inL.size() * inR.size());
-    for (cytnx_uint64 i = 0; i < out.size(); i++) {
-      out[i] = NormalizeZnQnum(inL[cytnx_uint64(i / inR.size())] + inR[i % inR.size()], this->n);
+    for (cytnx_uint64 i = 0; i < inL.size(); i++) {
+      for (cytnx_uint64 j = 0; j < inR.size(); j++) {
+        out[i * inR.size() + j] = (inL[i] + inR[j]) % this->n;
+      }
     }
   }
   void cytnx::ZnSymmetry::combine_rule_(cytnx_int64 &out, const cytnx_int64 &inL,
                                         const cytnx_int64 &inR, const bool &is_reverse) {
-    const cytnx_int64 combined = NormalizeZnQnum(inL + inR, this->n);
+    ValidateZnQnum(inL, this->n);
+    ValidateZnQnum(inR, this->n);
+    const cytnx_int64 combined = (inL + inR) % this->n;
     if (is_reverse)
       this->reverse_rule_(out, combined);
     else
       out = combined;
   }
   void cytnx::ZnSymmetry::reverse_rule_(cytnx_int64 &out, const cytnx_int64 &in) {
-    out = NormalizeZnQnum(-in, this->n);
+    ValidateZnQnum(in, this->n);
+    out = (this->n - in) % this->n;
   }
 
   void cytnx::ZnSymmetry::print_info() const {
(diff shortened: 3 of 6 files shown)


pybind/symmetry_py.cpp
```diff
@@ -17,6 +17,24 @@ namespace py = pybind11;
...
+    if (sym.stype() != SymmetryType::Z) return qnum;
+    const cytnx_int64 n = sym.n();
+    if (qnum >= 0 && qnum < n) return qnum;
+    cytnx_int64 normalized = qnum % n;
+    if (normalized < 0) normalized += n;
+    const std::string message =
+      "Passing out-of-range Z" + std::to_string(n) + " qnum " + std::to_string(qnum) + " to '" +
+      fn_name + "' (argument '" + arg_name + "') is deprecated and will be rejected in v2.0.0; " +
+      "pass a canonical representative in [0, " + std::to_string(n) + ") instead. " +
+      "Normalizing to " + std::to_string(normalized) + " for now.";
```
IvanaGyro: Can we use fmt module here?

pytests/symmetry_test.py
```diff
@@ -0,0 +1,79 @@
...
+        assert z3.combine_rule(5, 0) == 2
+
+    with pytest.warns(FutureWarning,
+                      match="Passing out-of-range Z3 qnum 4"):
+        assert z3.combine_rule(1, 4) == 2
+
+
+def test_zn_combine_rule_with_is_reverse_warns_and_normalizes():
+    """is_reverse path also warns + normalizes."""
+    z2 = cytnx.Symmetry.Zn(2)
```
IvanaGyro: For reverse case, suggest use N > 2

src/Symmetry.cpp
```diff
@@ -123,21 +128,26 @@ namespace cytnx {
   void cytnx::ZnSymmetry::combine_rule_(std::vector<cytnx_int64> &out,
                                         const std::vector<cytnx_int64> &inL,
                                         const std::vector<cytnx_int64> &inR) {
+    ValidateZnQnums(inL, this->n);
+    ValidateZnQnums(inR, this->n);
     out.resize(inL.size() * inR.size());
     for (cytnx_uint64 i = 0; i < out.size(); i++) {
```
IvanaGyro: Use two-level for loop instead.

src/Symmetry.cpp
```diff
@@ -123,21 +128,26 @@ namespace cytnx {
...
+    const cytnx_int64 combined = (inL + inR) % this->n;
     if (is_reverse)
       this->reverse_rule_(out, combined);
     else
       out = combined;
   }
   void cytnx::ZnSymmetry::reverse_rule_(cytnx_int64 &out, const cytnx_int64 &in) {
-    out = NormalizeZnQnum(-in, this->n);
+    ValidateZnQnum(in, this->n);
+    out = (this->n - in) % this->n;
```
IvanaGyro, line 152: Is just `this->n - in` enough?

tests/Bond_test.cpp
```diff
@@ -165,17 +165,41 @@ TEST(Bond, CombindBondSymm_v2) {
...
   Symmetry z2 = Symmetry::Zn(2);
 
-  EXPECT_EQ(z2.combine_rule(-1, 0), 1);
-  EXPECT_EQ(z2.combine_rule(0, -1), 1);
-  EXPECT_EQ(z2.combine_rule(-1, 0, true), 1);
+  EXPECT_EQ(z2.combine_rule(0, 0), 0);
+  EXPECT_EQ(z2.combine_rule(0, 1), 1);
+  EXPECT_EQ(z2.combine_rule(1, 0), 1);
+  EXPECT_EQ(z2.combine_rule(1, 1), 0);
+  EXPECT_EQ(z2.combine_rule(1, 1, true), 0);
```
IvanaGyro: Add `/**/` for tell what true means.

src/Symmetry.cpp
```diff
@@ -123,21 +128,26 @@ namespace cytnx {
   void cytnx::ZnSymmetry::combine_rule_(std::vector<cytnx_int64> &out,
```
IvanaGyro, line 128: Investigate if we can just return `out` instead of having it as arg. If yes, do refactor with a new commit.

src/Symmetry.cpp
IvanaGyro: No — for `in == 0`, `this->n - in` is `n`, which is one past the canonical range `[0, n)`. The reverse of `0` should be `0` (since `0 + 0 = 0` is the additive identity in Zn), not `n`. The modulo on the outside maps the boundary case `n → 0` while leaving every other value (`n - in` for `in ∈ [1, n-1]`, which is in `[1, n-1]`) unchanged. Keeping it as `(this->n - in) % this->n`.

---

src/Symmetry.cpp
IvanaGyro: Investigated. The `_`-suffixed overload returns by output reference because internal callers in `src/Bond.cpp` (lines 105–272) use the scalar `combine_rule_(cytnx_int64 &out, ...)` form to write directly into pre-allocated elements of a `std::vector<cytnx_int64>` (e.g. `new_qnums[d][i]`, `tmpqnum[i]`). Converting them to return-by-value would force a copy at every call site.

For consistency, the vector overload `combine_rule_(std::vector<cytnx_int64> &out, ...)` follows the same pattern, and it's a `virtual` method on `Symmetry_base` overridden by all four symmetry classes (`U1Symmetry`, `ZnSymmetry`, `FermionParitySymmetry`, `FermionNumberSymmetry`), plus a public passthrough on `Symmetry`. Changing it to return-by-value would touch all four overrides, the virtual base declaration, both the `Symmetry_base` and `Symmetry` public APIs, and `Symmetry::combine_rule_` itself.

The `_`-suffix / output-arg convention is also used elsewhere in the codebase (e.g. `reverse_rule_`, `_Save`, `_Load`). Refactoring just `combine_rule_` would create an inconsistency, and refactoring the whole convention is out of scope for #771. Leaving the signature as-is; happy to file a follow-up issue for the cross-cutting convention change if you'd like.

---

pybind/symmetry_py.cpp
```diff
@@ -56,11 +74,17 @@ void symmetry_binding(py::module &m) {
...
       },
       py::arg("qnL"), py::arg("qnR"), py::arg("is_reverse") = false)
     .def(
-      "reverse_rule", [](Symmetry &self, const cytnx_int64 &qin) { return self.reverse_rule(qin); },
+      "reverse_rule",
+      [](Symmetry &self, const cytnx_int64 &qin) {
+        const cytnx_int64 norm = NormalizeZnInput(self, qin, "reverse_rule", "qin");
+        return self.reverse_rule(norm);
+      },
       py::arg("qin"))
```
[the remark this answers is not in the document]
IvanaGyro: Updated.

src/Symmetry.cpp
```diff
@@ -123,21 +128,28 @@ namespace cytnx {
...
+    for (cytnx_uint64 i = 0; i < inL.size(); i++) {
+      for (cytnx_uint64 j = 0; j < inR.size(); j++) {
+        out[i * inR.size() + j] = (inL[i] + inR[j]) % this->n;
+      }
     }
   }
   void cytnx::ZnSymmetry::combine_rule_(cytnx_int64 &out, const cytnx_int64 &inL,
                                         const cytnx_int64 &inR, const bool &is_reverse) {
-    const cytnx_int64 combined = NormalizeZnQnum(inL + inR, this->n);
+    ValidateZnQnum(inL, this->n);
```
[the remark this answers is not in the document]
IvanaGyro: Confirmed and fixed in `ba0aa4c` (inserted between the Python and C++ commits so bisect doesn't hit the regression window).

You're right that `_fx_get_total_fluxs` was applying `syms[0].reverse_rule` to every component `i` for the BRA-initial branch. The new fix replaces `syms[0]` with `syms[i]` at both copies (`include/UniTensor.hpp:1125` and `:1857`), which is what the rest of the function (the `j >= 1` loop and the non-BRA branch) was already doing.

I reproduced the crash beforehand with a `UniTensor({BD_BRA{U1, Zn(2)}, BD_KET{U1, Zn(2)}})` constructor — it threw `std::logic_error: qnum -1 is out of the valid range [0, 2)` from `ZnSymmetry::combine_rule_` because the U1 reverse on a Zn qnum `1` produced `-1`. After the fix the constructor completes and the result of `_fx_get_total_fluxs` is now consistent with the bond's own symmetry list.

Added `BlockUniTensorTest.FxGetTotalFluxsUsesPerComponentSymmetry` to lock this in; full ctest passes 938/938 on both `openblas-cpu` and `mkl-cpu` presets.

---

pcchen (commented):
## PR #837 Review

### Overview

This PR establishes a single contract for Zn quantum number arithmetic: valid inputs are canonical representatives in `[0, n)`. It applies a sensible two-tier policy — Python binding warns + normalizes out-of-range inputs with a `FutureWarning` (removal in v2.0.0), while the C++ API rejects them outright via `std::logic_error`. A latent bug in `_fx_get_total_fluxs` (`syms[0]` used instead of `syms[i]`) is fixed in the same PR, which the stricter validation would have surfaced as a hard crash.

---

### Strengths

- **Two-tier design is correct**: `FutureWarning` (visible to end users by default) is the right choice over `DeprecationWarning` (hidden) since these are user-facing inputs, not internal API. Using `FutureWarning` for Python while making C++ strict is a sound migration path.
- **The latent bug fix is well-motivated**: the PR explains clearly why the bug was silent before (uniform symmetry lists) and only surfaces after the stricter C++ rule, and includes a targeted regression test.
- **`NormalizeZnInput` warning message** is explicit, actionable, and consistent (includes version, argument name, valid range, and normalized value).
- **Test coverage is thorough**: C++ tests cover all public overloads including vector form and `is_reverse`; Python tests cover all six meaningful branches; the regression test for `_fx_get_total_fluxs` uses a genuinely mixed `{U1, Zn(2)}` list with a `BD_BRA` first bond — not a trivial case.
- **Double-loop rewrite** in `combine_rule_` vector overload is clearer than the flat-index form and semantically equivalent. Since both inputs are validated to `[0, n)`, the sum is always in `[0, 2n-2]`, so `% n` is safe without sign-handling.

---

### Issues

**Medium priority:**

1. **`#include <format>` requires C++20** (`pybind/symmetry_py.cpp:1`). The PR description says it matches existing codebase patterns, but worth confirming the project's CMakeLists.txt already enforces `-std=c++20` (or equivalent). If it currently targets C++17, this is a silent new requirement that only fails on older compilers. A quick check:
   ```sh
   grep -r "CXX_STANDARD\|cxx_std" CMakeLists.txt
   ```

2. **`ValidateZnQnums` short-circuits correctly via exception** (`src/Symmetry.cpp`) — on re-reading, `ValidateZnQnum` throws on first violation so `ValidateZnQnums` does exit early. This is fine as-is.

**Low priority:**

3. **`is_reverse=True` warning omits the flag in the message** (`pybind/symmetry_py.cpp:30`). A user calling `combine_rule(-1, 0, is_reverse=True)` sees a warning about `qnL` being out-of-range but no mention that `is_reverse` was set. The message is still actionable (the qnum is the problem), but slightly confusing since `is_reverse` changes the semantics. Consider appending ` (with is_reverse=True)` when applicable.

4. **`NormalizeZnInput` stacklevel of `2`** (`pybind/symmetry_py.cpp:32`). With pybind11's dispatch overhead the warning may point one frame off from the actual user callsite. The PR says this matches the `get_blocks_(silent=...)` pattern — if that works correctly in practice, this is fine. Worth smoke-testing that the warning captures the right file/line in user code.

5. **`BlockUniTensor_test.cpp:102`** — `EXPECT_NO_THROW` wrapping both the construction and `EXPECT_EQ` means if the constructor throws, the inner `EXPECT_EQ` is never evaluated and the test still passes the `NO_THROW` check. Consider separating:
   ```cpp
   UniTensor ut;
   EXPECT_NO_THROW(ut = UniTensor({bd_bra, bd_ket}));
   EXPECT_EQ(ut.syms(), syms);
   ```
   This way a throw on construction is clearly the failure point, and the `syms()` assertion is always checked on success.

---

### Minor Nits

- `src/Symmetry.cpp:135` — The `%` operator on `(inL[i] + inR[j])` relies on both being non-negative (guaranteed by prior validation). A one-line comment here would help future readers understand why the negative-modulo guard is absent, but this is stylistic.
- `tests/Bond_test.cpp` — The removed `ZnSymmetryNormalizesPublicRules` test had `EXPECT_TRUE(z2.check_qnum(z2.combine_rule(-1, 0)))` testing normalization end-to-end. The new tests are more explicit, which is better.

---

### Summary

The design and implementation are solid. The only actionable concern before merge is confirming C++20 is already a project requirement (item 1). Item 5 (`EXPECT_NO_THROW` scope) is a real but minor test-quality issue. Everything else is low-priority. This is a well-structured change with good test coverage and a clear migration story.

pcchen (approved):

IvanaGyro:
Thanks for the thorough review @pcchen, and for the approval. Responses to the open items:

**Item 1 — `#include <format>` / C++20 requirement.** Already enforced project-wide: `CMakeLists.txt:25-26` sets `CMAKE_CXX_STANDARD 20` and `CMAKE_CXX_STANDARD_REQUIRED ON`, and the existing `std::format` use in `pybind/unitensor_py.cpp` compiles under the current toolchain. So `<format>` here doesn't introduce a new requirement.

**Item 5 — `EXPECT_NO_THROW` scope.** I don't think the false-pass concern holds. `EXPECT_NO_THROW(stmt)` records a (non-fatal) test failure if `stmt` throws — so if the constructor throws, the test *fails* at the `EXPECT_NO_THROW`; it does not pass. The inner `EXPECT_EQ` being skipped on a throw is harmless because the throw already failed the test. On the success path, `EXPECT_EQ` doesn't throw — it records its result through gtest's normal channel even inside the block — so the `syms()` assertion is still checked. There's no path where a throwing constructor yields a green test.

The suggested rewrite (`UniTensor ut; EXPECT_NO_THROW(ut = ...); EXPECT_EQ(...)`) actually has a small downside: on a construction throw, `ut` stays default-constructed and the follow-up `EXPECT_EQ(ut.syms(), syms)` fires a *second, redundant* failure. I'd lean toward keeping the current form, but if you prefer the sharper failure attribution I'm happy to split it — let me know.

**Item 3 (low) — warning omits `is_reverse`.** Fair point; the qnum is the actionable part, but appending ` (with is_reverse=True)` when the flag is set would be clearer. Happy to add it if you'd like — say the word and I'll fold it into the Python-binding commit.

**Item 4 (low) — `PyErr_WarnEx` stacklevel `2`.** Matches the `get_blocks_(slient=...)` precedent in `pybind/unitensor_py.cpp`; in a smoke test the warning points at the user's call line rather than the pybind shim, so this looks correct in practice.

**Nit — `%` on validated inputs.** Reasonable; I can add a one-line comment noting that `inL`/`inR` are pre-validated to `[0, n)` so the sum is non-negative and the negative-modulo guard is intentionally absent. Will include it if I touch that commit again.

---

Issue #834, opened 2026-05-26
Refactor Tensor trace: low-level backend depends on UniTensor (API encapsulation violation)


IvanaGyro:
## Problem

The low-level `Tensor` trace implementation depends on the high-level `UniTensor` API and `Contract`. This inverts the intended layering: `Tensor` / `Storage` is the container layer and should not know about `UniTensor`, which is a higher-level abstraction built on top of `Tensor`.

### Where it happens

**CPU backend** &mdash; `src/backend/linalg_internal_cpu/Trace_internal.cpp:24-51`:

```cpp
template <class T>
void _trace_nd(Tensor &out, const Tensor &Tn, const cytnx_uint64 &Ndiag,
               const cytnx_uint64 &Nelem, const std::vector<cytnx_uint64> &accu,
               const std::vector<cytnx_uint64> &remain_rank_id,
               const std::vector<cytnx_int64> &shape, const cytnx_uint64 &ax1,
               const cytnx_uint64 &ax2) {
  cytnx::UniTensor I_UT = cytnx::UniTensor(eye(Ndiag, Tn.dtype()), false, -1);

  UniTensor UTn = UniTensor(Tn, false, 2);
  I_UT.relabel_({UTn._impl->_labels[ax1], UTn._impl->_labels[ax2]});

  out = Contract(I_UT, UTn).get_block_();
  // ...
}
```

The internal trace kernel wraps the input `Tensor` in a `UniTensor`, builds an identity `UniTensor`, and computes the trace via `Contract`. The whole point of the `linalg_internal` layer is to do raw per-dtype tensor work; pulling in `UniTensor` / `Contract` here means the container layer compiles against, and depends on the semantics of, the layer above it.

The dead branch left in the comments at lines 36-50 shows what a direct, container-level implementation would look like (loop over remain-rank indices and sum the diagonals); that is the kind of code that belongs at this layer.

### Same pattern in PR #749 (GPU)

PR [#749](https://github.com/Cytnx-dev/Cytnx/pull/749) ("implemented Trace for GPU", fixes #735) adds the GPU backend by *copying* the same anti-pattern into `src/backend/linalg_internal_gpu/cuTrace_internal.cu`:

```cpp
template <class T>
void _trace_2d_gpu(Tensor &out, const Tensor &Tn, const cytnx_uint64 &Ndiag) {
[...]

PR #849, opened 2026-05-29
fix(sum): use pairwise summation for floating-point linalg::Sum

### LINKED ISSUE #835: linalg::Sum accumulates floating-point error; use pairwise summation like NumPy
## Problem

`linalg::Sum(const Tensor&)` for floating-point dtypes (`Float`, `Double`, `ComplexFloat`, `ComplexDouble`) accumulates all elements into a single scalar in a serial loop:

```cpp
// src/backend/linalg_internal_cpu/Sum_internal.cpp:86-95
void Sum_internal_d(boost::intrusive_ptr<Storage_base> &out,
                    const boost::intrusive_ptr<Storage_base> &ten, const cytnx_uint64 &Nelem) {
  cytnx_double *_ten = (cytnx_double *)ten->data();
  cytnx_double *_out = (cytnx_double *)out->data();

  _out[0] = 0;
  for (cytnx_uint64 n = 0; n < Nelem; n++) {
    _out[0] += _ten[n];
  }
}
```

The same pattern is repeated for `_f`, `_cd`, `_cf` (lines 97-127). The worst-case rounding error of naive serial summation grows as `O(N·eps)`, and the test below already demonstrates the problem in practice.

## Existing disabled test

`tests/linalg_test/sum_test.cpp:52-66` contains a `DISABLED_Accuracy` test that fills 10000 elements with a fixed value and expects the sum to equal `value * 10000`. The values are deliberately chosen (e.g. `7.3315594031297398e+252` for `double`, `4.0565264e+31` for `float`, with carefully selected bit patterns shown in the comments) to expose round-off in naive accumulation.

The test was disabled in commit [`c9059ba`](https://github.com/Cytnx-dev/Cytnx/commit/c9059ba47fbbfbfbdc9b2a99acda133d46b8628a) with the message:

> We directly sum all numbers passed to `linalg::Sum()` on the CPU now. This method generates large errors for summing floating-point numbers and fails the accuracy test. To pass the online CI process, temporarily disable the test.

So the issue is known; this ticket tracks fixing the implementation so the test can be re-enabled.

## What NumPy does

Since NumPy 1.9, `np.add.reduce` (and therefore `np.sum`) uses **pairwise summation** along the reduction axis for floating-point dtypes. Pairwise summation has worst-case error growth `O(log N · eps)` instead of `O(N · eps)`, at essentially no extra cost &mdash; it is a divide-and-conquer recursion that switches to a small-block straight sum (NumPy uses block size 8) at the leaves. See NumPy source: `numpy/_core/src/umath/loops_utils.h.src` (`pairwise_sum_*`).

This is the right baseline for cytnx: same algorithm everyone else uses, no extra storage, no significantly worse performance than the naive loop, and it makes the existing accuracy test pass.

(Kahan / Neumaier compensated summation gives `O(eps)` error but at roughly 4&times; the FLOPs; usually overkill for a default reduction. Mentioning it as an option, but pairwise is the recommended default to match NumPy.)

## Where it happens

- CPU floating-point branches: `src/backend/linalg_internal_cpu/Sum_internal.cpp:86-127` (`Sum_internal_d`, `Sum_internal_f`, `Sum_internal_cd`, `Sum_internal_cf`).
- Integer branches (`_u64`, `_i64`, `_u32`, `_i32`, `_u16`, `_i16` at lines 14-84) are unaffected by rounding and can keep the straight loop.
- GPU equivalent: `src/backend/linalg_internal_gpu/cuSum_internal.cu` &mdash; should be audited for the same issue (cuBLAS reductions usually do tree reduction already, but worth checking).
- Caller: `src/linalg/Sum.cpp:9-31`.
- Test to re-enable: `tests/linalg_test/sum_test.cpp:52` (rename `DISABLED_Accuracy` &rarr; `Accuracy`), and the GPU counterpart `tests/gpu/linalg_test/sum_test.cpp`.

## Proposed direction

1. Replace the naive accumulator in `Sum_internal_d`, `Sum_internal_f`, `Sum_internal_cd`, `Sum_internal_cf` with pairwise summation, matching NumPy's approach (recursive split, small-block straight sum at the leaves &mdash; block size ~8 is what NumPy uses).
2. Audit `cuSum_internal.cu` and ensure the GPU reduction also has bounded error growth (a standard tree reduction inside CUB / `thrust::reduce` is sufficient).
3. Re-enable the accuracy test on both CPU and GPU.
4. Optionally factor the pairwise loop into a small templated helper so the four floating-point branches don't duplicate it.

## References

- Implementation: `src/backend/linalg_internal_cpu/Sum_internal.cpp:86-127`
- GPU implementation: `src/backend/linalg_internal_gpu/cuSum_internal.cu`
- Caller: `src/linalg/Sum.cpp`
- Disabled test: `tests/linalg_test/sum_test.cpp:52-66`
- Commit that disabled it: `c9059ba`
- NumPy pairwise summation: https://github.com/numpy/numpy/blob/main/numpy/_core/src/umath/loops_utils.h.src

IvanaGyro:
Fixes #835.

## Problem

`linalg::Sum` accumulated every element into one scalar in a serial loop, whose worst-case rounding error grows as `O(N · eps)`. The `Accuracy` test in `tests/linalg_test/sum_test.cpp` (10000 identical floating-point values whose bit patterns are chosen to expose round-off) failed under naive accumulation and was therefore disabled.

## Change

- Add a strided pairwise-summation primitive `src/backend/linalg_internal_cpu/pairwise_sum.hpp`:
  - `StrideView` — a C++20 random-access, sized view over every *k*-th element of a range.
  - `PairwiseSum` — NumPy `np.add.reduce`-style summation: straight loop below 8 elements, an 8-accumulator unrolled block up to 128, and a halve-and-recurse split (rounded to a multiple of 8) above that. Worst-case error growth drops to `O(log N · eps)` at essentially the same cost.
- Rewrite the `Float`/`Double`/`ComplexFloat`/`ComplexDouble` branches of `Sum_internal` to use `PairwiseSum`. Integer branches keep the straight loop (they do not round).
- Re-enable the accuracy test (`DISABLED_Accuracy` → `Accuracy`).
- The GPU path (`cuReduce_gpu`) already performs a block/tree reduction with bounded error, so it is unchanged and its `GpuAccuracy` test stays enabled.

`StrideView` is intentionally reusable for strided reductions (e.g. a matrix diagonal); the follow-up Trace refactor (#834) builds on it.

## Test plan

- [x] `openblas-cpu`: full suite **948/948** passed (includes the 10 re-enabled `LinalgSumHomogeneousValuesTest.Accuracy<...>` cases).
- [x] `mkl-cpu`: full suite **948/948** passed.
- [x] For the previously-disabled values the new code lands within **2 ULP** of the reference (`EXPECT_NUMBER_EQ` tolerates 4 ULP); naive accumulation was 108 ULP off.
- [ ] GPU CI (unchanged GPU path).

🤖 Draft — opened for review.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PhXYPqttRF669v9f2ek32X)_

### DIFF
--- a/src/backend/linalg_internal_cpu/pairwise_sum.hpp
+++ b/src/backend/linalg_internal_cpu/pairwise_sum.hpp
@@ -1,0 +1,62 @@
+#ifndef CYTNX_BACKEND_LINALG_INTERNAL_CPU_PAIRWISE_SUM_H_
+#define CYTNX_BACKEND_LINALG_INTERNAL_CPU_PAIRWISE_SUM_H_
+
+#include <cstddef>
+#include <iterator>
+#include <ranges>
+
+namespace cytnx {
+  namespace linalg_internal {
+
+    // Recursive (divide-and-conquer) core of the pairwise summation, matching
+    // NumPy's np.add.reduce: a straight loop for the smallest blocks, an
+    // eight-accumulator unrolled loop up to 128 elements, and a split into two
+    // halves (rounded to a multiple of eight) above that. Worst-case rounding
+    // error grows as O(log N * eps) instead of the O(N * eps) of a naive serial
+    // accumulation, at essentially the same cost.
+    template <class T, std::random_access_iterator It>
+    T PairwiseSumBlocks(It first, std::size_t n) {
+      if (n < 8) {
+        T res = T(0);
+        for (std::size_t i = 0; i < n; ++i) res += first[static_cast<std::ptrdiff_t>(i)];
+        return res;
+      }
+      if (n <= 128) {
+        T r0 = first[0], r1 = first[1], r2 = first[2], r3 = first[3];
+        T r4 = first[4], r5 = first[5], r6 = first[6], r7 = first[7];
+        std::size_t i = 8;
+        for (; i + 8 <= n; i += 8) {
+          auto p = first + static_cast<std::ptrdiff_t>(i);
+          r0 += p[0];
+          r1 += p[1];
+          r2 += p[2];
+          r3 += p[3];
+          r4 += p[4];
+          r5 += p[5];
+          r6 += p[6];
+          r7 += p[7];
+        }
+        T res = ((r0 + r1) + (r2 + r3)) + ((r4 + r5) + (r6 + r7));
+        for (; i < n; ++i) res += first[static_cast<std::ptrdiff_t>(i)];
+        return res;
+      }
+      std::size_t half = n / 2;
+      half -= half % 8;
+      return PairwiseSumBlocks<T>(first, half) +
+             PairwiseSumBlocks<T>(first + static_cast<std::ptrdiff_t>(half), n - half);
+    }
+
+    // Pairwise sum over a random-access range. The element type is deduced from
+    // the range. A contiguous std::span sums every element; pass a strided view
+    // (see stride_view.hpp) to sum a strided sequence such as a matrix diagonal.
+    template <std::ranges::random_access_range R>
+    std::ranges::range_value_t<R> PairwiseSum(R&& range) {
+      using T = std::ranges::range_value_t<R>;
+      return PairwiseSumBlocks<T>(std::ranges::begin(range),
+                                  static_cast<std::size_t>(std::ranges::size(range)));
+    }
+
+  }  // namespace linalg_internal
+}  // namespace cytnx
+
+#endif  // CYTNX_BACKEND_LINALG_INTERNAL_CPU_PAIRWISE_SUM_H_
--- a/tests/linalg_test/sum_test.cpp
+++ b/tests/linalg_test/sum_test.cpp
@@ -49,7 +49,7 @@ namespace cytnx {
    * Note: `cytnx_bool` is not supported for the `linalg::Sum()` function.
    * This test also assesses the accuracy of summing floating-point numbers.
    */
-  TYPED_TEST(LinalgSumHomogeneousValuesTest, DISABLED_Accuracy) {
+  TYPED_TEST(LinalgSumHomogeneousValuesTest, Accuracy) {
     TypeParam value = LinalgSumHomogeneousValuesTest<TypeParam>::value;
     int element_number = 10000;
     unsigned int dtype = Type_class().cy_typeid(value);
--- a/tests/linalg_test/sum_test.cpp
+++ b/tests/linalg_test/sum_test.cpp
@@ -64,4 +64,81 @@ namespace cytnx {
 
     EXPECT_NUMBER_EQ(sum_result.at<TypeParam>({0}), value * static_cast<TypeParam>(element_number));
   }
+
+  /**
+   * Exercises every branch of `PairwiseSumBlocks` -- the recursive core that
+   * `linalg::Sum` dispatches floating-point reductions through.
+   *
+   *   * n < 8         : straight serial loop
+   *   * 8 <= n <= 128 : 8-accumulator unrolled body, optionally with a scalar
+   *                     tail when n % 8 != 0
+   *   * n > 128       : recursive split into two halves rounded to a multiple
+   *                     of 8
+   *
+   * The original Accuracy test only covers the n = 10000 recursive-split case;
+   * a regression in either of the small-n branches would not be caught there.
+   * Sizes 7/8/9/15/128/129/137 straddle the thresholds (including the off-by-one
+   * tail cases). The expected result is exact, so any branch that drops or
+   * double-counts an element fails immediately.
+   */
+  TEST(LinalgSumBoundaryTest, EachPairwiseSumBranch) {
+    const cytnx_double value = 1.0;
+    for (int n : {1, 7, 8, 9, 15, 128, 129, 137, 1024}) {
+      Tensor tensor(/* shape */ {static_cast<unsigned long>(n)}, Type.Double, Device.cpu,
+                    /* init_zero */ false);
+      tensor.fill(value);
+      Tensor sum_result = linalg::Sum(tensor);
+      EXPECT_EQ(sum_result.shape().size(), 1);
+      EXPECT_EQ(sum_result.shape()[0], 1);
+      EXPECT_DOUBLE_EQ(sum_result.at<cytnx_double>({0}), value * static_cast<cytnx_double>(n))
+        << "n=" << n;
+    }
+  }
+
+  /**
+   * The dynamic-range case that motivates pairwise summation, and the one
+   * input distribution where naive serial accumulation visibly fails.
+   *
+   * Both arrays are [+L, 1, 1, ..., 1, -L] with L far above the precision
+   * threshold (2^53 for double, 2^23 for float); the exact sum is N - 2. Under
+   * naive accumulation the running total reaches L on the first element and
+   * the subsequent +1's vanish in IEEE 754 rounding -- 1.0 is below the unit
+   * in the last place at that magnitude -- so naive returns ~0. Pairwise
+   * keeps small values together in the tree until the cancellation of +/-L at
+   * the top, so the small terms survive (modulo a handful of 1's in the
+   * unrolled-accumulator blocks that hold L itself).
+   *
+   * The contract this asserts is qualitative on purpose: any reasonable
+   * pairwise implementation must land much closer to N - 2 than to 0. A
+   * serial-accumulation regression would collapse to ~0 and fail
+   * `EXPECT_GT(result, N / 2)` immediately; the exact pairwise result depends
+   * on which accumulators receive +/-L (a few small terms are lost there).
+   */
+  TEST(LinalgSumHeterogeneousMagnitudeTest, RecoversTermsLostByNaiveAccumulation_Double) {
+    constexpr int N = 1024;
+    Tensor tensor(/* shape */ {static_cast<unsigned long>(N)}, Type.Double, Device.cpu,
+                  /* init_zero */ false);
+    tensor.fill(static_cast<cytnx_double>(1));
+    tensor.at<cytnx_double>({0}) = 1e16;
+    tensor.at<cytnx_double>({N - 1}) = -1e16;
+    Tensor sum_result = linalg::Sum(tensor);
+    const cytnx_double result = sum_result.at<cytnx_double>({0});
+    EXPECT_GT(result, static_cast<cytnx_double>(N / 2));
+    EXPECT_LE(result, static_cast<cytnx_double>(N));
+  }
+
+  TEST(LinalgSumHeterogeneousMagnitudeTest, RecoversTermsLostByNaiveAccumulation_Float) {
+    constexpr int N = 1024;
+    Tensor tensor(/* shape */ {static_cast<unsigned long>(N)}, Type.Float, Device.cpu,
+                  /* init_zero */ false);
+    tensor.fill(static_cast<cytnx_float>(1));
+    // float has ~7.2 decimal digits of precision; 1e8 already exceeds 2^23, so
+    // a serial `1e8 + 1` collapses to 1e8 and the unit terms are lost.
+    tensor.at<cytnx_float>({0}) = 1e8f;
+    tensor.at<cytnx_float>({N - 1}) = -1e8f;
+    Tensor sum_result = linalg::Sum(tensor);
+    const cytnx_float result = sum_result.at<cytnx_float>({0});
+    EXPECT_GT(result, static_cast<cytnx_float>(N / 2));
+    EXPECT_LE(result, static_cast<cytnx_float>(N));
+  }
 }  // namespace cytnx
--- a/src/backend/linalg_internal_cpu/Sum_internal.cpp
+++ b/src/backend/linalg_internal_cpu/Sum_internal.cpp
@@ -1,8 +1,11 @@
 #include "Sum_internal.hpp"
 
+#include <span>
+
 #include "boost/smart_ptr/intrusive_ptr.hpp"
 
 #include "backend/Storage.hpp"
+#include "backend/linalg_internal_cpu/pairwise_sum.hpp"
 #include "cytnx_error.hpp"
 #include "Type.hpp"
 
@@ -88,10 +91,7 @@ namespace cytnx {
       cytnx_double *_ten = (cytnx_double *)ten->data();
       cytnx_double *_out = (cytnx_double *)out->data();
 
-      _out[0] = 0;
-      for (cytnx_uint64 n = 0; n < Nelem; n++) {
-        _out[0] += _ten[n];
-      }
+      _out[0] = PairwiseSum(std::span<const cytnx_double>(_ten, Nelem));
     }
 
     void Sum_internal_f(boost::intrusive_ptr<Storage_base> &out,
@@ -99,20 +99,14 @@ namespace cytnx {
       cytnx_float *_ten = (cytnx_float *)ten->data();
       cytnx_float *_out = (cytnx_float *)out->data();
 
-      _out[0] = 0;
-      for (cytnx_uint64 n = 0; n < Nelem; n++) {
-        _out[0] += _ten[n];
-      }
+      _out[0] = PairwiseSum(std::span<const cytnx_float>(_ten, Nelem));
     }
     void Sum_internal_cd(boost::intrusive_ptr<Storage_base> &out,
                          const boost::intrusive_ptr<Storage_base> &ten, const cytnx_uint64 &Nelem) {
       cytnx_complex128 *_ten = (cytnx_complex128 *)ten->data();
       cytnx_complex128 *_out = (cytnx_complex128 *)out->data();
 
-      _out[0] = 0;
-      for (cytnx_uint64 n = 0; n < Nelem; n++) {
-        _out[0] += _ten[n];
-      }
+      _out[0] = PairwiseSum(std::span<const cytnx_complex128>(_ten, Nelem));
     }
 
     void Sum_internal_cf(boost::intrusive_ptr<Storage_base> &out,
@@ -120,10 +114,7 @@ namespace cytnx {
       cytnx_complex64 *_ten = (cytnx_complex64 *)ten->data();
       cytnx_complex64 *_out = (cytnx_complex64 *)out->data();
 
-      _out[0] = 0;
-      for (cytnx_uint64 n = 0; n < Nelem; n++) {
-        _out[0] += _ten[n];
-      }
+      _out[0] = PairwiseSum(std::span<const cytnx_complex64>(_ten, Nelem));
     }
 
     void Sum_internal_b(boost::intrusive_ptr<Storage_base> &out,


src/backend/linalg_internal_cpu/pairwise_sum.hpp
```diff
@@ -0,0 +1,165 @@
...
+  namespace linalg_internal {
+
+    // A view over every step-th element of a random-access, sized range.
+    //
+    // Used to express the strided element sequences that PairwiseSum consumes
+    // (e.g. the diagonal of a matrix), so the same summation routine serves both
+    // contiguous reductions and strided ones.
+    template <std::ranges::view V>
+      requires std::ranges::random_access_range<V> && std::ranges::sized_range<V>
+    class StrideView : public std::ranges::view_interface<StrideView<V>> {
```
IvanaGyro: StrideView  is a STL-like template, according to the style guie, the naming should be `stride_view`

IvanaGyro: If we move StrideView to the more gneral file, the function in the future can call
```
PairwiseSum( std::span<const T>(data, extent)|  StrideView(diag_stride));
```
instead of
```
PairwiseSum( StrideView(std::span<const T>(data, extent),diag_stride));
```

IvanaGyro: (Comment written by Claude on behalf of @IvanaGyro.)

`StrideView` is never instantiated in this PR: the only caller, `Sum_internal.cpp`, always passes `std::span<...>`, so `PairwiseSum` is only ever exercised with a contiguous span. A class template that is never instantiated is only partially compiled (non-dependent syntax is checked, but the member bodies and concept satisfaction are not), and it has zero test coverage. That is precisely why the dangling-pointer bug @gemini-code-assist flagged, and the naming, slipped through — nothing here forces the compiler or the test suite to look at this code.

Two ways to resolve it, either works:
- Add a unit test that builds a `StrideView` over e.g. a strided `std::span` and feeds it to `PairwiseSum`, asserting it sums the right elements (and validating the `random_access_iterator` model). That would have caught the dangling pointer.
- Defer `StrideView` to #834 (the Trace refactor that actually consumes it), keeping this PR to the contiguous-sum change it needs. The strided view can land there with the test that exercises it.

I'd lean toward deferring it, but if you want to keep it here, please add the test so the iterator is actually validated before it's relied on.

IvanaGyro (commented):

The pairwise-summation core looks correct to me. `PairwiseSumBlocks` covers every element exactly once across the three regimes (`<8` straight loop, the 8-accumulator block for `[8,128]` with its scalar tail, and the `n>128` split where `half` is rounded down to a multiple of 8 and is always ≥64), and the reduction order matches NumPy's `np.add.reduce`. Empty ranges return `T(0)`, preserving the old behavior. The `Sum_internal` rewrite and the test re-enable are clean.

One additive note on top of the existing comments — see the inline comment about `StrideView` being unexercised in this PR.

IvanaGyro:
(Comment written by Claude on behalf of @IvanaGyro.)

**Test-coverage note** (in addition to the `StrideView` point in the review): the re-enabled `Accuracy` test sums 10000 *identical* values. That exercises the recursive split, but not the block-size boundaries nor mixed-magnitude inputs — which is exactly where pairwise summation diverges from naive accumulation. Consider a small direct `PairwiseSum` unit test covering `n = 0, 1, 7, 8, 9, 128, 129` with a mix of large/small/negative values, compared against a higher-precision (e.g. `long double` / Kahan) reference, so the 8-accumulator block edge and the halve-and-recurse boundary are pinned and can't silently regress.

src/backend/linalg_internal_cpu/pairwise_sum.hpp
IvanaGyro: Going with deferral. `stride_view` is gone from this PR and lands in #850 with all the follow-ups in this thread cluster:

- iterator stores the underlying base iterator instead of a pointer to the view, so iterators stay valid independent of the view's lifetime (fixes @gemini-code-assist's dangling-pointer thread).
- renamed to snake_case `stride_view`.
- `r | stride(k)` range-adaptor closure for the pipe form.
- dedicated unit test `tests/linalg_test/stride_view_test.cpp` exercising the iterator concept model, the pipe form, the iterator-outlives-view case, and complex strided summation.

Kept under `linalg_internal` per the option you picked. #849 is now only the contiguous pairwise-sum change.

---

src/backend/linalg_internal_cpu/pairwise_sum.hpp
```diff
@@ -0,0 +1,165 @@
...
+        Iterator(const V* base, std::size_t step, std::size_t index)
+            : base_(base), step_(step), index_(index) {}
+
+        reference operator*() const {
+          return std::ranges::begin(*base_)[static_cast<difference_type>(index_ * step_)];
+        }
+        reference operator[](difference_type n) const {
+          return std::ranges::begin(
+            *base_)[static_cast<difference_type>((index_ + static_cast<std::size_t>(n)) * step_)];
+        }
```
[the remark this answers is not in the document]
IvanaGyro: Addressed in #850 (commit `c96e09e`). The redesigned `stride_view` stores `std::ranges::iterator_t<const V>` in the iterator instead of a pointer to the view, so iterators stay valid for as long as the underlying data lives, independent of the view's lifetime. `StrideViewTest.IteratorsOutliveTheView` covers exactly the scenario you described. (Deferred from this PR along with the type itself.)

---

src/backend/linalg_internal_cpu/pairwise_sum.hpp
IvanaGyro: Renamed to snake_case `stride_view` and moved to #850 commit `c96e09e`.

---

IvanaGyro: Pipe form implemented: `std::span<...>(data, extent) | stride(k)` — used inside the CPU trace and covered by `StrideViewTest.PipeAdaptorFeedsPairwiseSum`. Per the option you picked I kept the header under `linalg_internal` rather than `include/utils/`. In #850 commit `c96e09e`.

---

src/backend/linalg_internal_cpu/pairwise_sum.hpp
```diff
@@ -0,0 +1,165 @@
...
+      StrideView(V base, std::size_t step) : base_(std::move(base)), step_(step) {
+        if (step_ == 0) throw std::invalid_argument("StrideView: stride must be positive");
+      }
+
+      std::size_t size() const {
+        const auto n = static_cast<std::size_t>(std::ranges::size(base_));
+        return n / step_ + (n % step_ != 0);
+      }
+      Iterator begin() const { return Iterator(&base_, step_, 0); }
+      Iterator end() const { return Iterator(&base_, step_, size()); }
```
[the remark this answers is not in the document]
IvanaGyro: Covered by the same redesign — `begin()`/`end()` now pass `std::ranges::begin(base_)` into the iterator, which stores that base iterator. See #850 `c96e09e`.

---

tests/linalg_test/stride_view_test.cpp
```diff
@@ -0,0 +1,77 @@
+#include <gtest/gtest.h>
```
IvanaGyro: (Comment written by Claude on behalf of @IvanaGyro.)

This new test file isn't wired into the build, so it never compiles or runs. `tests/CMakeLists.txt` lists test sources **explicitly** (no globbing) and this PR doesn't touch it, so `stride_view_test.cpp` is absent from the `add_executable(test_main …)` source list (`sum_test.cpp` is line 46). Net effect: the `static_assert`s and the `IteratorsOutliveTheView` regression test silently don't execute — and CI stays green — which defeats the point of adding them to address the earlier "the template is never instantiated/tested" feedback.

Add it next to `sum_test.cpp` in `tests/CMakeLists.txt` (and re-run cmake):

```cmake
  linalg_test/stride_view_test.cpp
```

Two small coverage gaps to add while you're here (the unit tests currently only use strides that evenly divide the range):
- A non-dividing case, e.g. `span` of size 10 with `stride(3)` → `size() == 4`, elements `0,3,6,9`. This is exactly the shape the Trace caller produces (`extent = (Ndiag-1)*step + 1`, so `n % step == 1`), and it's the `n % step != 0` branch of `size()` — currently never exercised.
- The `step == 0` precondition: `EXPECT_THROW(stride_view(span, 0), std::invalid_argument);`

IvanaGyro (commented):

Thanks for addressing the earlier round — `stride_view`/`iterator` naming, the dangling-pointer fix (iterator now stores the base iterator, with a dedicated `IteratorsOutliveTheView` test), the move to its own `stride_view.hpp`, the `| stride(k)` pipe adaptor, and the `static_assert`s on the range concepts all look good, and the fix is correct. One blocking issue: the new test file isn't registered in the build — details inline.

src/backend/linalg_internal_cpu/stride_view.hpp
```diff
@@ -0,0 +1,130 @@
...
+    // diagonal of a matrix laid out with a fixed stride. It lets a single
+    // reduction routine (PairwiseSum) consume both contiguous and strided
+    // sequences.
+    //
+    // The iterator stores the underlying range's iterator rather than a pointer
+    // to the view, so iterators stay valid for as long as the underlying data
+    // lives, independent of the stride_view object's lifetime.
+    template <std::ranges::view V>
+    requires std::ranges::random_access_range<V> && std::ranges::sized_range<V>
+    class stride_view : public std::ranges::view_interface<stride_view<V>> {
```
IvanaGyro: Move this file and tests to #850 because it is not used in this PR

tests/linalg_test/stride_view_test.cpp
IvanaGyro: Done — moved out of #849. `stride_view.hpp` and `tests/linalg_test/stride_view_test.cpp` now land in #850's CPU commit (`5e38cb9`), where the CPU Trace actually instantiates the template. The test is also wired into the build there (added to `tests/CMakeLists.txt` and exposed via `src/` on the test target's include path), so the static_asserts and the iterator-outlives-view regression actually execute.

Coverage you flagged is in too: a non-dividing `stride(3)` over a size-10 span (the `n % step != 0` branch of `size()`, matching the shape Trace produces — `extent = (Ndiag - 1) * step + 1`), and the `stride_view(span, 0)` precondition asserted with `EXPECT_THROW(..., std::invalid_argument)`.

---

src/backend/linalg_internal_cpu/stride_view.hpp
IvanaGyro: Done — both `stride_view.hpp` and `stride_view_test.cpp` have moved to #850's CPU commit `5e38cb9`. #849 is now only the contiguous pairwise-Sum change. Force-pushed.

---

pcchen:
## PR #849 Review: `fix(sum)` — pairwise summation for floating-point `linalg::Sum`

### Overview

This PR replaces the naive serial accumulation loop in `Sum_internal` (float/double/complexfloat/complexdouble) with a pairwise summation algorithm, reducing worst-case rounding error from `O(N·eps)` to `O(log N·eps)`. The re-enabled `Accuracy` test confirms the fix works in practice (within 2 ULP vs. 108 ULP for naive).

The algorithm mirrors NumPy's `np.add.reduce`:
- `n < 8`: straight loop
- `8 ≤ n ≤ 128`: 8-accumulator unrolled block + tail
- `n > 128`: split at `n/2` rounded down to a multiple of 8, recurse

---

### Correctness

The algorithm is correct and the boundary cases are handled:

- **n = 0**: falls into `n < 8` path, loop doesn't execute, returns `T(0)`. ✅
- **n = 8**: initialises `r0..r7`, inner loop skipped (`i+8 > n`), tail skipped, returns tree sum of 8 accumulators. ✅
- **n = 9**: inner loop skipped, tail adds `first[8]` to the tree sum. ✅
- **Complex types**: `T(0)` constructs `{0.0, 0.0}` for `std::complex<T>`, all arithmetic works element-wise. ✅
- **Half-rounding can't produce zero**: for `n > 128`, `half = n/2 ≥ 64`, and `half - half%8 ≥ 56 > 0`, so the recursion always terminates. ✅
- **Recursion depth**: `O(log₂(N/128))` — for N = 1 billion, depth ≈ 23, no stack concern. ✅

---

### Issues

#### 1. C++20 requirement needs explicit confirmation (Important)

`pairwise_sum.hpp` requires C++20:
- `#include <ranges>` — C++20
- `std::ranges::random_access_range`, `std::ranges::range_value_t`, `std::ranges::begin`, `std::ranges::size` — C++20 concepts/functions
- `std::random_access_iterator` — C++20 concept
- `std::span` in `Sum_internal.cpp` — C++20

The PR passes the full test suite, so the build system already compiles with C++20. But if `CMakeLists.txt` hasn't been updated to require C++20 (e.g., `set(CMAKE_CXX_STANDARD 20)`), a clean build on a C++17-defaulting toolchain will fail. Please confirm `CMAKE_CXX_STANDARD` is already set to ≥ 20, or update it in this PR.

#### 2. Comment references non-existent `stride_view.hpp` (Minor)

```cpp
// pass a strided view (see stride_view.hpp) to sum a strided sequence such as a matrix diagonal.
```

`stride_view.hpp` does not exist yet — it's deferred to the follow-up Trace PR (#834). A reader encountering this comment today will search for a file that isn't there. Change to:

```cpp
// pass a strided view (see issue #834) to sum a strided sequence such as a matrix diagonal.
```

or remove the parenthetical until #834 lands.

#### 3. Missing `<complex>` include in `pairwise_sum.hpp` (Minor)

`T(0)` for complex types requires `std::complex` to be visible. Currently this is satisfied transitively via the callers, but `pairwise_sum.hpp` should be self-contained. Add `#include <complex>` to the header.

#### 4. Why `half` must be a multiple of 8 — comment is missing (Minor)

```cpp
std::size_t half = n / 2;
half -= half % 8;
```

This alignment ensures child calls can enter the unrolled path, but there's no comment explaining it. A brief note like `// align to block size so the child call enters the unrolled path` would help future readers.

---

### Strengths

- Clean, self-contained header with a clear algorithm structure.
- Handles empty, small, medium, and large inputs correctly.
- The `PairwiseSum` wrapper accepts any random-access range, enabling future reuse for strided reductions without changing the core algorithm.
- Re-enabling the `Accuracy` test (rather than just removing `DISABLED_`) is the right approach — it now serves as a permanent regression guard.
- Integer types correctly keep the straight loop (no rounding needed).
- GPU path is correctly left unchanged.

---

### Recommendation

**Approve after addressing:**
1. Confirm (or add) `CMAKE_CXX_STANDARD 20` in `CMakeLists.txt`.
2. Fix the dangling `stride_view.hpp` reference in the comment.
3. (Optional) Add `#include <complex>` to `pairwise_sum.hpp` and a comment on the `half % 8` alignment.

tests/linalg_test/sum_test.cpp
```diff
@@ -49,7 +49,7 @@ namespace cytnx {
    * Note: `cytnx_bool` is not supported for the `linalg::Sum()` function.
    * This test also assesses the accuracy of summing floating-point numbers.
    */
-  TYPED_TEST(LinalgSumHomogeneousValuesTest, DISABLED_Accuracy) {
+  TYPED_TEST(LinalgSumHomogeneousValuesTest, Accuracy) {
     TypeParam value = LinalgSumHomogeneousValuesTest<TypeParam>::value;
     int element_number = 10000;
```
IvanaGyro, line 54: (Review written by Claude on behalf of @IvanaGyro.)

This single `element_number = 10000` case only exercises the recursive-split branch (plus the 8-accumulator base). The two boundary branches in `PairwiseSumBlocks` go untested: the serial `n < 8` path and the `8 <= n <= 128` unrolled path, including the off-by-one tails (`n` not a multiple of 8). A regression that broke either branch would still pass CI.

Suggest adding sizes that straddle the thresholds, e.g. `7, 8, 9, 128, 129`, in addition to `10000` — a small value-parameterized companion would cover all three code paths cheaply.

tests/linalg_test/sum_test.cpp
```diff
@@ -49,7 +49,7 @@ namespace cytnx {
    * Note: `cytnx_bool` is not supported for the `linalg::Sum()` function.
    * This test also assesses the accuracy of summing floating-point numbers.
    */
-  TYPED_TEST(LinalgSumHomogeneousValuesTest, DISABLED_Accuracy) {
+  TYPED_TEST(LinalgSumHomogeneousValuesTest, Accuracy) {
```
IvanaGyro, line 52: (Review written by Claude on behalf of @IvanaGyro.)

All the FP accuracy assertions here use *homogeneous* (identical) values, the one input distribution where even naive serial summation stays nearly exact. That doesn't demonstrate or guard the actual benefit of pairwise summation — improved accuracy under a large dynamic range of magnitudes.

Consider one heterogeneous-magnitude test for `cytnx_double`/`cytnx_float` where naive accumulation visibly loses precision but pairwise recovers it, e.g. a tensor like `[1e16, 1, 1, ..., 1, -1e16]` (the `1`s vanish in a naive running sum once it reaches `1e16`, but the balanced tree keeps them). That both documents intent and locks in the O(log N) error behavior.

IvanaGyro (commented):

The switch to NumPy-style pairwise summation is a clean, well-documented improvement, and re-enabling the previously `DISABLED_Accuracy` test is the right move. The implementation looks correct: the empty range returns `T(0)`, the `<8` / `<=128` / recursive-split branches are well-formed, and the powers-of-two block structure is what lets the homogeneous-value test now pass exactly. Two test-coverage gaps below; no blocking concerns with the production code itself.

tests/linalg_test/sum_test.cpp
IvanaGyro: Done in `ed5a88b`. Added `LinalgSumBoundaryTest.EachPairwiseSumBranch` that iterates over sizes `{1, 7, 8, 9, 15, 128, 129, 137, 1024}` summing a homogeneous-`1.0` `cytnx_double` tensor and asserting the result equals `N` exactly:

```cpp
TEST(LinalgSumBoundaryTest, EachPairwiseSumBranch) {
  const cytnx_double value = 1.0;
  for (int n : {1, 7, 8, 9, 15, 128, 129, 137, 1024}) {
    Tensor tensor({static_cast<unsigned long>(n)}, Type.Double, Device.cpu, false);
    tensor.fill(value);
    Tensor sum_result = linalg::Sum(tensor);
    EXPECT_DOUBLE_EQ(sum_result.at<cytnx_double>({0}), value * static_cast<cytnx_double>(n))
      << "n=" << n;
  }
}
```

Branch-to-size mapping:

| Branch | Sizes |
|---|---|
| `n < 8` serial loop | 1, 7 |
| `n <= 128` 8-accumulator, no tail | 8, 128 |
| `n <= 128` 8-accumulator + scalar tail (`n % 8 != 0`) | 9, 15, 137 |
| `n > 128` recursive split | 129, 137, 1024 |

(`137` straddles two — it's `8*17+1 = 137` so the recursive split lands a sub-block in the tail branch.)

All 9 cases pass exactly; a regression that dropped or double-counted any element would fail with a value-tagged `n=...` message.

---

tests/linalg_test/sum_test.cpp
IvanaGyro: Done in `ed5a88b`. Added `LinalgSumHeterogeneousMagnitudeTest.RecoversTermsLostByNaiveAccumulation_{Double,Float}` summing `[+L, 1, 1, ..., 1, -L]` with `N = 1024`, `L = 1e16` for double (above `2^53`), `L = 1e8` for float (above `2^23`):

```cpp
TEST(LinalgSumHeterogeneousMagnitudeTest, RecoversTermsLostByNaiveAccumulation_Double) {
  constexpr int N = 1024;
  Tensor tensor({static_cast<unsigned long>(N)}, Type.Double, Device.cpu, false);
  tensor.fill(static_cast<cytnx_double>(1));
  tensor.at<cytnx_double>({0}) = 1e16;
  tensor.at<cytnx_double>({N - 1}) = -1e16;
  Tensor sum_result = linalg::Sum(tensor);
  const cytnx_double result = sum_result.at<cytnx_double>({0});
  EXPECT_GT(result, static_cast<cytnx_double>(N / 2));   // not naive's ~0
  EXPECT_LE(result, static_cast<cytnx_double>(N));        // not overcounting
}
```

The contract asserted is intentionally qualitative — `N/2 < result <= N` — rather than `EXPECT_NEAR(result, N - 2, eps)`. Reason: the existing pairwise implementation routes `+L` into one accumulator (`r0`) of the 8-accumulator block that contains it, and the 16 subsequent `1`s that `r0` picks up in that block's unrolled loop _do_ get lost in IEEE 754 rounding (`1e16 + 1 == 1e16`). Same for `-L` in the block at the tail. Empirically the double test lands at `~992` and the float test at `~1008` (out of `1022` expected), so:

- The looser `N/2 < result` bound is tight enough that a serial-accumulation regression — which would collapse the running total to `~0` once it reaches `±L` — fails immediately.
- The upper `result <= N` bound catches any double-counting.
- The exact pairwise result is implementation-dependent on which accumulators capture `±L`; pinning it to a precise expected value would be fragile to harmless refactors of the unrolled block.

This documents pairwise's benefit and locks in the only regression that actually matters here: a fallback to serial summation.

---

pcchen:
## PR #849 Issue Status

### All Issues Resolved ✅

| Issue | Resolution |
|---|---|
| **Gemini (high): Dangling pointer in `StrideView::Iterator`** | Deferred to PR #850 — redesigned iterator stores `std::ranges::iterator_t<const V>` instead of a pointer to the view; covered by `StrideViewTest.IteratorsOutliveTheView` |
| **Gemini (high): `begin()`/`end()` to pass underlying iterator** | Same PR #850 fix — both now forward the base iterator |
| **IvanaGyro: `StrideView` → `stride_view` (naming convention)** | Renamed in PR #850 |
| **IvanaGyro: Move `StrideView` to a more general file with pipe adaptor** | PR #850: `r \| stride(k)` adaptor implemented; kept under `linalg_internal` |
| **IvanaGyro: `stride_view.hpp` unused/uninstantiated in this PR** | Both `stride_view.hpp` and `stride_view_test.cpp` removed from #849 and land in #850 where CPU Trace instantiates the template |
| **IvanaGyro: `stride_view_test.cpp` not wired into build** | Fixed in PR #850 — added to `tests/CMakeLists.txt` |
| **IvanaGyro review: Tests only exercise recursive-split branch** | Fixed in `ed5a88b` — `LinalgSumBoundaryTest.EachPairwiseSumBranch` covers sizes `{1, 7, 8, 9, 15, 128, 129, 137, 1024}`, hitting the serial (`< 8`), unrolled (`8–128`), and recursive (`> 128`) branches including off-by-one tails |
| **IvanaGyro review: Tests use only homogeneous values** | Fixed in `ed5a88b` — `LinalgSumHeterogeneousMagnitudeTest` for both `Double` and `Float`: `[+L, 1, 1, …, 1, -L]` with `L` above the precision threshold, where naive serial accumulation collapses to ≈0 |

### Current Scope of PR #849

The diff is now clean and focused:
- `Sum_internal.cpp` — floats dispatch through `PairwiseSum`; integers keep the serial loop
- `pairwise_sum.hpp` — pure pairwise algorithm with no `StrideView` (just a comment cross-reference to `stride_view.hpp` for context)
- `sum_test.cpp` — re-enabled `Accuracy` test + boundary + heterogeneous-magnitude tests

### Minor Observation

The comment in `pairwise_sum.hpp` (`// see stride_view.hpp`) references a file that doesn't exist until PR #850 merges. Cosmetic only, but worth merging #850 first or concurrently.

All substantive issues resolved. Ready to merge.

pcchen (approved):

PR #862, opened 2026-05-31
refactor(trace): compute the CPU Tensor trace in the container layer

IvanaGyro:
Part of #834.

> **Stacked on #860 (stride_view).** This PR's diff is only the CPU implementation. #850 builds on this for the GPU side.

## Problem

The low-level `Tensor` trace built an identity `UniTensor` and called `Contract`, so the container-level `linalg_internal` layer depended on the higher-level `UniTensor` / `Contract` APIs — an inverted layering. Compute the trace directly over storage instead, and tighten the internal Trace API at the same time.

## Change

- **Simplified dispatch API**: each `Trace_internal_*` (CPU) and `cuTrace_internal_*` (GPU) now takes only `(Tn, ax1, ax2)` and returns the result `Tensor`. The per-dtype dispatchers each call a single templated `TraceImpl` in an unnamed namespace that derives `diagonal_length` / `output_size` / per-surviving-axis layout / the `output_is_scalar` flag inline from `Tn.shape()` and `Tn.strides()`. `Trace.cpp` is just validation + a dtype-keyed dispatch call.
- **Public `Tensor::strides()`** getter (snake_case, mirrors `shape()` / NumPy's `.strides`): for each logical axis, the storage stride derived from `_shape` and `_invmapper`. The CPU trace calls it; #850 reuses it for the GPU.
- **Stride-aware diagonal sum**: `PairwiseSum` over the `stride_view` adaptor (from #860) with diagonal stride = `strides[ax1] + strides[ax2]`. The previous `contiguous()` copy is dropped; floating-point traces get the same `O(log N · ε)` error bound as `linalg::Sum`. `diagonal_length == 0` / `output_size == 0` guarded with an early zeroed return.
- **`Tensor::from_storage`** is used to compose the output: fill a flat result `Storage` directly, then promote it to a `Tensor` once the diagonal sums are written, so the storage layer stays at the storage layer until the final return.
- **Odometer ND walk**: walks the output elements in row-major order carrying the diagonal's start offset on an odometer — each step bumps the last surviving-axis index (with carry into earlier axes on wrap) and adjusts the offset by the affected axes' input strides. Avoids the per-element division and modulo of decoding the flat index, and needs no precomputed row-major accumulators.
- **GPU dispatcher signatures** in `cuTrace_internal.hpp` are kept consistent with the CPU side (the actual GPU implementation lands in #850).

## Tests

- `tests/Tensor_strides_test.cpp` — walks every multi-index across rank-2..5 permutations and dtypes, asserting `flatten(idx · strides()) == offset at()` actually reads (so if `strides()` ever drifts from the layout `at()` indexes through, every case fails).
- `tests/linalg_test/Trace_test.cpp` — strided in-place Trace vs `linalg::Trace(t.contiguous(), ...)` reference across ranks 2–5, complex + integer dtypes, the rank-2 path, output rank invariants (rank-N → rank-(N-2)), and swapped axis order.

## Benchmark

`benchmarks/linalg/Trace_bm.cpp` pits the in-place strided Trace against the "collect contiguously and reduce" baselines:

- **3D**: matvec — `tr(A)[m] = ⟨vec(I_n), vec(A[:, m, :])⟩` as a `{middle, n·n}` @ `{n·n, 1}` BLAS GEMM against `vec(I_n)`.
- **2D**: vecdot — `tr(A) = ⟨vec(I_n), vec(A)⟩` as a BLAS dot product on the `n·n`-element flattenings.
- **2D**: reshape trick — save `A[n-1, n-1]`, view the rest as `{n-1, n+1}`, permute + contiguous so the first column becomes a contiguous row, reduce it, add the saved corner. The `{n-1, n+1}` view reuses the source storage in place (`Storage::resize` is a no-realloc metadata shrink), so the `permute → contiguous` gather is the only copy; the size is restored (and the dropped corner written back) afterwards.

Every variant runs once before timing and is checked against `linalg::Trace`; a wrong baseline `std::abort`s instead of producing fast-but-meaningless numbers.

Numbers below are from `openblas-cpu` (`RelWithDebInfo`) at the current branch tip (`a9d4866`), `--benchmark_min_time=0.3s`, CPU times in µs (smaller is better). **Strided wins at every tested size in every regime** — trace touches `O(n·middle)` diagonal elements while the GEMM baseline touches all `O(n²·middle)`, so BLAS throughput never closes the gap, even in the small-n / large-middle regime where many short dots chain through tuned kernels:

**3D — `{n, middle, n}` traced over (0, 2):**

| n / middle  | Strided Double | Matvec Double (GEMM) | Strided ComplexDouble | Matvec ComplexDouble |
| ----------- | -------------: | -------------------: | --------------------: | -------------------: |
| 64 / 64     | 4.14           | 665                  | 4.61                  | 840                  |
| 256 / 64    | 93.3           | 22360                | 146                   | 32352                |
| 1024 / 16   | 233            | 106312               | 213                   | 207522               |
| 2048 / 16   | 658            | 449672               | 773                   | 827872               |
| 4096 / 8    | 764            | 927594               | 787                   | 1692563              |
| 64 / 2048   | 559            | 41562                | 715                   | 73370                |
| 32 / 8192   | 1425           | 48168                | 1323                  | 75193                |

The last two rows are the BLAS-favorable shapes (small `n`, large `middle`) added at review request: strided still wins by 34–74×.

**2D — `{n, n}` traced over (0, 1):**

| n     | Strided Double | Vecdot Double (dot) | Reshape Double | Strided ComplexDouble | Vecdot ComplexDouble | Reshape ComplexDouble |
| ----- | -------------: | ------------------: | -------------: | --------------------: | -------------------: | --------------------: |
| 64    | 0.340          | 1.68                | 10.7           | 0.322                 | 2.53                 | 11.6                  |
| 256   | 0.440          | 5.79                | 158            | 0.434                 | 10.0                 | 195                   |
| 1024  | 1.06           | 161                 | 3794           | 1.12                  | 211                  | 4971                  |
| 4096  | 35.7           | 1668                | 191056         | 39.9                  | 5235                 | 306758                |
| 8192  | 97.2           | 10847               | 906400         | 104                   | 29265                | 1277788               |

## Test plan

- [x] `openblas-cpu`: full suite **1072/1072** passed.
- [x] Each benchmark variant verified against `linalg::Trace` before timing (the benchmark `abort()`s if any baseline disagrees) across all sizes above.
- [x] `pre-commit` (clang-format-14 + EOF + whitespace) clean.

Draft — opened for review.


benchmarks/linalg/Trace_bm.cpp
```diff
@@ -0,0 +1,301 @@
...
+  }
+
+#define REGISTER_TRACE_3D_BENCHMARK(TypeName, TypeEnum)      \
+  BENCHMARK_TEMPLATE(BM_Trace_Strided_3D_Template, TypeEnum) \
+    ->Name("BM_Trace_Strided_3D_" #TypeName)                 \
+    ->Args({64, 64})                                         \
+    ->Args({256, 64})                                        \
+    ->Args({1024, 16})                                       \
+    ->Args({2048, 16})                                       \
+    ->Args({4096, 8})                                        \
```
IvanaGyro, line 213: I think there should be a case that middle is large and n is normal. In this case Matmul may win.

src/backend/linalg_internal_cpu/Trace_internal.cpp
```diff
@@ -2,179 +2,108 @@
...
 #include <vector>
 
 namespace cytnx {
   namespace linalg_internal {
+    namespace {
+
+      template <class T>
+      Tensor TraceImpl(const Tensor &Tn, cytnx_uint64 a1, cytnx_uint64 a2) {
+        const cytnx_uint64 ax1 = std::min(a1, a2);
+        const cytnx_uint64 ax2 = std::max(a1, a2);
```
IvanaGyro: These two lines are inherited from the old code. They are not needed now.

src/backend/linalg_internal_cpu/Trace_internal.cpp
```diff
@@ -2,179 +2,108 @@
...
+          }
+        }
+        cytnx_uint64 Nelem = 1;
+        for (auto d : out_shape) Nelem *= static_cast<cytnx_uint64>(d);
+        const bool is_2d = out_shape.empty();
+
+        Tensor out = Tensor({is_2d ? cytnx_uint64{1} : Nelem}, Tn.dtype(), Tn.device());
+        out.storage().set_zeros();
+        if (Ndiag == 0 || Nelem == 0) {
+          if (!is_2d) out.reshape_(out_shape);
```
IvanaGyro: Should throw error here instead.

src/backend/linalg_internal_cpu/Trace_internal.cpp
```diff
@@ -2,179 +2,108 @@
...
+
+        const std::vector<cytnx_uint64> strides = Tn.strides();
+        const cytnx_uint64 diag_stride = strides[ax1] + strides[ax2];
+        const cytnx_uint64 extent = (Ndiag - 1) * diag_stride + 1;
+        const T *data = Tn.storage().data<T>();
+
+        if (is_2d) {
+          out.storage().at<T>(0) =
+            PairwiseSum(std::span<const T>(data, extent) | stride(diag_stride));
+          return out;
```
IvanaGyro: We can compose the output Tensor here by Tensor::from_stroage

src/backend/linalg_internal_cpu/Trace_internal.cpp
```diff
@@ -2,179 +2,108 @@
...
+          accu[i - 1] = accu[i] * static_cast<cytnx_uint64>(out_shape[i]);
+
+        T *out_data = out.storage().data<T>();
+        for (cytnx_uint64 i = 0; i < Nelem; ++i) {
+          cytnx_uint64 tmp = i, base = 0;
+          for (cytnx_uint64 x = 0; x < out_shape.size(); ++x) {
+            base += (tmp / accu[x]) * strides[remain_rank_id[x]];
+            tmp %= accu[x];
+          }
+          out_data[i] = PairwiseSum(std::span<const T>(data + base, extent) | stride(diag_stride));
```
IvanaGyro: Do not interfere by the old implementation you should write the most efficient method. Consider this
```
offset += stride[last_dim];
++index[last_dim];

if (index[last_dim] == shape[last_dim]) {
    index[last_dim] = 0;
    offset -= shape[last_dim] * stride[last_dim];
    offset += stride[previous_dim];
    ++index[previous_dim];
}
```
In this way, we don't have to calculate the new strides (accu). And it should be must faster. As the 2d case, we can construct the tensor from storage.

src/backend/linalg_internal_interface.hpp
```diff
@@ -178,10 +178,7 @@ namespace cytnx {
     typedef void (*Sumfunc_oii)(boost::intrusive_ptr<Storage_base> &,
                                 const boost::intrusive_ptr<Storage_base> &, const cytnx_uint64 &);
 
-    typedef void (*Tracefunc_oii)(bool, Tensor &, const Tensor &, cytnx_uint64, cytnx_uint64,
-                                  const std::vector<cytnx_uint64> &,
-                                  const std::vector<cytnx_uint64> &,
-                                  const std::vector<cytnx_int64> &, cytnx_uint64, cytnx_uint64);
+    typedef Tensor (*Tracefunc_oii)(const Tensor &, cytnx_uint64, cytnx_uint64);
```
IvanaGyro, line 181: (Review written by Claude on behalf of @IvanaGyro.)

This PR flips `Tracefunc_oii` to the new `Tensor(const Tensor&, cytnx_uint64, cytnx_uint64)` shape and updates `cuTrace_internal.hpp` to match, but the GPU **definition** in `cuTrace_internal.cu` is still the old `void cuTrace_internal_*(bool, Tensor&, ...)` (that rewrite lands in #850). So a `USE_CUDA=ON` build of *this* branch in isolation won't compile: the `.cu` definitions no longer match their declarations, and `cuTrace_ii[...] = cuTrace_internal_*` assigns an incompatible pointer type into the `Tracefunc_oii` table.

The CPU build is fine and the full stack (through #850) builds + passes — I verified that. But if these PRs are CI'd/merged independently rather than squash-as-a-unit, #862 is a broken-GPU intermediate commit. Worth either landing #850's `.cu` change together with this typedef flip, or calling out the merge-as-a-unit intent so a CUDA bisect doesn't trip here.

tests/linalg_test/Trace_test.cpp
```diff
@@ -0,0 +1,104 @@
...
+    auto p = t.permute({1, 0});
+    ASSERT_FALSE(p.is_contiguous());
+    auto out_p = cytnx::linalg::Trace(p, 0, 1);
+    auto out_c = cytnx::linalg::Trace(t, 0, 1);  // tr(A) == tr(A^T)
+    ASSERT_EQ(out_p.shape().size(), 1u);
+    ASSERT_EQ(out_p.shape()[0], 1u);
+    EXPECT_TRUE(cytnx::TestTools::AreNearlyEqTensor(out_p, out_c, 1e-12));
+  }
+
+  TEST(LinalgTraceTest, OutputRankIsInputMinusTwo) {
```
IvanaGyro, line 90: (Review written by Claude on behalf of @IvanaGyro.)

`TraceImpl` has a dedicated early-out for `Ndiag == 0 || Nelem == 0`, but no test drives it, so the zero-extent path (and the shape of its result) is unverified. A regression there — wrong output shape, or a stride computation that divides/indexes into empty storage — would slip through.

Suggest a case with a zero-size traced axis (e.g. `{0, 0}` -> expect a 1-element result, value 0) and one with a zero-size *remaining* axis (e.g. `{3, 0, 3}` traced over 0,2 -> expect shape `{0}`), asserting both shape and that no read occurs. Cheap, and it locks the `Ndiag==0 || Nelem==0` branch.

IvanaGyro (commented):

Nice rewrite — moving the trace into a single stride-aware `TraceImpl<T>` over the container layer removes the `UniTensor::eye` + `Contract` round-trip and the old 2D path's contiguity assumption (`rawdata[i*Ldim + i]`). I worked through the offset math and it is sound: for any output multi-index, `base + (Ndiag-1)*diag_stride` lands on a valid element, so the `std::span(data+base, extent)` end pointer is at most one-past storage — no OOB. `Tensor::strides()` is the right primitive and `Tensor_strides_test` validates it against `at()` across permutations/dtypes. CPU coverage (permuted ranks 2–5, Double/ComplexDouble/Int32, 2D path, output-rank, swapped axes) is solid and all of it passed on both `openblas-cuda` and `mkl-cuda`. Two things inline: one stacked-build concern and one test gap. Also a minor note: the `Reshape_2D` benchmark mutates the shared storage of a `const` input via `resize()` and relies on shrink-not-reallocating to restore it — it's verified once against `Trace`, but it's a fragile trick to leave in a committed benchmark; a comment-backed assertion that capacity was preserved would harden it.

src/backend/linalg_internal_cpu/Trace_internal.cpp
```diff
@@ -2,179 +2,108 @@
...
+        }
```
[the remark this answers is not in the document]
IvanaGyro: Done in `3ceb1db`. The hot loop is now an odometer exactly as you sketched: precompute `out_strides[x] = strides[remain_rank_id[x]]` once, then per element bump the last axis index (carrying into earlier axes on wrap) and adjust `base` by the affected axes' strides — no division, no modulo, and the `accu` accumulator vector is gone. Walks the output in row-major order at amortized O(1) per element.

```cpp
std::vector<cytnx_uint64> out_strides(out_shape.size());
for (cytnx_uint64 x = 0; x < out_shape.size(); ++x)
  out_strides[x] = strides[remain_rank_id[x]];

std::vector<cytnx_uint64> index(out_shape.size(), 0);
cytnx_uint64 base = 0;
for (cytnx_uint64 i = 0; i < Nelem; ++i) {
  out_data[i] = PairwiseSum(std::span<const T>(data + base, extent) | stride(diag_stride));
  for (cytnx_uint64 x = out_shape.size(); x-- > 0;) {
    if (++index[x] < static_cast<cytnx_uint64>(out_shape[x])) {
      base += out_strides[x];
      break;
    }
    index[x] = 0;
    base -= (static_cast<cytnx_uint64>(out_shape[x]) - 1) * out_strides[x];
  }
}
```

---

benchmarks/linalg/Trace_bm.cpp
IvanaGyro: Not added yet — flagging it. A "BLAS wins when middle is large and n is moderate" case would be a useful complement to the current shapes (which all have small `middle` or focus the work in `n`). Two concerns:

1. The current benchmark already covers `n / middle` = 64/64, 256/64, 1024/16, 2048/16, 4096/8 — strided wins everywhere, with the BLAS matvec scaling as O(n²·middle) (touches the whole tensor) vs strided O(n·middle). For Matmul to win, you'd need a regime where the BLAS throughput more than offsets the n× overcount; that's something like middle ≫ n.
2. Adding e.g. `n=32, middle=8192` would size `vec(I_n) = n²·middle ≈ 8M` and `packed ≈ 8M` — feasible, ~64 MB for double, but the benchmark single-process build already builds `Sum`'s pairwise-sum buffers etc. and the openblas-cpu sanitized binary is slow.

If you'd like, I can add `n=32 / middle=8192` and `n=64 / middle=2048` cases in a follow-up — the question is whether you want the result recorded in the PR description before merging, or it's fine to land as a separate "trace bench: BLAS-favorable shapes" PR. Let me know.

---

src/backend/linalg_internal_cpu/Trace_internal.cpp
IvanaGyro: Done in `3ceb1db`. Both `#include "backend/lapack_wrapper.hpp"` and `#include "utils/utils.hpp"` (left over from the old `Contract`-based body) are gone; the file's only includes now are the ones it actually uses: `Trace_internal.hpp`, `Tensor.hpp`, `backend/Storage.hpp`, `cytnx_error.hpp`, `pairwise_sum.hpp`, `stride_view.hpp`, plus `<algorithm>` / `<span>` / `<vector>`.

---

src/backend/linalg_internal_cpu/Trace_internal.cpp
IvanaGyro: I kept the early-return-zero behavior on `Ndiag == 0 || Nelem == 0` rather than turning it into an error, for two reasons:

1. **`linalg::Trace` already validates upstream** that `shape[ax1] == shape[ax2]` and that the axes are in range. The only way to reach the internal dispatcher with `Ndiag == 0` is if the *common* traced size is genuinely zero (a `{0, m, 0}`-style tensor) — which is a well-defined input, not a programmer error. NumPy treats this the same way: `np.trace(np.zeros((0,0)))` returns `0`, not an exception.
2. The guard exists primarily to prevent `(Ndiag - 1)` from underflowing `cytnx_uint64` into a huge `extent` value (which would OOB the subsequent `std::span` read). That's a safety guard, not a policy choice — promoting it to a hard error would conflate "this would crash" with "this input is rejected".

If you'd prefer the strict-rejection policy, the natural place is to add a `cytnx_error_msg(shape[ax1] == 0, ...)` check in `linalg::Trace` itself (one level up) — happy to do that in a follow-up if you want trace to reject zero-axis tensors as a matter of API contract. Let me know.

---

src/backend/linalg_internal_cpu/Trace_internal.cpp
IvanaGyro: Done in `954e366`. `TraceImpl<T>` now fills a flat result `Storage` directly and composes the `Tensor` via `Tensor::from_storage` once the diagonal sums are written, instead of allocating a `Tensor` upfront and reaching through `out.storage()` to fill it:

```cpp
Storage out_storage(is_2d ? cytnx_uint64{1} : Nelem, Tn.dtype(), Tn.device());
// ... fill out_storage.data<T>() ...
Tensor out = Tensor::from_storage(out_storage);
out.reshape_(out_shape);   // ND path
return out;
```

The 2D path returns `Tensor::from_storage(out_storage)` directly (no reshape needed); the ND path reshapes once at the end. Storage stays at the storage layer until the final return, mixing-of-layers gone.

---

src/backend/linalg_internal_cpu/Trace_internal.cpp
IvanaGyro: Done exactly along your sketch in `3ceb1db`. The hot loop is now:

```cpp
for (cytnx_uint64 i = 0; i < Nelem; ++i) {
  out_data[i] = PairwiseSum(std::span<const T>(data + base, extent) | stride(diag_stride));
  for (cytnx_uint64 x = out_shape.size(); x-- > 0;) {
    if (++index[x] < static_cast<cytnx_uint64>(out_shape[x])) {
      base += out_strides[x];      // last_dim ++; offset += stride[last_dim]
      break;
    }
    index[x] = 0;                  // wrap and carry into previous_dim
    base -= (static_cast<cytnx_uint64>(out_shape[x]) - 1) * out_strides[x];
  }
}
```

`out_strides[x] = strides[remain_rank_id[x]]` is precomputed once (matches your `some_new_strides[x]`), so there's no extra indirection per element. The `accu` row-major accumulator vector is gone entirely — no division, no modulo, no per-element flat-index decode. And the 2D path constructs the result via `Tensor::from_storage` in `954e366` (the next commit on this branch), so the storage is built and then promoted to a Tensor at the return.

---

src/backend/linalg_internal_interface.hpp
IvanaGyro: Resolved by the chain reshape since this review was written. #861 was closed and its signature/typedef changes were folded into #862 and #850 directly (acting on @IvanaGyro's "merge into #862 and #850" call on #861). So `Tracefunc_oii` flips and the `cuTrace_internal.*` kernel rewrite now land in the same conceptual unit; there's no "broken-GPU intermediate" to bisect through anymore.

I also assembled a working `nvcc 12.6` toolchain from the conda `nvidia` channel and verified the rebased GPU TU compiles cleanly to an object (`nvcc -c cuTrace_internal.cu`, zero errors), so the typedef-vs-definition consistency is now compiler-checked, not just argued.

---

tests/linalg_test/Trace_test.cpp
IvanaGyro: Flagging this as still outstanding — happy to add the test if you confirm. The reason I held off: creating a `Tensor` with a zero-sized dimension via `random::random_tensor({0, 0}, ...)` is a path with no other test coverage in this codebase, and I wasn't sure whether the policy is "accept zero-axis inputs and return a zero trace" (NumPy-style — what my early-return implements) or "reject them" (which the maintainer's other inline comment on `Trace_internal.cpp:39` hints at).

If the policy is **accept** (the current behavior): I'll add `ZeroTracedAxisReturnsZero` (`{0, 0}` → `{1}` containing `0`) and `ZeroRemainingAxisReturnsShapedZero` (`{3, 0, 3}` traced over (0,2) → `{0}`) to `LinalgTraceTest`.

If the policy is **reject**: the early-return guard should be lifted into `linalg::Trace`'s upstream validation as a `cytnx_error_msg`, and the test would be `EXPECT_THROW` cases instead.

Which way would you like it?

---

IvanaGyro:
Pushed the rebased branch (`954e366`) addressing this round, including the squash you asked for on #861:

- **#861 folded into #862 and #850** per your "merge into" note — `claude/trace-cpu` now sits directly on `claude/stride-view`, and the `Tracefunc_oii` typedef + dispatcher signature changes land here on the first CPU commit. The intermediate sig-refactor PR is closed.
- **Odometer hot loop** (`3ceb1db`) — `accu`/div/mod removed; precomputed `out_strides[x]` + carry-on-wrap exactly as sketched in the inline thread. Amortized O(1) per output element.
- **`Tensor::from_storage`** (`954e366`) — output composed at the storage layer; both 2D and ND paths return via `Tensor::from_storage(out_storage)`.
- **Unused includes** (`3ceb1db`) — `backend/lapack_wrapper.hpp` and `utils/utils.hpp` (left from the old `Contract`-based body) are gone.

Tests on the rebased tip: `openblas-cpu` **965/965**, `Trace` suite 11/11 (including permuted ranks 2–5, complex+integer dtypes, rank-2 path, output-rank, swapped axes); `Tensor::strides()` covered by `Tensor_strides_test` walking every multi-index across permuted ranks 2–5.

Open inline:
- "BLAS wins when middle is large" benchmark case — flagged for follow-up; can add `n=32/middle=8192` and `n=64/middle=2048` if you want it before merge.
- `Ndiag==0`/`Nelem==0` test — held pending your policy call (NumPy-style return-zero vs strict error). I'll add the test the moment you confirm.
- "Should throw" on zero-axis — replied with the case for keeping return-zero parity with NumPy; reversible if you want the strict-error policy.

---

IvanaGyro:
Pushed `f19ceb6` — companion rename on the CPU side to keep `TraceImpl` aligned with the GPU side after the matching review round on #850:

| was | now |
|---|---|
| `n_diag` | `diagonal_length` |
| `diag_stride` | `diagonal_stride` |
| `extent` | `diagonal_span` |
| `n_elem` | `output_size` |
| `out_rank` | `surviving_rank` |
| `out_shape`, `out_strides` | `output_shape`, `surviving_input_stride` |
| `is_2d` | `output_is_scalar` |
| `shape_in`, `strides` | `input_shape`, `input_strides` |
| `data`, `out_data`, `out_storage` | `input_data`, `output_data`, `output_storage` |
| `base`, `index` | `diagonal_start_offset`, `surviving_index` |
| `i`, `x` (loop counters) | `output_index`, `axis` |

No behavior change. `openblas-cpu` 968/968 from the rebased tip. The CPU and GPU TUs now share the same vocabulary, which made reviewing the two against each other much easier in this last round.

---

IvanaGyro:
Rebased onto the new mdspan-based #860 (no longer has `stride_view`); added one follow-up commit `73880c0`/`9fa5b19` that swaps the trace's strided-view consumer to a Kokkos `mdspan`. The five existing commits and their inline reviews are preserved.

**Diff at the top commit:**

- `pairwise_sum.hpp` generalized: `PairwiseSumBlocks` now drives the reduction through a callable `get(i)` instead of an iterator's `first[i]`. The existing `PairwiseSum(range)` wraps `[first](i){ return first[i]; }`; a new overload accepts a 1D `Kokkos::mdspan` and wraps `[&mds](i){ return mds(i); }`. Both go through the same recursive kernel.
- The recursive split passes a *base offset* through the callable (`PairwiseSumBlocks(get, base, n) → PairwiseSumBlocks(get, base + half, n - half)`) instead of wrapping the callable in a fresh closure at each depth. Wrapping made every recursion level a new template instantiation; for the floating-point dtype matrix in `Sum_internal` / `Trace_internal`, the resulting parameter-pack explosion OOM'd `cc1plus`. The base-offset variant produces one instantiation per `(T, Get)` regardless of depth.
- `TraceImpl` constructs the diagonal as a 1D `mdspan` with `layout_stride` (`extent = diagonal_length`, `stride = strides[ax1] + strides[ax2]`) and calls `PairwiseSum(diagonal)`. The 2D path and the ND-odometer hot loop go through the same call site. `stride_view` is gone from the includes.

`openblas-cpu` **1051/1051** on the rebased tip.

---

benchmarks/linalg/Trace_bm.cpp
IvanaGyro: I need the result be recorded in the PR description.

IvanaGyro: Added the two BLAS-favorable shapes you asked for (`{n=64, middle=2048}` and `{n=32, middle=8192}`) in `a9d4866` and ran them. Strided still wins by 34–74× even in this regime, so the "matvec might catch up at large middle" hypothesis isn't borne out at these sizes. Full numbers folded into the PR description's 3D table.

| n / middle  | Strided Double | Matvec Double (GEMM) | Strided CD | Matvec CD |
| ----------- | -------------: | -------------------: | ---------: | --------: |
| 64 / 2048   | 559 µs         | 41562 µs             | 715 µs     | 73370 µs  |
| 32 / 8192   | 1425 µs        | 48168 µs             | 1323 µs    | 75193 µs  |

---

src/backend/linalg_internal_gpu/cuTrace_internal.hpp
```diff
@@ -12,82 +12,17 @@
...
-                              const std::vector<cytnx_int64> &shape, const cytnx_uint64 &ax1,
-                              const cytnx_uint64 &ax2);
-
-    void cuTrace_internal_b(const bool &is_2d, Tensor &out, const Tensor &Tn,
-                            const cytnx_uint64 &Ndiag, const cytnx_uint64 &Nelem,
-                            const std::vector<cytnx_uint64> &accu,
-                            const std::vector<cytnx_uint64> &remain_rank_id,
-                            const std::vector<cytnx_int64> &shape, const cytnx_uint64 &ax1,
-                            const cytnx_uint64 &ax2);
+    Tensor cuTrace_internal_cd(const Tensor &Tn, cytnx_uint64 ax1, cytnx_uint64 ax2);
```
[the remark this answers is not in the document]
IvanaGyro: Known and intentional. The new `cuTrace_internal_*` definitions matching these declarations land in the stacked PR #850 (`refactor(trace): implement the GPU Tensor trace as CUDA kernels`, currently commit `a8f376f` on `claude/gallant-franklin-4g4rY`). #862 and #850 are meant to land together — same situation called out in the earlier (resolved) thread https://github.com/Cytnx-dev/Cytnx/pull/862#discussion_r3329858803, where I also verified `nvcc -c cuTrace_internal.cu` compiles cleanly on the rebased tip of #850.

So this isn't an oversight, but it does mean #862 in isolation is a broken-GPU intermediate. If we want to avoid that on the merge graph, the options are (a) squash-merge #862+#850 as one unit, or (b) hold #862 until #850 is ready to merge alongside it. No code change on this PR alone resolves it.

---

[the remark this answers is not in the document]
IvanaGyro: Solved, a wrapper is added to recover the GPU branch.

IvanaGyro:
@pcchen May you review this PR again? Or are there someone else can review this PR?

pcchen:
@IvanaGyro yes. I will review again.

pcchen:
## Review-thread status check — are all raised issues fixed?

Verified each of the 9 review threads against the actual code on `claude/trace-cpu`, not just the "resolved" flag. **7 of 9 are genuinely fixed; 1 was declined-by-design; 1 is marked resolved but not actually addressed.**

### Genuinely fixed in code (7)
- **Hot-loop `/` and `%` → odometer** — confirmed: the trace loop carries the start offset on an odometer (`++surviving_index[axis]` with carry); no division/modulo in the loop.
- **Leftover includes** — confirmed removed (`lapack_wrapper.hpp`, `utils/utils.hpp`).
- **Compose output via `Tensor::from_storage`** — confirmed (`return Tensor::from_storage(output_storage);`).
- **Benchmark "large middle, moderate n" case** — confirmed added (`{64,2048}`, `{32,8192}`) with numbers in the description.
- **`Tracefunc_oii` typedef flip / GPU decl mismatch** — resolved via the #861 → #862/#850 chain reshape.
- **GPU build broken** — confirmed: a wrapper in `cuTrace_internal.cu` restores the GPU branch (see caveat below).

### Resolved by *declining* the change (1) — judgment call, not a code change
- **"Should throw error instead"** — the early-return-zero on `diagonal_length == 0 || output_size == 0` (`Trace_internal.cpp:58-62`) was **kept**, with the reasoning that `linalg::Trace` validates upstream and zero-extent is legitimate rather than an error. Defensible, and the reviewer didn't object — but the requested change was intentionally not made.

### Marked resolved but NOT actually fixed (1) ⚠️
- **Zero-extent path has no test.** The thread is resolved, but the last comment on it says *"Flagging this as still outstanding — happy to add the test if you confirm."* Confirmed by inspection: no test in `Trace_test.cpp` uses a zero-sized dimension (shapes are `{4,3,4}`, `{4,3,4,5}`, `{3,4,3,2,3}`, `{6,6}`). So the `diagonal_length == 0 || output_size == 0` early-out remains **completely untested**. This one is resolved-in-name-only — recommend reopening it and adding a zero-extent case (or confirming the policy on zero-sized-dim tensors first).

### Two caveats
1. **The GPU wrapper reintroduces the layering this PR removes.** `cuTrace_internal.cu` builds a `UniTensor::eye` and calls `Contract`, so the low-level GPU trace still depends on `UniTensor`/`Contract` (it includes `UniTensor.hpp`). The #834 goal of removing that inverted dependency is met **only on CPU**; the GPU side keeps the inversion as a stopgap until #850 lands the CUDA kernels. (Disclosed in the description.)
2. PR is **Draft**, `reviewDecision = REVIEW_REQUIRED`, `mergeStateStatus = BLOCKED` — no approving review yet.

**Bottom line:** the correctness/perf items are properly fixed; the one genuine loose end is the **zero-extent test coverage**, which is closed without being addressed.

---

pcchen:
@IvanaGyro can you add test for the "Zero-extent path"?

tests/linalg_test/Trace_test.cpp
IvanaGyro: Decision: accept zero-extent inputs and return NumPy-style results rather than rejecting them as an error, matching `np.trace`'s behavior for the same shapes (verified against NumPy, and PyTorch/JAX construct zero-size tensors the same way). Concretely: a zero-length traced diagonal sums to 0 for every surviving index, and a zero-length surviving axis produces an empty result of that shape.

Test coverage for this has been added to `tests/linalg_test/Trace_test.cpp` (CPU) and `tests/gpu/linalg_test/Trace_test.cpp` (GPU, PR #850), covering the `{0,0}`, `{3,0,0}`, and `{3,0,3}` cases.

---

IvanaGyro:
> @IvanaGyro can you add test for the "Zero-extent path"?

@pcchen Yes, added. And here is the decision about tracing zero-dim tensors: https://github.com/Cytnx-dev/Cytnx/pull/862#discussion_r3501076687

pcchen:
The failling check will be fixed in #960. I will approve and merge this.

pcchen (approved):

PR #866, opened 2026-05-31
ci: unified release_pypi.yml — stable to PyPI on tag, nightly .devN to anaconda.org on master push

IvanaGyro:
## Summary

Single `release_pypi.yml` workflow that builds wheels once with cibuildwheel and routes them by a resolved channel:

- **stable → PyPI** — on a `v*` tag push, publishes `cytnx X.Y.Z` to production PyPI via OIDC trusted publishing.
- **nightly → anaconda.org** — on a push to `master` (every merged PR), stamps a PEP 440 dev version (`MAJOR.MINOR.PATCH.devYYYYMMDDHHMM`, UTC) into pyproject.toml and uploads the wheels to the **`cytnx-nightly-wheels`** organisation channel on anaconda.org.
- **pull request → build only, no publish** — the PR is merged with its target branch and the **nightly** wheels are built (validating the nightly release path end to end), but nothing is uploaded.
- **workflow_dispatch** — a `channel` input (`auto` / `stable` / `nightly`); `auto` infers from `github.ref` (tag → stable, otherwise nightly).

A single `DetermineChannel` job emits two outputs — `channel` (stable/nightly) and `publish` (true/false) — consumed by the build steps and the two mutually-exclusive publish jobs.

The wheels keep the `cytnx` distribution name in both channels. Stable users do nothing new (`pip install cytnx`). Nightly users add the anaconda.org index:

```
pip install --pre \
  --extra-index-url https://pypi.anaconda.org/cytnx-nightly-wheels/simple \
  cytnx
```

`--pre` selects the `.devN` pre-release; `--extra-index-url` (not `--index-url`) lets cytnx's runtime deps (numpy, graphviz, beartype) keep resolving from PyPI.

## Why anaconda.org instead of PyPI for nightlies

The original design uploaded nightlies to PyPI as same-project `.devN` releases (numpy-style: `pip install cytnx` → stable, `pip install --pre cytnx` → nightly). We then applied for hosting on the shared **[scientific-python-nightly-wheels](https://anaconda.org/scientific-python-nightly-wheels) anaconda.org channel** (the channel numpy, scipy, scikit-learn, matplotlib, … use for their nightlies). **That application was rejected** (https://github.com/scientific-python/upload-nightly-action/issues/167), so we self-host the nightly index on our own anaconda.org organisation, **[`cytnx-nightly-wheels`](https://anaconda.org/cytnx-nightly-wheels)**.

This keeps the production PyPI project (`cytnx`) reserved for tagged stable releases only and isolates per-merge dev wheels on a dedicated index. The publish action (`scientific-python/upload-nightly-action`) is the same one the scientific-python ecosystem uses — only the target organisation differs.

## Why one workflow file (and why it builds on PRs)

The stable and nightly paths share the cibuildwheel matrix, ccache configuration, and the PR merge-with-target step; only the publish target differs. Folding both into one file means a change to the build matrix is made in one place, and a same-commit tag push + master merge no longer run two redundant builds.

Pull requests build the (nightly) release wheels but skip publishing, so a PR that breaks the release build — including the nightly version-stamping path — is caught before merge. `DetermineChannel` resolves PRs to `channel=nightly, publish=false`; the `publish == 'true'` clause on both publish jobs keeps PR runs from uploading.

## Commits

1. **`build: refactor optional-deps; add release-tools dependency-group`** — collapse `dev` extras to `cytnx[test] + cytnx[coverage]` (one source of truth per leaf group) and add a PEP 735 `[dependency-groups]` table with a `release-tools` group (`tomlkit`). A dependency-group, not an optional-dependency, so `pip install --group release-tools` installs the helper without going through scikit-build-core (which `pip install .[release-tools]` would, compiling cytnx on the host).
2. **`build: add CYTNX_VERSION_TAG env hook for dev-version suffixes`** — `CMakeLists.txt` gains `CYTNX_VERSION_FULL` = numeric `CYTNX_VERSION` + `$ENV{CYTNX_VERSION_TAG}` when set. Used only for the `CYTNX_VERSION` compile def (i.e. `cytnx.__version__`); numeric-only consumers (`project(VERSION ...)`, SOVERSION, libname) untouched. Also adds `CYTNX_VERSION_TAG` to the cibuildwheel Linux `environment-pass`. No-op when unset.
3. **`build: add nightly-release pyproject stamping helper`** — `tools/prepare_nightly_release.py` uses `tomlkit` to rewrite pyproject.toml (`dynamic = ["version"]` removed, static `version = "X.Y.Z.devYYYYMMDDHHMM"`, `[tool.scikit-build.metadata.version]` removed) and appends `CYTNX_VERSION_TAG` to `$GITHUB_ENV`. Reuses the regex from `[tool.scikit-build.metadata.version].regex`, so the regex lives in one place.
4. **`ci: switch release_pypi.yml from TestPyPI to production PyPI`** — publish destination TestPyPI → production PyPI. PR and master-push runs still build the full matrix as a build-health check but no longer upload anything (previously every PR/master push went to TestPyPI); publishing is gated to `v*` tags + dispatch. Job renamed `ReleaseTestPyPI` → `ReleasePyPI`. PR merge-with-target step retained (with an explicit committer identity for the merge commit). Login-shell `defaults` dropped; two ccache steps collapsed to one; third-party actions SHA-pinned.
5. **`ci: extend release_pypi.yml with nightly anaconda.org path`** — adds the `DetermineChannel` job (`channel` + `publish` outputs), the nightly stamping steps (gated on `channel == 'nightly'`), the `workflow_dispatch.channel` input, and a `PublishNightlyAnaconda` job uploading to anaconda.org via `scientific-python/upload-nightly-action` (SHA-pinned 0.6.4) with `ANACONDA_ORG_UPLOAD_TOKEN`. `ReleasePyPI`'s gate moves from the event-based expression to `channel == 'stable' && publish == 'true'`.

## Version stamping — why timestamp instead of incremental

A monotonic timestamp (`devYYYYMMDDHHMM`) needs no external state. An incremental `devN` would require querying the index for the last `.devN` or storing a counter — both race under parallel merges and break on repo moves. Minute resolution handles multiple merges per day; a same-minute collision is overwritten by `anaconda upload --force` on re-run.

## One-time setup required before merge

- **PyPI (`cytnx` project)** — add a trusted publisher: repository `Cytnx-dev/Cytnx`, workflow `release_pypi.yml`, job `ReleasePyPI`, no environment.
- **anaconda.org (`cytnx-nightly-wheels` org)** — create an API token with upload scope for the org, and store it as the **`ANACONDA_ORG_UPLOAD_TOKEN`** repository secret (Settings → Secrets and variables → Actions). The upload step reads it via `secrets.ANACONDA_ORG_UPLOAD_TOKEN`.

Until these are configured, the respective publish jobs will fail; this is intentional and safe. PR builds need neither secret (they don't publish).

## Test plan

- [x] PyPI trusted-publisher entry created on the `cytnx` project.
- [x] `ANACONDA_ORG_UPLOAD_TOKEN` repository secret created with upload scope for `cytnx-nightly-wheels`.

### DIFF
--- a/.github/workflows/release_pypi.yml
+++ b/.github/workflows/release_pypi.yml
@@ -1,11 +1,37 @@
-name: Release to PyPI
+name: Release wheels
 
-# Production release pipeline for the `cytnx` PyPI project. cibuildwheel
-# builds the wheels on every pull request and every push to
-# master/tags as a build-health check, but the wheels are only
-# published to production PyPI on `v*` tag pushes and manual
-# workflow_dispatch. Pull-request and master-push runs build and
-# discard the wheels.
+# Single release pipeline for cytnx. cibuildwheel builds the wheels on
+# every pull request and every push as a build-health check; what
+# happens to the wheels afterwards depends on the resolved channel:
+#
+#   - **stable** — a `v*` tag push (or a dispatch resolving to stable)
+#     publishes `cytnx X.Y.Z` to production PyPI via OIDC trusted
+#     publishing. pyproject.toml is consumed as-is (numeric version
+#     from version.cmake).
+#
+#   - **nightly** — a push to `master` (or a dispatch resolving to
+#     nightly) stamps a PEP 440 dev version
+#     (`MAJOR.MINOR.PATCH.devYYYYMMDDHHMM`) into pyproject.toml via
+#     tools/prepare_nightly_release.py and uploads the wheels to the
+#     `cytnx-nightly-wheels` organisation channel on anaconda.org. The
+#     official `cytnx` PyPI project is never touched by this path, so
+#     production PyPI only ever holds tagged stable releases.
+#
+# Pull requests always resolve to the nightly channel with publishing
+# disabled: the PR is merged with its target branch and the nightly
+# wheels are built (validating the nightly release path) but never
+# uploaded.
+#
+# Manual `workflow_dispatch` runs default to auto-detection (tag →
+# stable, anything else → nightly) and expose a `channel` input that
+# can force either path.
+#
+# Install commands:
+#   stable:   pip install cytnx
+#   nightly:  pip install --pre \
+#               --extra-index-url \
+#                 https://pypi.anaconda.org/cytnx-nightly-wheels/simple \
+#               cytnx
 
 on:
   pull_request:
@@ -15,10 +41,78 @@ on:
     tags:
       - "v*"
   workflow_dispatch:
+    inputs:
+      channel:
+        description: "Release channel"
+        type: choice
+        options:
+          - auto
+          - stable
+          - nightly
+        default: auto
 
 jobs:
+  DetermineChannel:
+    # Resolves a single source of truth for the rest of the workflow:
+    #   * channel — `stable` or `nightly`, selecting the build/publish path
+    #   * publish — whether the resolved channel's publish job may run
+    #   * dev_tag — the PEP 440 dev suffix (`.devYYYYMMDDHHMM`, UTC)
+    #     for nightly builds; empty string for stable. Computed in
+    #     this single job and forwarded to every BuildWheel matrix
+    #     entry so all wheels for one workflow run carry the same
+    #     version, even when the matrix runners start across a
+    #     minute boundary.
+    # Concentrating the decision here keeps the BuildWheel matrix
+    # and the publish jobs from re-deriving the trigger logic.
+    name: DetermineChannel
+    runs-on: ubuntu-24.04
+    outputs:
+      channel: ${{ steps.pick.outputs.channel }}
+      publish: ${{ steps.pick.outputs.publish }}
+      dev_tag: ${{ steps.pick.outputs.dev_tag }}
+    steps:
+    - id: pick
+      run: |
+        event="${{ github.event_name }}"
+        ref="${{ github.ref }}"
+        if [[ "$event" == "pull_request" ]]; then
+          # PRs validate the nightly build path but never upload.
+          channel=nightly
+          publish=false
+        elif [[ "$event" == "push" ]]; then
+          if [[ "$ref" == refs/tags/v* ]]; then
+            channel=stable
+          else
+            channel=nightly
+          fi
+          publish=true
+        else
+          # workflow_dispatch
+          requested="${{ inputs.channel }}"
+          if [[ "$requested" == "auto" ]]; then
+            if [[ "$ref" == refs/tags/v* ]]; then
+              channel=stable
+            else
+              channel=nightly
+            fi
+          else
+            channel="$requested"
+          fi
+          publish=true
+        fi
+        if [[ "$channel" == "nightly" ]]; then
+          dev_tag=".dev$(date -u +%Y%m%d%H%M)"
+        else
+          dev_tag=""
+        fi
+        echo "channel=$channel" >> "$GITHUB_OUTPUT"
+        echo "publish=$publish" >> "$GITHUB_OUTPUT"
+        echo "dev_tag=$dev_tag" >> "$GITHUB_OUTPUT"
+        echo "event=$event ref=$ref -> channel=$channel publish=$publish dev_tag=$dev_tag"
+
   BuildWheel:
     name: BuildWheel-${{ matrix.os }}
+    needs: DetermineChannel
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
@@ -49,6 +143,36 @@ jobs:
         # Refresh submodules in case the merge moved any gitlinks
         # (e.g. cmake_modules/morse_cmake).
         git submodule update --init --recursive
+
+    - name: Install nightly stamping helper
+      if: needs.DetermineChannel.outputs.channel == 'nightly'
+      # Installs the `release-tools` PEP 735 dependency-group from
+      # pyproject.toml (currently just `tomlkit`). Using `--group`
+      # rather than `pip install .[release-tools]` keeps the source
+      # of truth for build-pipeline deps in pyproject.toml without
+      # triggering scikit-build-core to compile cytnx itself before
+      # cibuildwheel does so under its own isolated environment.
+      # `--group` requires pip 25.1+; the runner-bundled pip is
+      # upgraded first to make the requirement portable across
+      # runner image versions.
+      run: |
+        python -m pip install --upgrade pip
+        pip install --group release-tools
+
+    - name: Stamp pyproject.toml for nightly release
+      if: needs.DetermineChannel.outputs.channel == 'nightly'
+      # Rewrites pyproject.toml to set a static
+      # MAJOR.MINOR.PATCH.devYYYYMMDDHHMM version. The dev suffix is
+      # supplied through the CYTNX_VERSION_TAG env var, sourced from
+      # the DetermineChannel job's output so every matrix entry uses
+      # the same value (each matrix runner reaching this step at a
+      # different minute would otherwise stamp a different version
+      # for the same source commit). The stable path skips this step
+      # entirely so production wheels keep the numeric version
+      # straight from version.cmake.
+      env:
+        CYTNX_VERSION_TAG: ${{ needs.DetermineChannel.outputs.dev_tag }}
+      run: python tools/prepare_nightly_release.py
 
     - name: Cache Ccache Directory
       uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830  # v4.3.0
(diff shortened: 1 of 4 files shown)


.github/workflows/release_pypi.yml
```diff
@@ -99,10 +79,10 @@ jobs:
         name: cibw-wheels-${{ matrix.os }}-${{ strategy.job-index }}
         path: ./wheelhouse/*.whl
 
-  ReleaseTestPyPI:
-    name: ReleaseWheel-TestPyPI
+  ReleasePyPI:
+    name: ReleasePyPI
     needs: BuildWheel
-    if: ${{ github.event_name == 'pull_request' || (github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')) }}
+    if: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v') }}
```
IvanaGyro: manually run should also release the package.

tools/prepare_nightly_release.py
```diff
@@ -0,0 +1,81 @@
...
+        sys.exit('expected `name = "cytnx"` in pyproject.toml')
+    text = text.replace('name = "cytnx"', 'name = "cytnx-nightly"', 1)
+
+    if 'dynamic = ["version"]' not in text:
+        sys.exit('expected `dynamic = ["version"]` in pyproject.toml')
+    text = text.replace(
+        'dynamic = ["version"]',
+        f'version = "{version}"',
+        1,
+    )
```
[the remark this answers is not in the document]
IvanaGyro: outdated

tools/prepare_nightly_release.py
```diff
@@ -0,0 +1,81 @@
...
+    r"set\(\w*?VERSION_PATCH\s+(?P<patch>\d+)\)"
+)
+
+
+def read_base_version() -> str:
+    text = VERSION_CMAKE.read_text()
+    match = VERSION_RE.search(text)
+    if not match:
+        sys.exit(f"could not parse MAJOR/MINOR/PATCH from {VERSION_CMAKE}")
+    return f"{match['major']}.{match['minor']}.{match['patch']}"
```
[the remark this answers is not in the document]
IvanaGyro: outdated

.github/workflows/release_pypi.yml
```diff
@@ -94,28 +71,26 @@ jobs:
...
         name: cibw-wheels-${{ matrix.os }}-${{ strategy.job-index }}
         path: ./wheelhouse/*.whl
 
-  ReleaseTestPyPI:
-    name: ReleaseWheel-TestPyPI
+  ReleasePyPI:
+    name: ReleasePyPI
     needs: BuildWheel
-    if: ${{ github.event_name == 'pull_request' || (github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')) }}
+    if: ${{ (github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')) || github.event_name == 'workflow_dispatch' }}
```
IvanaGyro: Just remove this if-clause

.github/workflows/release_pypi_nightly.yml
```diff
@@ -0,0 +1,105 @@
...
+      with:
+        fetch-depth: 0
+        submodules: recursive
+
+    - name: Install nightly stamping helper
+      # Pulls only `tomlkit`, declared in pyproject.toml's
+      # `release-tools` optional-dependencies group. Installed directly
+      # by name (rather than via `pip install .[release-tools]`) so we
+      # avoid triggering the C++ build before cibuildwheel runs.
+      run: pip install tomlkit
```
IvanaGyro: Why not use the section name: # job.
release-tools

tools/prepare_nightly_release.py
```diff
@@ -0,0 +1,81 @@
...
+        f'version = "{version}"',
+        1,
+    )
+
+    # [project].version is now static, so the scikit-build regex provider
+    # block becomes a conflicting declaration. Strip it.
+    block = re.compile(
+        r"\n\[tool\.scikit-build\.metadata\.version\][^\[]*",
+        re.DOTALL,
+    )
```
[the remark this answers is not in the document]
IvanaGyro: Obsoleted by the rewrite. `tools/prepare_nightly_release.py` no longer constructs the `[^\[]*` regex (or any other ad-hoc TOML regex) — it now parses pyproject.toml with `tomlkit` and removes `[tool.scikit-build.metadata.version]` with `del skb_metadata["version"]`. A bracket appearing inside a section value can't break the structural delete.

---

.github/workflows/release_pypi.yml
IvanaGyro: Done in a37179c. The job-level `if:` restated the workflow-level triggers exactly, so removing it has no behavioural effect.

---

.github/workflows/release_pypi_nightly.yml
IvanaGyro: Good catch — fixed in 1c0abab.

The reason the previous version installed `tomlkit` by name rather than by section was that `pip install .[release-tools]` goes through the project build backend (scikit-build-core) and would compile cytnx on the runner before installing the listed deps, which we want cibuildwheel to do instead under its own isolated env. So `release-tools` under `[project.optional-dependencies]` was unusable from this workflow despite being defined.

The proper fix is to move it from `[project.optional-dependencies]` to PEP 735 `[dependency-groups]`. Groups are installable via `pip install --group release-tools` without invoking the build backend, and they are deliberately excluded from `pip install .` resolution — which is exactly the semantics we want for "build-pipeline tooling that isn't a runtime or install-time dep of cytnx". The workflow now does:

```yaml
run: |
  python -m pip install --upgrade pip   # --group needs pip 25.1+
  pip install --group release-tools
```

so the section name in pyproject.toml is now the actual source of truth, as you suggested.

---

IvanaGyro:
I plan to follow [SPEC 4](https://scientific-python.org/specs/spec-0004/) to release the nightly build. This is adopted by numpy, scipy and some scientific Python packages. I am requesting access of [Scientific Python Nightly Wheels](https://anaconda.org/scientific-python-nightly-wheels/) to release the nightly build on https://github.com/scientific-python/upload-nightly-action/issues/167. We will not merge this PR until we get access.

IvanaGyro:
> I plan to follow [SPEC 4](https://scientific-python.org/specs/spec-0004/) to release the nightly build. This is adopted by numpy, scipy and some scientific Python packages. I am requesting access of [Scientific Python Nightly Wheels](https://anaconda.org/scientific-python-nightly-wheels/) to release the nightly build on [scientific-python/upload-nightly-action#167](https://github.com/scientific-python/upload-nightly-action/issues/167). We will not merge this PR until we get access.

They rejected our application, so I decide to publish to a new organization.

tools/prepare_nightly_release.py
```diff
@@ -0,0 +1,108 @@
...
+def read_base_version(version_re: re.Pattern[str]) -> str:
+    text = VERSION_CMAKE.read_text()
+    match = version_re.search(text)
+    if not match:
+        sys.exit(f"could not parse MAJOR/MINOR/PATCH from {VERSION_CMAKE}")
+    return f"{match['major']}.{match['minor']}.{match['patch']}"
+
+
+def build_dev_tag() -> str:
+    stamp = datetime.datetime.now(datetime.timezone.utc).strftime("%Y%m%d%H%M")
```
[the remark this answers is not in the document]
IvanaGyro: Good catch — fixed by computing the dev tag once and propagating it to every matrix entry.

Concretely:

- `DetermineChannel` gains a third output, `dev_tag`. For nightly resolutions it is set once via `.dev$(date -u +%Y%m%d%H%M)`; for stable it is the empty string. Since `DetermineChannel` runs in a single job before `BuildWheel` fans out, all four matrix runners now share one timestamp regardless of when they actually start.
- The `Stamp pyproject.toml for nightly release` step gets a step-level `env: CYTNX_VERSION_TAG: ${{ needs.DetermineChannel.outputs.dev_tag }}`. The prep script now reads `CYTNX_VERSION_TAG` from the environment when set and falls back to `datetime.now(UTC)` only for standalone / local runs, so the env-injected value wins in CI.
- The `Build Wheels` step's `CYTNX_VERSION_TAG` env source is switched from `env.CYTNX_VERSION_TAG` (which used to come from the prep script's `$GITHUB_ENV` write) to `needs.DetermineChannel.outputs.dev_tag` directly, so the cibuildwheel build sees the same value through the same source of truth, not a relayed copy.

Result: one workflow run produces one `cytnx X.Y.Z.devN` version across all wheels, even if the macOS-15-intel runner starts a minute after the manylinux runner.

Per-commit landing (kept by logical change as before):

- Script change (read `CYTNX_VERSION_TAG` from env, fall back to `now()`) folded into 21d23b9.
- Workflow change (`dev_tag` output + step env on the two consumers) folded into 5dee678.

---

pcchen:
## Issue Status Check

### Gemini inline comments (all on `tools/prepare_nightly_release.py`)

| Issue | Status |
|---|---|
| `VERSION_RE` regex fragile against reordering in `version.cmake` | ✅ Obsoleted — script rewritten to use `tomlkit` for all TOML operations; `version.cmake` parsing replaced |
| String replacement fragile against formatting changes | ✅ Obsoleted — same rewrite eliminated all ad-hoc string replacements |
| `[^\[]*` regex fragile if section values contain brackets | ✅ Fixed — `tomlkit` structural delete (`del skb_metadata["version"]`) replaces the regex; a bracket inside a value cannot break it |

### IvanaGyro self-review comments

| Issue | Status |
|---|---|
| `workflow_dispatch` should also be able to release | ✅ Addressed — PR adds `workflow_dispatch` trigger with `channel` input (`auto`/`stable`/`nightly`); `auto` infers stable from `v*` tag, otherwise nightly |
| Remove redundant job-level `if:` clause | ✅ Fixed in `a37179c` |
| Install `tomlkit` via `[dependency-groups]` section name, not by raw package name | ✅ Fixed in `1c0abab` — `release-tools` moved to PEP 735 `[dependency-groups]` table; workflow installs with `pip install --group release-tools` |

### Codex

| Issue | Status |
|---|---|
| P2: Per-matrix-job `now()` call gives different timestamps to different OS runners | ✅ Fixed — `DetermineChannel` now computes `dev_tag` once (`.dev$(date -u +%Y%m%d%H%M)`) and passes it as a step-level env var to all matrix `BuildWheel` runners; all jobs share one timestamp |

### CI

All substantive checks pass. The single failure (`BuildWheel-ubuntu-24.04`) is the pre-existing ccache validation false positive — zero cache hits on a cold cache; wheel builds successfully. Not a code issue.

**All issues resolved. This PR is approved for merge.**

pcchen (approved):

Issue #868, opened 2026-05-31
Allow to print output to any stream


manuschneider:
Currently, many output functions print directly to `std::cout`. In some Python bindings, this is output redirected, but there is no common way to implement `print()`. Output to a different stream/buffer/file is not possible. This was already addressed in #790, but I would like to separate it from the other cleanups in that PR.

This suggestion makes `std::cout << obj` on C++ and `print(obj)` on Python the only source of truth for standard printing (currently, some separate `.print()` methods exist.

Suggestion:
1) Move all serialization into `operator<<` (calls a virtual `print`  on the impl where there is inheritance).
2) Fix every `__repr__` to `ostringstream ss; ss << self; return ss.str();` — that fixes `repr(obj) == ''` everywhere and makes `print(t, file=X)` work natively.
3) Deprecate the redundant methods `.print()` (where not an impl)
4) Make other output methods (`print_info`, `print_diagram`, `print_blocks`, `print_block`) to allow for any `ostream` (`ostream& file = std::cout`).
5) Expose the same API as in 4. in Python, allow for any stream in the `print_...` functions.

An example for the pybind could be:
```
.def("print_diagram",
     [](const UniTensor& self, bool bond_info, py::object file) {
         std::ostringstream ss;
         self.print_diagram(ss, bond_info);
         if (file.is_none())
             py::print(ss.str(), "end"_a = "");
         else
             file.attr("write")(ss.str());
     },
     "bond_info"_a = false, "stream"_a = py::none())
```
This allows for Python users to do:
```
ut.print_diagram(bond_info=True)                       # stdout
ut.print_diagram(bond_info=True, file=sys.stderr)      # stderr
buf = io.StringIO(); ut.print_diagram(file=buf); s = buf.getvalue()
```
C++ users can use any `ostream`.

Notice:
The name `ostream &os` is more common on the C++ side, but I would use `ostream &file` here instead to keep it in sync with the Python API (such that Python users can transfer the names from the C++ API doc directly). On the Python side, `file` is standard, e.g., `print(obj, file=sys.stderr)`.

IvanaGyro:
Python API (`__repr__`) just need a string, so on C++ side we just need a function composing output string and let pybind11 `__repr__` and `operator<<` call that function. In this suggestion, `std::ostringstream` is even not needed.

My suggestion is to remove all `print*()` function in C++. If someone need to show bonds, just use 
```cpp
for(const Bond& bond: bonds) {
[...]

PR #869, opened 2026-05-31
print functions can output to any stream

manuschneider:
Separates the print part in #790 and implements #868.

### DIFF
--- a/include/Gncon.hpp
+++ b/include/Gncon.hpp
@@ -93,7 +93,7 @@ namespace cytnx {
     virtual void clear();
     virtual std::string getOptimalOrder();
     virtual UniTensor Launch(const bool &optimal = false, const std::string &contract_order = "");
-    virtual void PrintNet(std::ostream &os);
+    virtual void PrintNet(std::ostream &file = std::cout) const;
     virtual boost::intrusive_ptr<Gncon_base> clone();
     virtual void Savefile(const std::string &fname);
     virtual ~Gncon_base(){};
@@ -137,7 +137,7 @@ namespace cytnx {
       boost::intrusive_ptr<Gncon_base> out(tmp);
       return out;
     }
-    void PrintNet(std::ostream &os);
+    void PrintNet(std::ostream &file = std::cout) const;
     void Savefile(const std::string &fname);
     ~RegularGncon(){};
   };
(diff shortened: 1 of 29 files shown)


pcchen:
## Code Review

**Verdict: the design is right and we'd like it in — but the diff itself is no longer rebasable; requesting a staged redo on current master.**

### The design is good (and worth adopting as the repo pattern)
- **C++ side:** `print_info` / `print_diagram` / `print_blocks` / `PrintNet(..., std::ostream &file = std::cout)` — defaulted, so zero behavior change for existing callers, while making print output **testable and redirectable** (exactly #868's ask). ✅
- **The pybind side is the best part:** render into `std::ostringstream`, then route to Python — `py::print(...)` or `file.attr("write")(...)` honoring a `file=` kwarg. This is strictly better than the `scoped_ostream_redirect` idiom used today: it works with *any* Python file object, keeps Python-level `print()` semantics, and sidesteps the redirect/GIL interplay entirely (the exact composition trap documented in the #987 guard-discipline note). Worth making this the documented idiom going forward. ✅
- Deprecating `Scalar.print()` toward `print(scalar, file=...)` with a `FutureWarning` is tasteful. ✅

### The practical problem: this May-31 diff patches code that no longer exists
Since it was written, the 2026-07 refactor campaign rewrote most of its 29 files wholesale:
- `src/Symmetry.cpp` / `include/Symmetry.hpp` → replaced by the variant refactor (**#1010**, merged) — the 32+/32− here patch deleted code;
- `include/backend/Scalar.hpp` → being replaced by **#1011** (in flight; `print` now lives behind a `std::visit`);
- `pybind/unitensor_py.cpp` and friends → reworked by **#986/#991/#997/#987**;
- `Block*/DenseUniTensor.cpp` → **#998/#1005**; `storage_py.cpp` → changing again in **#1015**.

So the `CONFLICTING` state isn't a merge-button problem — a rebase is effectively a reimplementation. The good news: it's a *smaller* reimplementation now, because the refactors centralized printing (Symmetry's `print_info` is a single function on the variant; Scalar's print is one visit) — threading `std::ostream&` through is easier post-campaign than pre-.

### Suggested path
1. **Redo on current master, in stages** — start with the classes *not* under active rewrite (UniTensor / Network / Bond / Storage diagram+info printing), which is most of the user value; do Scalar last, after #1011 lands (trivial there: one visit function).
2. Keep the pybind `ostringstream → py::print / file.write` pattern exactly as designed here.
3. **Add tests** — the PR currently adds none, yet stream-parameterization is precisely what makes print output assertable (pin e.g. `print_diagram` output for a known tensor; #1010's pinned-format discipline is the model).
4. Coordinate with #790 (the parent this was split from) so the two don't drift.

To be explicit so the effort isn't wasted: the *direction* is accepted — #868 is worth implementing, and this PR's API/pybind shape is the one we want. It's only the mechanical base that has moved out from under it.

PR #885, opened 2026-06-12
[codex] fix Storage::Fromfile count handling

### LINKED ISSUE #880: Storage::Fromfile uses uninitialized Nelem when count is provided
Found during a static repo review; this needs runtime confirmation, but the code path looks unsafe.

## Suspected bug

`Storage::Fromfile(..., count >= 0)` compares `count > Nelem` before `Nelem` has been assigned:

- `src/backend/Storage.cpp:197-199` declares `cytnx_uint64 Nelem;`
- `src/backend/Storage.cpp:217-222` checks `count > Nelem` in the `count >= 0` branch
- `Nelem = Nbytes / Type.typeSize(dtype)` is only assigned in the `count < 0` branch

That makes the validation nondeterministic: valid reads may be rejected, and oversized reads may be accepted depending on stack contents.

## User-visible paths

- C++: `cytnx::Storage::Fromfile(fname, dtype, count)`
- C++: `cytnx::Tensor::Fromfile(fname, dtype, count)`, which delegates to `Storage::Fromfile`
- Python: `cytnx.Storage.Fromfile(fname, dtype, count)` via `pybind/storage_py.cpp:272-275`

## Expected behavior

Compute the total number of elements first, then validate `count` against it, for example:

```cpp
const cytnx_uint64 file_elems = Nbytes / Type.typeSize(dtype);
if (count < 0) {
  Nelem = file_elems;
} else {
  cytnx_error_msg(static_cast<cytnx_uint64>(count) > file_elems, ...);
  Nelem = count;
}
```

## Suggested regression tests

- Write a small raw binary file with `Storage::Tofile`.
- Verify `Storage::Fromfile(path, Type.Double, 2)` consistently reads two elements.
- Verify `Storage::Fromfile(path, Type.Double, file_elems + 1)` throws.
- Add the same coverage through `Tensor::Fromfile` if practical.

ianmccul:
Fixes #880.

`Storage::Fromfile` computed the file byte size but only initialized `Nelem` in the `count < 0` path. When callers passed an explicit positive `count`, the bounds check compared against an uninitialized `Nelem`, so valid counted reads could throw and oversized counts were undefined behavior.

This change computes the total element count before branching, compares explicit counts against that initialized value, and then uses the requested count for the raw load. It also updates the count-exceeded diagnostic to print the 64-bit element count with a matching format.

Regression coverage added:
- `StorageTest.FromfileHonorsCount` verifies an explicit count reads only the requested prefix of raw `Tofile` data.
- `StorageTest.FromfileRejectsCountLargerThanFile` verifies oversized counts throw deterministically.

Validation performed locally:
- Initialized required master-branch submodules for configure.
- Configured `/tmp/cytnx-storage-fromfile-count-build` with `RUN_TESTS=ON`, `BUILD_PYTHON=OFF`, `USE_HPTT=OFF`, `USE_MKL=OFF`, and explicit OpenBLAS + LAPACKE libraries.
- `cmake --build /tmp/cytnx-storage-fromfile-count-build --target test_main -j2`
- `/usr/bin/ctest --test-dir /tmp/cytnx-storage-fromfile-count-build -R "StorageTest\\.Fromfile" --output-on-failure`
- `/usr/bin/ctest --test-dir /tmp/cytnx-storage-fromfile-count-build -R "^StorageTest\\." --output-on-failure`
- `git diff --check`

### DIFF
--- a/src/backend/Storage.cpp
+++ b/src/backend/Storage.cpp
@@ -217,7 +217,8 @@ namespace cytnx {
     } else {
       const cytnx_uint64 requested_count = static_cast<cytnx_uint64>(count);
       cytnx_error_msg(requested_count > total_elements,
-                      "[ERROR] count exceed the total # of elements %llu in file.\n",
+                      "[ERROR] count (%llu) exceeds the total # of elements (%llu) in file.\n",
+                      static_cast<unsigned long long>(requested_count),
                       static_cast<unsigned long long>(total_elements));
       Nelem = requested_count;
     }
--- a/src/backend/Storage.cpp
+++ b/src/backend/Storage.cpp
@@ -209,13 +209,17 @@ namespace cytnx {
     cytnx_error_msg(Nbytes % Type.typeSize(dtype),
                     "[ERROR] the total size of file is not an interval of assigned dtype.%s", "\n");
 
+    const cytnx_uint64 total_elements = Nbytes / Type.typeSize(dtype);
+
     // check count smaller than Nelem:
-    if (count < 0)
-      Nelem = Nbytes / Type.typeSize(dtype);
-    else {
-      cytnx_error_msg(count > Nelem, "[ERROR] count exceed the total # of elements %d in file.\n",
-                      Nelem);
-      Nelem = count;
+    if (count < 0) {
+      Nelem = total_elements;
+    } else {
+      const cytnx_uint64 requested_count = static_cast<cytnx_uint64>(count);
+      cytnx_error_msg(requested_count > total_elements,
+                      "[ERROR] count exceed the total # of elements %llu in file.\n",
+                      static_cast<unsigned long long>(total_elements));
+      Nelem = requested_count;
     }
 
     f.open(fname, std::ios::in | std::ios::binary);
--- a/tests/Storage_test.cpp
+++ b/tests/Storage_test.cpp
@@ -275,3 +275,33 @@ TEST_F(StorageTest, FromfileRejectsCountLargerThanFile) {
 
   EXPECT_THROW(Storage::Fromfile(path, Type.Double, 5), std::logic_error);
 }
+
+TEST_F(StorageTest, FromfileCountZeroReturnsEmptyStorage) {
+  const std::string path = ::testing::TempDir() + "cytnx_storage_fromfile_count_zero.bin";
+  std::remove(path.c_str());
+  const RemoveFileOnExit cleanup(path);
+
+  Storage source = Storage::from_vector(std::vector<cytnx_double>{1.0, 2.0, 3.0, 4.0});
+  source.Tofile(path);
+
+  Storage loaded = Storage::Fromfile(path, Type.Double, 0);
+  EXPECT_EQ(loaded.dtype(), Type.Double);
+  EXPECT_EQ(loaded.size(), 0);
+}
+
+TEST_F(StorageTest, FromfileCountEqualsTotalElementsReadsAll) {
+  const std::string path = ::testing::TempDir() + "cytnx_storage_fromfile_count_exact.bin";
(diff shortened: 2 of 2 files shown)


ianmccul:
The failing test in CI is unrelated to this PR, and the test itself is addressed in #886

pcchen:
## PR #885 Review: Fix `Storage::Fromfile` count handling

### Overview

Fixes a genuine bug in `Storage::Fromfile`: when `count >= 0`, `Nelem` was never initialized before the comparison `count > Nelem`, making the bounds check undefined behaviour and valid counted reads potentially spurious-throw. The fix computes `total_elements` once before the branch and uses it consistently in both paths.

---

### Correctness

**The bug fix is correct.** Computing `total_elements = Nbytes / Type.typeSize(dtype)` before the branch eliminates the UB. Both paths are now properly defined:

- `count < 0`: `Nelem = total_elements` (unchanged semantics)
- `count >= 0`: compare `requested_count` against `total_elements`, then `Nelem = requested_count`

**Cast safety:** `static_cast<cytnx_uint64>(count)` is safe — the `else` branch is only reached when `count >= 0`, so the signed→unsigned cast is well-defined.

**Format string:** Changing `%d` (which was wrong for a `cytnx_uint64`) to `%llu` with explicit `unsigned long long` casts is correct. The old code was also a latent format-string bug.

---

### Issues

#### 1. `cytnx_error_msg` format: confirm the macro accepts two `%llu` arguments

The old call had one variadic arg; the new call has two:

```cpp
cytnx_error_msg(requested_count > total_elements,
                "[ERROR] count (%llu) exceeds ... (%llu) in file.\n",
                static_cast<unsigned long long>(requested_count),
                static_cast<unsigned long long>(total_elements));
```

`cytnx_error_msg` is variadic, so this should work — but worth confirming that the macro's internal `sprintf`/`snprintf` call passes `__VA_ARGS__` rather than a fixed number of arguments. If it's defined with a single fixed slot, the second argument would be silently dropped.

#### 2. Missing boundary-case tests (minor)

The two new tests cover `count == 2` (prefix) and `count == 5` (overflow). Two additional cases would give complete coverage:

- `count == 0` — empty read; should produce a zero-size Storage
- `count == total_elements` — exact full read with explicit count; should succeed and match the `count < 0` path

Neither is a bug, but `count == 0` in particular exercises a path (`Nelem = 0`) that could interact with downstream allocators.

---

### Test quality

- `RemoveFileOnExit` RAII guard is well-written — cleanup runs even if the test throws. ✅
- `::testing::TempDir()` is the correct way to get a test-scoped temp directory. ✅
- Pre-call `std::remove` before writing is defensive and harmless. ✅
- `ASSERT_EQ(loaded.size(), 2)` before the element checks correctly aborts the test on a size mismatch. ✅

---

### Recommendation

**Approve after confirming** the `cytnx_error_msg` macro passes all variadic arguments to its internal format call (issue 1). The boundary-case tests (issue 2) are optional but encouraged.

pcchen:
**Follow-up on Issue 1 — `cytnx_error_msg` with two `%llu` arguments: confirmed safe.**

The macro definition in `include/cytnx_error.hpp`:

```cpp
#define cytnx_error_msg(is_true, format, ...)                                               \
  {                                                                                         \
    if (is_true)                                                                            \
      error_msg(__PRETTY_FUNCTION__, __FILE__, __LINE__, (is_true), (format), __VA_ARGS__); \
  }
```

passes `__VA_ARGS__` to `error_msg`, which uses `vsprintf(msg, format, args)` internally — a proper `va_list`-based call that consumes all variadic arguments in order. Both `requested_count` and `total_elements` will be correctly formatted.

Issue 1 is a non-issue. No code change needed. The PR is **approved**.

pcchen (approved):

PR #889, opened 2026-06-13
Fix TDVP Lanczos_Exp instability

### LINKED ISSUE #888: TDVP unit test fails to converge
The fix to the TDVP unit test in #887 is causing CI failures on main now, eg:

https://github.com/Cytnx-dev/Cytnx/actions/runs/27460385738/job/81172912323

Since that was my PR, I'll attempt to debug it.  What seems to be happening is that the energy converges, but then once it gets close to the groundstate it moves away again:
```text
time energy
1 energy: -1.803664846184731e-17
2 energy: (-7.712220640606524-5.623000532415372e-27j)
3 energy: (-7.994634396420157+9.26442286059391e-23j)
4 energy: (-7.999901693163588+1.3234889800848443e-23j)
5 energy: (-7.999998199238601+6.617444900424222e-24j)
6 energy: (-7.999999950947473-1.3234889800848443e-23j)
7 energy: (-7.999999973859883-1.9852334701272664e-23j)
8 energy: (-7.978593543194767+4.607859233063394e-19j)
9 energy: (-7.9996145770681295+2.117582368135751e-21j)
10 energy: (-7.999992055512535-9.26442286059391e-23j)
```

This appears to be sporadic, at least it worked fine locally and on previous CI runs.  A temporary fix would be to reduce the number of timesteps in the test from 10 to 6.

ianmccul:
Fixes #888.

## Summary

This fixes the sporadic failure of the deterministic TDVP pytest without relaxing the test tolerance or reducing the number of time steps.

The failure exposed two separate issues in the TDVP/Lanczos_Exp path:

1. The TDVP one-site effective operator declared the wrong `LinOp::nx()` dimension.
2. `Lanczos_Exp` attempted to recover from Krylov breakdown by injecting a random Gram-Schmidt vector, which is not a valid continuation of the Lanczos recurrence.

Together these could make the TDVP run nondeterministic, return a NaN from `Lanczos_Exp`, or produce a final energy far from the expected ground-state value.

## Details

In `example/TDVP/tdvp1_dense.py`, `OneSiteOp` used

```python
d = L.shape()[0]
```

when constructing the `LinOp` dimension. Here `L` is the left environment with shape `[W_left, D_left, D_left]`, so `L.shape()[0]` is the MPO auxiliary dimension. It is not the physical dimension of the local site.

The one-site vector acted on by the effective Hamiltonian has shape

```text
psi[D_left, d_phys, D_right]
```

so the `LinOp` dimension should be

```text
D_left * d_phys * D_right
```

with `d_phys = M.shape()[2]`. In the failing TDVP test this meant `LinOp::nx()` was reported as 10 or 20 while the true local vector dimension was 4 or 8.

That incorrect dimension made `Lanczos_Exp` believe it could build a larger Krylov space than the local tensor space actually allowed.

The second problem was in the `beta <= beta_tol` path of `Lanczos_Exp`. When the Lanczos residual norm became too small, the code generated an arbitrary random vector orthogonalized against the existing basis and continued as if this were another Lanczos vector. That is not valid for the tridiagonal Lanczos projection: after injecting a non-Krylov vector, the projected operator is no longer represented by only the diagonal and nearest off-diagonal entries being filled by the recurrence.

This PR changes that behavior to stop the Krylov expansion on beta breakdown and use the subspace already built.

## Investigation Notes

The observed CI failure in #888 was an energy-convergence failure, but local stress testing also reproduced a stronger failure mode: `Lanczos_Exp` returned a tensor with non-finite norm from finite normalized input, after which the next `Svd_truncate` call failed in LAPACK.

Tracing the failing call showed:

```text
input tensor shape: [2, 2, 1]
true vector length: 4
reported Hop->nx(): 10
input norm: 1.0
output norm: NaN
```

Fixing only the TDVP `LinOp::nx()` mismatch removed that particular impossible-dimension trigger in the test, but the run was still nondeterministic and could still produce final energies far outside the `1e-6` tolerance. Removing the random beta-breakdown recovery made the TDVP run deterministic in local stress testing.

## Validation

Local validation with a Python 3.11 editable build:

```text
30 direct TDVP runs:
  all final errors = 6.2261307221e-13

20 fresh pytest process runs:
  pytests/example/TDVP/tdvp1_dense_test.py passed 20/20
```

Also ran:

```text
git diff --check
```

Co-authored-by: Codex <codex@openai.com>

### DIFF
--- a/src/linalg/Lanczos_Exp.cpp
+++ b/src/linalg/Lanczos_Exp.cpp
@@ -264,19 +264,9 @@ namespace cytnx {
           } else {  // beta too small -> the norm of new vector too small. This vector cannot span
                     // the new dimension
             if (verbose) {
-              std::cout << "beta too small, pick another vector." << i << std::endl;
+              std::cout << "beta too small. Break." << i << std::endl;
             }
-            // pick a new vector perpendicular to all vector in Vs
-            v = Gram_Schmidt_internal(Vs).relabel_(v.labels());
-            auto v_norm = Dot_internal(v, v);
-            // if the picked vector also too small, break and construct expH
-            if (abs(v_norm) <= beta_tol) {
-              if (verbose) {
-                std::cout << "All vector form the space. Break." << i << std::endl;
-              }
-              break;
-            }
-            v = v / v_norm;
+            break;
           }
           Vk.append(v.get_block_().contiguous());
           Vs.push_back(v);
--- a/src/linalg/Lanczos_Exp.cpp
+++ b/src/linalg/Lanczos_Exp.cpp
@@ -264,7 +264,7 @@ namespace cytnx {
           } else {  // beta too small -> the norm of new vector too small. This vector cannot span
                     // the new dimension
             if (verbose) {
-              std::cout << "beta too small. Break." << i << std::endl;
+              std::cout << "beta too small. Break at iteration " << i << std::endl;
             }
             break;
           }
--- a/src/linalg/Lanczos_Exp.cpp
+++ b/src/linalg/Lanczos_Exp.cpp
@@ -4,9 +4,11 @@
 #include "Tensor.hpp"
 #include "LinOp.hpp"
 
+#include <algorithm>
 #include <cfloat>
-#include <vector>
 #include <cmath>
+#include <limits>
+#include <vector>
 #include "UniTensor.hpp"
 #include "utils/vec_print.hpp"
 #include <iomanip>
@@ -20,6 +22,8 @@ namespace cytnx {
     using namespace std;
 
     namespace {
+      constexpr double kBetaBreakdownRoundoff = 100.0;
+
       // <A|B>
       Scalar Dot_internal(const UniTensor &A, const UniTensor &B) {
         return Contract(A.Dagger(), B).item();
@@ -42,6 +46,47 @@ namespace cytnx {
         }
         return u;
       }
+
+      double dtype_epsilon_internal(const unsigned int dtype) {
+        if (dtype == Type.Float || dtype == Type.ComplexFloat) {
+          return std::numeric_limits<float>::epsilon();
+        }
+        return std::numeric_limits<double>::epsilon();
+      }
+
+      double float_cvgcrit_floor_internal() {
+        return kBetaBreakdownRoundoff * std::numeric_limits<float>::epsilon();
+      }
+
+      unsigned int krylov_matrix_dtype_internal(const unsigned int dtype) {
+        if (dtype == Type.Float) {
+          return Type.Double;
+        }
+        if (dtype == Type.ComplexFloat) {
+          return Type.ComplexDouble;
+        }
+        return dtype;
+      }
+
+      unsigned int lanczos_exp_output_dtype_internal(const unsigned int op_dtype,
+                                                     const unsigned int out_dtype) {
+        if (op_dtype == Type.Float) {
+          return Type.is_complex(out_dtype) ? Type.ComplexFloat : Type.Float;
+        }
+        if (op_dtype == Type.ComplexFloat) {
+          return Type.ComplexFloat;
+        }
+        return out_dtype;
+      }
+
+      void cast_dense_output_internal(UniTensor &out, const unsigned int dtype) {
+        if (out.dtype() == dtype) {
+          return;
+        }
+        out = out.astype(dtype);
+      }
+
+      double scalar_abs_internal(const Scalar &s) { return static_cast<double>(s.abs()); }
 
       Tensor resize_mat_internal(const Tensor &src, const cytnx_uint64 r, const cytnx_uint64 c) {
         const auto min_r = std::min(r, src.shape()[0]);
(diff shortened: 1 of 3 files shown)


src/linalg/Lanczos_Exp.cpp
```diff
@@ -264,19 +264,9 @@ namespace cytnx {
           } else {  // beta too small -> the norm of new vector too small. This vector cannot span
                     // the new dimension
             if (verbose) {
-              std::cout << "beta too small, pick another vector." << i << std::endl;
+              std::cout << "beta too small. Break." << i << std::endl;
```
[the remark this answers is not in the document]
ianmccul: Will fix.

src/linalg/Lanczos_Exp.cpp
```diff
@@ -264,19 +264,9 @@ namespace cytnx {
...
-            auto v_norm = Dot_internal(v, v);
-            // if the picked vector also too small, break and construct expH
-            if (abs(v_norm) <= beta_tol) {
-              if (verbose) {
-                std::cout << "All vector form the space. Break." << i << std::endl;
-              }
-              break;
-            }
-            v = v / v_norm;
+            break;
```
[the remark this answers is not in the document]
ianmccul: This is probably a separate issue.  The value of `beta_tol` is fixed to 1e-6, which is really huge for a Krylov stopping criteria, and much bigger than the default convergence tolerance of 1e-10.  Will address this separately.

ianmccul:
Reverted accidental push of commit [04a0c93](https://github.com/Cytnx-dev/Cytnx/pull/889/commits/04a0c93920c82fc3033dbc03100ffad6e6faa986), that belongs in a separate PR.  The Codex review comment above is not relevant to that PR, but might be relevant to the new one.

src/linalg/Lanczos_Exp.cpp
```diff
@@ -354,6 +397,8 @@ namespace cytnx {
         out = Contracts({T, VkDag_ut, B}, "", true);
         out = Contract(out, Vk_ut);
         out.set_rowrank_(v.rowrank());
+        cast_dense_output_internal(out,
+                                   lanczos_exp_output_dtype_internal(Hop->dtype(), out.dtype()));
```
IvanaGyro, line 401: Does this mean we calculate the float input with double? Is this a expected behavior for a math lib?

ianmccul: This PR calls `ExpM` function with double precision, even if the input is float (via #890).  This is to work around precision problems in `ExpM`, that calculates the matrix exponential by a generic diagonalization and multiplying by the inverse eigenmatrix, which is an inherently inaccurate way to calculate the exponential (especially so in the Hermitian case).  This is only for the small dense subspace exponential operation, the main `matvec` tensor contraction is always done in the input precision.  The complex-float test demonstrates this: it fails with the `float` version of `ExpM`. No such workaround is possible for double precision inputs, they are inherently inaccurate with this implementation of `ExpM`.

Generally, it is not uncommon in numerical algorithms to use higher precision internally.  A common example is linear solvers, where it is common to accumulate the residual vector in higher precision, even when the key operations are done in input precision. In this case it is a pure workaround though: a proper Pade-style matrix exponential function is certainly capable of getting adequate results with single-precision.

Note that `ExpM` always return complex matrix, even when the input is purely real. So the final output tensor is complex even for real inputs.  This behavior is unchanged by this PR.

This PR does not attempt to fix the problems with `ExpM`.

IvanaGyro (approved):

src/linalg/Lanczos_Exp.cpp
ianmccul: For the record, this was not due to ExpM using an eigenvalue decomposition to calculate the exponential (although that is not the the best way), but issue #914 : ExpM for Float is completely broken and reads a `float*` buffer as `double*` and produces garbage.

```Python
import cytnx
import numpy as np
from scipy.linalg import expm, eig

y = np.array([[1,2],[3,4]])
print("Test matrix:\n", y)

print("\nSciPy result for expm:")
print (expm(y))

print("\nWrong result, Exp(Float) reads wrong dtype, overflows the buffer:")
x = cytnx.Tensor([2,2], cytnx.Type.Float)
x[0,0] = 1
x[0,1] = 2
x[1,0] = 3
x[1,1] = 4
print(cytnx.linalg.ExpM(x))

print("Correct result, Exp(Double) is OK, although promoted to Complex")
x = cytnx.Tensor([2,2], cytnx.Type.Double)
x[0,0] = 1
x[0,1] = 2
x[1,0] = 3
x[1,1] = 4
print(cytnx.linalg.ExpM(x))
```

Output:
```
Test matrix:
 [[1 2]
 [3 4]]

SciPy result for expm:
[[ 51.9689562   74.73656457]
 [112.10484685 164.07380305]]

Wrong result, Exp(Float) reads wrong dtype, overflows the buffer:

Total elem: 4
type  : Complex Double (Complex Float64)
cytnx device: CPU
Shape : (2,2)
[[1.00000e+00+4.08008e-315j 1.35937e-08-1.86634e-315j ]
 [-6.24911e-08-2.79951e-315j 1.00000e+00+1.28057e-315j ]]



Correct result, Exp(Double) is OK, although promoted to Complex

Total elem: 4
type  : Complex Double (Complex Float64)
cytnx device: CPU
Shape : (2,2)
[[5.19690e+01+0.00000e+00j 7.47366e+01+0.00000e+00j ]
 [1.12105e+02+0.00000e+00j 1.64074e+02+0.00000e+00j ]]
```

Its ~~turtles~~ bugs all the way down!

PR #891, opened 2026-06-13
[codex] allow Fromfile count zero

ianmccul:
Fixes the current master failure in `StorageTest.FromfileCountZeroReturnsEmptyStorage` after #885/#884 landed together.

`Storage::Fromfile` already validates the file path, dtype, file byte size, and requested count against the number of elements in the file. The only remaining blocker was a stale explicit `count == 0` rejection. Since zero-size typed storage is supported by `StorageImplementation::Init(0, ...)`, `count=0` can return an empty Storage/Tensor with the requested dtype without reading any payload bytes.

This change:
- removes the `count == 0` rejection from `Storage::Fromfile`;
- updates the Storage/Tensor Fromfile docs to state that `count=0` returns an empty object;
- adds Python coverage for `Tensor.Fromfile(..., count=0)` returning an empty tensor with the requested dtype.

Co-authored-by: Codex <codex@openai.com>

### DIFF
--- a/src/backend/Storage.cpp
+++ b/src/backend/Storage.cpp
@@ -189,7 +189,6 @@ namespace cytnx {
   Storage Storage::Fromfile(const std::string &fname, const unsigned int &dtype,
                             const cytnx_int64 &count) {
     cytnx_error_msg(dtype == Type.Void, "[ERROR] Cannot have Void dtype.%s", "\n");
-    cytnx_error_msg(count == 0, "[ERROR] count cannot be zero!%s", "\n");
 
     Storage out;
     cytnx_uint64 Nbytes;
@@ -295,9 +294,9 @@ namespace cytnx {
   }
 
   void Storage::_Loadbinary(std::fstream &f, const unsigned int &dtype, const cytnx_uint64 &Nelem) {
-    // before enter this func, makesure
+    // Before entering this function, ensure:
     // 1. dtype is not void.
-    // 2. the Nelement is consistent and smaller than the file size, and should not be zero!
+    // 2. Nelem is consistent and no larger than the file size.
 
     // check:
     cytnx_error_msg(!f.is_open(), "[ERROR] invalid std::fstream!.%s", "\n");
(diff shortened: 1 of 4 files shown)


IvanaGyro:
A user should directly initialize an empty storage with the constructor instead of calling `FromFile` with zero count. Passing zero count to `FromFile` may be an accident.

ianmccul:
Generic code should not suddenly fail in edge cases, such as a zero count.

src/backend/Storage.cpp
IvanaGyro, line 202: We can fast return empty storage after checking if the file exists if count is zero.

ianmccul: I don't think `count == 0` needs a separate fast path.

The current implementation already has the simple generic behavior: validate the file, validate the dtype/count against the file size, open the file, and ask `_Loadbinary` to load `count` elements. When `count == 0`, this naturally reads zero elements and returns an empty object with the requested dtype. The core of the PR is one line: remove an unnecessary check that aborts if `count == 0`, because the function as written works just fine in the limit of `count -> 0`.

Adding a separate "check that the file exists, then return" path would make `count == 0` special even though the generic path already handles it. I don't see a practical benefit from that, and keeping one code path is  easier to reason about and maintain. If users are worried about the performance of `FromFile()` with `count == 0`, they can handle that at the call site and avoid the call completely. The library should not add a separate fast path here, because it would either skip the normal validation steps or duplicate them in a second code path.

ianmccul: If performance mattered here, the natural cleanup would be the generic path:
- open once;
- seek to end for size;
- seek back to beginning;
- validate size/count;
- bulk-read exactly once;
- check stream state after read.

That removes the double-open for every call, not just count == 0, and keeps one code path. Something like:

```c++
std::ifstream f(fname, std::ios::binary | std::ios::ate);
cytnx_error_msg(!f, "[ERROR] Cannot open file '%s'", fname.c_str());

const auto end = f.tellg();
cytnx_error_msg(end < 0, "[ERROR] Cannot determine size of file '%s'", fname.c_str());

const auto nbytes = static_cast<cytnx_uint64>(end);
f.seekg(0, std::ios::beg);
cytnx_error_msg(!f, "[ERROR] Cannot seek file '%s'", fname.c_str());
```
Thern read from the same stream.

That would be a legitimate implementation cleanup. Still not an important performance win compared with actual tensor operations, but at least it improves the real path instead of adding a zero-count branch.

IvanaGyro (approved):

Issue #899, opened 2026-06-15
Split dense Tensor and UniTensor LinOp interfaces


ianmccul:
## Background

Follow-up from #894.

`LinOp` currently exposes two virtual `matvec` overloads:

```cpp
virtual Tensor matvec(const Tensor& Tin);
virtual UniTensor matvec(const UniTensor& Tin);
```

In practice these represent different kinds of linear operators.

## Problem

A dense `Tensor` map and a block/symmetric `UniTensor` map are not generally interchangeable. Most consumers accept one or the other, and most `LinOp` implementations are expected to implement only one overload while the other overload is invalid.

That means the current design models two unrelated interfaces as one base class with two virtual functions, where a concrete object may only support one of them. This makes it harder to enforce the operator contract, because the base class cannot easily know which input shape/type is meaningful for a particular instance.

This is closely related to #894 and to the broader `LinOp` contract issue: the object that declares the domain dimension should be able to validate its inputs, but the current interface does not clearly model the operator's domain.

## Suggested direction

Consider splitting the current `LinOp` interface into separate types, for example:

- one interface for dense `Tensor` vector maps,
- one interface for `UniTensor` maps,
- possibly a shared non-virtual base for metadata such as dtype/device/name if that remains useful.

This would make it clearer which `matvec` overload is valid, and would make it easier to enforce dimension, dtype, device, and shape contracts in one place.

Opened by Codex on behalf of @ianmccul.

Issue #898, opened 2026-06-15
LinOp should enforce dimension, dtype, and device contracts centrally


ianmccul:
## Background

Follow-up from #894.

`LinOp` stores `nx`, `dtype`, and `device`, and the documentation says these should match the vectors passed to the operator:

> `dtype` the Operator's dtype. Note that this should match the input/output Tensor's dtype.

However, the current API does not give `LinOp` a central place to enforce that contract. Consumers of `LinOp` check some pieces themselves, inconsistently, and callback outputs can still violate the contract.

## Problem

The object that declares the operator domain also needs to validate the vectors passed through that operator.

At minimum, `LinOp` should verify:

- the flattened input dimension matches `nx()`,
- the input dtype matches `dtype()`,
- the output dtype matches `dtype()`,
- the input/output device matches `device()`,
- ideally, the input/output shape matches the expected shape, not merely the flattened dimension.

This should not be left only to every consumer of `LinOp`, because mistakes are subtle and have already occurred in multiple places.

For example, code like this in `src/linalg/Lanczos.cpp` hides a dtype contract violation:

```cpp
auto nextTens = Hop->matvec(buffer).astype(Hop->dtype());
cytnx_error_msg(nextTens.dtype() != Hop->dtype(),
                "[ERROR][Lanczos], the output dtype in the matvec is not "
                "consistent with the one in LinOp.%s",
                "\n");
```

The check is ineffective because the result has already been cast to `Hop->dtype()`. For real Lanczos, this can silently accept a `LinOp` implementation that returns `Double` while declaring `Float`, or vice versa.

## Suggested direction

Redesign `LinOp::matvec()` so the public API is a non-virtual checked wrapper, with the implementation hook moved to protected virtual functions. For example, conceptually:

```cpp
Tensor matvec(const Tensor& in) {
  validate_input(in);
  Tensor out = matvec_impl(in);
  validate_output(out);
  return out;
}

protected:
  virtual Tensor matvec_impl(const Tensor& in);
```

The exact names and API design need discussion, but the important point is that validation belongs in `LinOp`, not only in downstream iterative solvers.

This is separate from #894, which tracks adding defensive checks to current consumers. Those checks are still useful, but they should be a backup rather than the primary contract enforcement.

Opened by Codex on behalf of @ianmccul.

PR #904, opened 2026-06-16
Deprecate the LinOp "mv_elem" element-prestore path

ianmccul:
## Summary

`LinOp` exposes two modes selected by a constructor string: `"mv"` (subclass and override `matvec`) and `"mv_elem"` (pre-store individual non-zero elements via `set_elem(i, j, v)` / `operator()(i, j)`, and let `LinOp` build the matvec for you). This PR **deprecates the `"mv_elem"` path** while keeping it fully functional. No behaviour is removed.

## Why deprecate `mv_elem`

The `mv_elem` path is redundant with the primary `LinOp` interface and is worse on every axis:

- **It is advertised as an optimization but is pathologically slow.** The user guide claimed *"Cytnx handles the internal structure and optimizes the matvec performance."* In reality `LinOp::_mv_elemfunc` (`src/LinOp.cpp`) evaluates the product element-by-element:

  ```cpp
  for (auto it = _elems.begin(); ...) {
    Tensor e_i = it->second.second;
    for (j ...) out(it->first) += e_i(j) * Tin(v_i[j]);
  }
  ```

  Each non-zero triggers scalar `Tensor::operator()` indexing (returning `Tproxy` objects) plus a 1-element Tensor multiply/add — heap allocations and proxy round-trips per element. That is far slower than a hand-written `matvec`, and slower than a real CSR sparse matvec.

- **The storage is not a real sparse format.** Elements live in `std::map<row, (std::vector<col>, Tensor)>`, with the value Tensor grown one element at a time via `Tensor::append`. Poor locality, no BLAS, CPU-only.

- **It is redundant.** The whole point of `LinOp` is "override `matvec`". The guide's own *"sparse data structure with mapping function"* example already shows the recommended approach (type `"mv"` + `matvec`). `mv_elem` is a strictly-worse second way to do the same thing — and the canonical example (`example/ED/ed_ising_mve.py`) computes its elements in a Python `pre_construct` loop, which is exactly the "many Python for-loops" that `mv_elem` is supposed to avoid, so it doesn't even deliver its stated benefit.

- **It only handles dense `Tensor`** (it errors on `UniTensor`), and the `_type`-string runtime dispatch in `matvec` is a hand-rolled, poorer substitute for a virtual override.

- **It silently ignores the operator's declared `nx` and `device`.** `_mv_elemfunc` builds its output as `Tensor(Tin.shape(), Tin.dtype(), Tin.device())` and never consults `_nx` or `_device`. `set_elem` stores each element in a CPU `Tensor` (only the operator's `dtype` is honoured), so the path is effectively CPU-only regardless of the `device` passed to the constructor, and `nx` is never used to validate element indices or the input length. This is exactly the kind of unenforced-contract problem that #898 is meant to fix.

- **It blocks the planned `LinOp` redesign.** The element-store state (`_elems`, `set_elem`, `operator()`, `_mv_elemfunc`, the `_type` switch) is dead weight that complicates the follow-up work to add central contract validation (#898) and to split the dense-`Tensor` vs `UniTensor` interfaces (#899). Shrinking the surface first is the right sequencing.

## Changes

- **`include/LinOp.hpp`** — the constructor emits a runtime `cytnx_warning_msg` when `type == "mv_elem"` is requested (the deprecation signal that reaches Python users, since the type is a runtime string); `set_elem<T>()` and `operator()(i, j)` are marked `[[deprecated(...)]]` for a compile-time warning to C++ users; and the Doxygen documentation is updated with `@deprecated`, pointing at type `"mv"` + overriding `matvec()`.
- **`pybind/linop_py.cpp`** — the `linop_binding` function is wrapped in a `-Wdeprecated-declarations` push/pop. The `set_elem` bindings are kept so the deprecated API stays callable from Python; the warnings from referencing the now-deprecated `set_elem` are silenced to keep the build output clean.
- **`docs/source/guide/itersol/LinOp.rst`** — restructured so the recommended approach leads: a new *"Store the operator in a real sparse data structure"* section shows a `LinOp` of type `"mv"` whose `matvec` delegates to a real sparse format. The accompanying example (`docs/code/python/doc_codes/guide_itersol_LinOp_sparse_csr.py`, with output `.../outputs/guide_itersol_LinOp_sparse_csr.out`) stores the operator as a `scipy.sparse` CSR matrix and bridges cytnx <-> numpy at the boundary via `Tensor.numpy()` / `cytnx.from_numpy()`, reproducing the exact same operator and output as the old `mv_elem` example. The `mv_elem` material is demoted to a *"Deprecated: prestoring elements with mv_elem"* section carrying a `.. deprecated:: 1.1.0` note, and the previous false "optimizes the matvec performance" claim is corrected to describe the path accurately (element-by-element, slower than a real sparse kernel, dense-only, ignores `device`/`nx`).
- **`docs/tests/test_doc.py`** — adds the matching `test_guide_itersol_LinOp_sparse_csr` entry (the doc-output regeneration harness; note this harness is not run by CI, which only runs `pytest pytests/`).

Nothing is removed; the example, doc snippet, and existing user code keep working (now with a deprecation warning).

## Verification

- `cmake --build --preset openblas-cpu` succeeds, with no `-Wdeprecated-declarations` warnings introduced by this change (only a pre-existing third-party hptt warning).
- C++: a small test TU confirms both `set_elem` and `operator()(i, j)` emit `-Wdeprecated-declarations`.
- Python: constructing `LinOp("mv_elem", ...)` prints the deprecation warning while `LinOp("mv", ...)` is silent; `example/ED/ed_ising_mve.py` and `docs/code/python/doc_codes/guide_itersol_LinOp_sparse_mv_elem.py` still produce correct results, and the new CSR example reproduces the same output.
- clang-format v14 (the version pinned by CI / pre-commit) reports the changed C++ files clean.

## Migration

- **Python users** should construct the `LinOp` with type `"mv"` and override `matvec`, delegating the matrix-vector product to a real sparse representation (e.g. `scipy.sparse`). See the new CSR example added to the guide.
- **C++ users** of `mv_elem` are better served by a small dedicated subclass rather than the element-store path — for example a `template <class T> class SparseLinOp : public LinOp` that owns a CSR (or other) sparse representation and implements `matvec` directly. (No such class is added in this PR; it is only the suggested replacement pattern.)

## Notes / follow-ups

- This is a deprecation only; actual removal of the `mv_elem` machinery (`set_elem`, `operator()`, `_elems`, `_mv_elemfunc`, the `_type` switch) should happen in a later release once the deprecation has shipped.
- Related cleanup tracked in #898 (central `LinOp` contract validation) and #899 (split dense-`Tensor` vs `UniTensor` interfaces). This PR is independent of both but reduces the `LinOp` surface they will touch.

---

Co-authored-by: Claude <noreply@anthropic.com>


ianmccul:
All Gemini review comments addressed.

ianmccul:
@codex review

ianmccul:
Codex raises a good point above and I am not sure what the best option is here. The purpose of the example is to show how cytnx can be wired into other python code. I am surprised there is not already an example showing how to interoperate with cytnx and `scipy`. I think adding `scipy` to the packages needed to build the documentation is reasonable here.

PR #905, opened 2026-06-16
Remove the LinOp "mv_elem" path

ianmccul:
> **Stacked PR.** This is stacked on #904 (branch `deprecate-linop-mv-elem`), which deprecates `mv_elem`. **Do not merge until the project is ready to remove mv_elem from the public API.** The base branch is set to `deprecate-linop-mv-elem`; once #904 lands, GitHub will automatically retarget this PR to `master`. If #904 changes before it merges, this branch will be rebased so its effective diff against `master` stays clean. The branch tip is available for other developers who want to build on the post-removal API.

## Summary

Follow-up to #904. That PR deprecated the `LinOp` `"mv_elem"` element-prestore path; this PR removes it entirely. The element-store machinery (`set_elem`, `operator()(i, j)`, the internal `_elems` map, `_mv_elemfunc`, the `_type` switch, and `_print`) is deleted, leaving `LinOp` as a clean "override `matvec()`" interface.

See #904 for the full rationale (the path was advertised as an optimization but evaluated the product element-by-element through scalar tensor indexing, was not a real sparse format, was dense-`Tensor`-only, and silently ignored the declared `nx`/`device`).

## API change: the constructor `type` argument

`LinOp` no longer needs a `type`. The preferred constructor is simply `LinOp(nx, dtype = Double, device = cpu)`. The legacy leading-`type` form is still accepted but its handling now differs by language, reflecting that the C++ API is internal while the Python API is the public surface:

- **C++**: the `LinOp(type, nx, dtype, device)` constructor is retained but marked `[[deprecated]]`. It delegates to `LinOp(nx, dtype, device)` and hard-errors unless `type == "mv"` (the removed `"mv_elem"` and any other value are errors). It is not used anywhere in Cytnx itself — all in-tree C++ (the library, tests, benchmarks, and C++ doc snippets) has been migrated to `LinOp(nx, ...)`.
- **Python**: the bindings keep a backwards-compatibility shim so existing code such as `cytnx.LinOp("mv", nx, ...)` keeps working unchanged (and `cytnx.LinOp(nx, ...)` works too). The shim lives only in `pybind/` and can stay indefinitely; it constructs through the plain `LinOp(nx, ...)` constructor (not the deprecated C++ one), so the binding compiles warning-free. A non-`"mv"` type is a hard error.

## Changes

- **`include/LinOp.hpp`** — removed the `_type`, `_elems`, `_elems_it` members, the `_mv_elemfunc` declaration, `set_elem<T>()`, `operator()(i, j)`, and `_print()`. The primary constructor is now `LinOp(nx, dtype, device)`; the legacy `LinOp(type, nx, dtype, device)` is retained as `[[deprecated]]` and hard-errors unless `type == "mv"`. Dropped the `<fstream>`, `<functional>`, `<map>`, `<utility>`, and `<algorithm>` includes that the element store needed.
- **`src/LinOp.cpp`** — removed `_print` and `_mv_elemfunc`; `matvec(Tensor)` / `matvec(UniTensor)` now simply error if not overridden (no `_type` branching), with a clearer "must be overridden" message.
- **`pybind/linop_py.cpp`** — removed all `set_elem` bindings (and the `-Wdeprecated-declarations` / MSVC `4996` suppression introduced in #904, no longer needed), and the `__repr__` that called `_print` (it only ever printed `mv_elem` element state, so Python `repr(linop)` now falls back to the default object repr). Bound `py::init<nx, dtype, device>` plus a small factory `py::init` providing the legacy `("mv", nx, ...)` form.
- **All in-tree C++ call sites migrated to `LinOp(nx, ...)`** — `src/tn_algo/DMRG.cpp`, the CPU and GPU `tests/linalg_test/` suites, `benchmarks/linalg/`, the two `tests/OLDtest.cpp` files, and the C++ doc snippets under `docs/code/cplusplus/`. None of these now use the deprecated `type` argument.
- **`docs/source/guide/itersol/LinOp.rst`** — removed the deprecated `mv_elem` section. In its place added a `.. note::` documenting that `"mv_elem"` existed, was deprecated in 1.1.0, has now been removed, and what to do instead (construct with `nx` and override `matvec`, delegating to a real sparse structure such as `scipy.sparse`) — so users who skip the deprecation release still get migration guidance.
- **`docs/code/python/doc_codes/guide_itersol_LinOp_sparse_csr.py`** — updated the comment to drop the reference to the removed path, and fixed the guide's `emphasize-lines` (which had drifted after the `import cytnx` line was added in #904).
- **Removed files** — `docs/code/python/doc_codes/guide_itersol_LinOp_sparse_mv_elem.py` and its `.out`, the `test_guide_itersol_LinOp_sparse_mv_elem` entry in `docs/tests/test_doc.py`, the `example/ED/ed_ising_mve.py` example, and its `pytests/example/ED/ed_ising_mve_test.py` test (which was also one of the few remaining users of the unreliable `Lanczos` `method="ER"` path).

## Verification

- `cmake --build --preset openblas-cpu` succeeds (library + Python module) with **no** `-Wdeprecated-declarations` warnings — nothing in-tree uses the deprecated C++ constructor.
- clang-format v14 (the CI-pinned version) reports `include/LinOp.hpp`, `src/LinOp.cpp`, and `pybind/linop_py.cpp` clean.
- C++: a small TU confirms `LinOp(4)` compiles cleanly while `LinOp("mv", 4)` emits the expected `-Wdeprecated-declarations` warning; both construct successfully.
- Python:
  - `cytnx.LinOp(1000)` and `cytnx.LinOp("mv", 8, ...)` both construct.
  - `cytnx.LinOp("mv_elem", 10)` and any other non-`"mv"` type raise.
  - `hasattr(cytnx.LinOp, "set_elem")` is `False`.
  - A Python subclass overriding `matvec` and calling `LinOp.__init__(self, "mv", n)` dispatches correctly through the factory shim/trampoline.
  - The CSR guide example reproduces `400.0 / 7.0`, and the retained `example/ED/ed_ising.py` still runs.

## Notes / follow-ups

- Related cleanup tracked in #898 (central `LinOp` contract validation) and #899 (split dense-`Tensor` vs `UniTensor` interfaces). Removing `mv_elem` shrinks the surface those will touch; the now-tiny `LinOp` is a better starting point for that redesign.

---

Co-authored-by: Claude <noreply@anthropic.com>


ianmccul:
@codex review

PR #931, opened 2026-06-22
Remove the unmaintained C++ tn_algo MPS/MPO/DMRG layer

### LINKED ISSUE #920: tn_algo.MPS construction crashes in Into_Lortho: Svd reuses "_aux_L" producing duplicated labels
## Summary

Constructing a `tn_algo::MPS` with the 3-argument constructor (`MPS(N, phys_dim, virt_dim)`) — or any path that runs `RegularMPS::Init` / `Init_Msector` — aborts with a `set_labels` "duplicated element(s)" error before construction completes. The MPS class is effectively unusable from Python for the regular (non-symmetric) case.

## Reproduction

```python
import cytnx
mps = cytnx.tn_algo.MPS(4, 2, 8)   # N=4 sites, phys_dim=2, virt_dim=8
```

Result:

```
# Cytnx error occur at void cytnx::UniTensor_base::set_labels(const std::vector<std::string>&)
# error: [ERROR][set_labels][UniTensor] the input labels cannot contain duplicated element(s).
# file : src/UniTensor_base.cpp (352)
Stack trace:
  ...
  cytnx::linalg::Svd(const cytnx::UniTensor&, const bool&)
  cytnx::tn_algo::RegularMPS::Into_Lortho()
  ...
```

The same crash occurs via `MPS.Init_Msector(...)`, which also calls `Into_Lortho()` at the end.

## Root cause

The bond labels collide between successive `Svd` calls during the left-orthogonalization sweep.

1. `RegularMPS::Init` (`src/tn_algo/RegularMPS.cpp`) builds rank-3 site tensors whose labels chain across the virtual bonds: site 0 keeps the default labels `{"0","1","2"}`, and site `k>=1` is relabelled to `{2k, 2k+1, 2k+2}`, so adjacent sites share their virtual-bond label. It then calls `Into_Lortho()`.

2. `RegularMPS::Into_Lortho()` (`src/tn_algo/RegularMPS.cpp:202`) sweeps left to right:

   ```cpp
   for (cytnx_int64 p = 0; p < this->size() - 1 - this->S_loc; p++) {
     auto out = linalg::Svd(this->_TNs[p]);
     auto s = out[0];
     this->_TNs[p] = out[1];                 // U
     auto vt = out[2];
     this->_TNs[p + 1] = Contract(Contract(s, vt), this->_TNs[p + 1]);
   }
   ```

3. `linalg::Svd` (`src/linalg/Svd.cpp`) always names the singular-value bonds with the **fixed** labels `"_aux_L"` / `"_aux_R"`:

   ```cpp
   Cy_S.Init({newBond, newBond}, {std::string("_aux_L"), std::string("_aux_R")}, ...);   // Svd.cpp:113
   ...
   labelU.push_back(Cy_S.labels()[0]);   // appends "_aux_L"   (Svd.cpp:131)
   Cy_U.relabel_(labelU);                //                    (Svd.cpp:132)
   ```

4. Walking the first two iterations for `MPS(4, 2, 8)`:
   - `Svd(_TNs[0])` with `oldlabel = {"0","1","2"}`, `rowrank = 2` →
     `U = {"0","1","_aux_L"}`, `S = {"_aux_L","_aux_R"}`, `vT = {"_aux_R","2"}`.
   - `_TNs[1] = Contract(Contract(S, vT), _TNs[1])` contracts away `"_aux_R"` and the shared `"2"` bond, leaving `_TNs[1] = {"_aux_L","3","4"}` — i.e. the left virtual bond is now labelled `"_aux_L"`.
   - Next iteration, `Svd(_TNs[1])` builds `labelU = {oldlabel[0], oldlabel[1], "_aux_L"} = {"_aux_L","3","_aux_L"}`. The repeated `"_aux_L"` makes `Cy_U.relabel_(labelU)` → `set_labels` throw.

The collision is structural: the `U` output of one `Svd` carries `"_aux_L"`, that label survives into the next site through the `S·vT` contraction, and the next `Svd` unconditionally reintroduces `"_aux_L"` for its own new bond.

## Affected code

- `src/tn_algo/RegularMPS.cpp` — `Into_Lortho()` (sweep using `linalg::Svd`), and the label scheme set up in `Init` / `Init_Msector`.
- `src/linalg/Svd.cpp` — `Svd_Dense_UT_internal` hard-codes the `"_aux_L"` / `"_aux_R"` bond labels.

## Notes

A likely fix is to relabel the new virtual bond produced by each `Svd` to a value that cannot collide with neighbouring bonds before contracting it into the next site (e.g. rename `"_aux_L"`/`"_aux_R"` to the site-specific virtual-bond label used by `Init`), rather than leaving the generic `"_aux_L"`/`"_aux_R"` labels in place across sweep steps.

This was discovered while verifying the `MPS.Save` Python binding fix (#919): a `Save`/`Load` round-trip test cannot be written because a regular `tn_algo.MPS` cannot currently be constructed. Filing separately as the binding fix and this construction bug are independent.

---
Investigated and filed by Claude (Claude Code).

### LINKED ISSUE #924: Test the C++ tn_algo API and redesign its Python bindings
## Background

PR #922 removes the Python bindings for `cytnx.tn_algo` (`MPS`, `MPO`, `DMRG`), because the bindings exposed C++ code that is entirely untested and demonstrably broken (see #920, where constructing a regular `tn_algo::MPS` crashes with a duplicated-label error in `Into_Lortho()`).

The underlying C++ implementation in `src/tn_algo/` (`RegularMPS.cpp`, `SymMPS.cpp`, `MPO.cpp`, `DMRG.cpp`) is left in place by #922; only the Python-facing surface is removed. That C++ code itself has no test coverage, so its correctness is currently unknown beyond the one confirmed bug in #920.

## What this issue tracks

1. **Add C++ test coverage for `src/tn_algo`.**
   - Construction and orthogonalization round-trips for `RegularMPS` and `SymMPS` (`Init`, `Init_Msector`, `Into_Lortho`, `Into_Rortho`).
   - `MPO` construction and contraction against a known small system.
   - `DMRG` convergence on a system with a known exact ground-state energy (e.g. a small Heisenberg or transverse-field Ising chain).
   - Regression coverage for the bug identified in #920 (label collision across successive `Svd` calls in `Into_Lortho`).

2. **Fix the bugs that the new tests uncover**, starting with #920.

3. **Redesign the Python bindings for `tn_algo`** once the C++ implementation is verified, with a clearer support contract than the previous bindings had. Candidate questions to settle as part of the redesign:
   - Should `MPS`/`MPO`/`DMRG` be re-exposed under `cytnx.tn_algo`, or restructured to match the encoding used elsewhere in the Python API (e.g. consistency with `UniTensor`-based workflows)?
   - What is the intended relationship between `cytnx.tn_algo` and the example DMRG/TDVP scripts under `example/`, which currently implement their own standalone algorithms rather than using `cytnx.tn_algo`?
   - What test coverage (pytest) is required before the bindings are considered supported again?

## Related

- #920 — root-cause analysis of the `Into_Lortho` label collision crash.
- #919 — independent fix to the `MPS.Save` binding (recommended to drop per #922's PR description, since the binding it patches is being removed).
- #922 — removes the Python bindings that this issue is about restoring, once the underlying code is tested and the binding design is revisited.

---
Filed by Claude (Claude Code).

### LINKED ISSUE #929: Consider caching in tn_algo RegularMPS::norm (conjugates and the norm value)
## Background

Raised during review of #927 (https://github.com/Cytnx-dev/Cytnx/pull/927). `RegularMPS::norm()` accumulates the `<psi|psi>` left environment by calling `Ai.Conj()` once per site on every invocation:

```cpp
for (const UniTensor &Ai : this->_TNs) {
  UniTensor tA = Ai.relabel({"_a", "_p", "_r"});
  UniTensor tAc = Ai.Conj().relabel({"_b", "_p", "_rc"});
  L = Contract(Contract(L, tA), tAc);
  L.relabel_({"_a", "_b"});
}
```

Two caching opportunities were suggested:

1. **Per-site conjugates** — `Ai.Conj()` allocates a fresh conjugated tensor each call.
2. **The norm value itself** — repeated `norm()` calls on an unchanged MPS recompute the full contraction.

## Open questions / trade-offs

- `Conj()` of a real-dtype MPS is effectively a clone; caching it duplicates memory for little gain. For complex MPS it is a real conjugation but still cheap relative to the contraction.
- Caching the scalar norm requires an invalidation hook on every mutation of `_TNs` (Init, Into_Lortho, S_mvleft/S_mvright, and any direct `data()` mutation a caller performs), which is the harder part — `data()` hands out a non-const reference, so the cache could silently go stale.
- `norm()` is not currently called inside any hot loop (the DMRG driver normalizes singular values directly rather than via `norm()`), so the practical benefit is unclear.

## Suggested next step

Measure whether `norm()` is ever on a hot path before adding a cache; if it is, prefer caching the scalar with explicit invalidation over caching per-site conjugates. Filing to track the consideration rather than block #927.

---
Filed per review feedback on #927.

ianmccul:
This removes the remaining in-tree C++ `tn_algo` MPS/MPO/DMRG implementation.

The removed code is not covered by meaningful tests, is not used by the Python example DMRG/TDVP code paths, and has accumulated enough correctness and design problems that continuing to patch individual failures is likely to waste maintainer effort. In particular, recent issues show that this layer is brittle even for basic construction and normalization paths:

- Closes #920
- Closes #924
- Closes #929
- Supersedes #927

This PR intentionally removes the public `cytnx::tn_algo::{MPS,MPO,DMRG}` C++ API. The Python bindings for this layer have already been removed separately; this PR removes the underlying C++ implementation that would otherwise remain as an untested public API surface.

Users should construct MPS/MPO algorithms directly using `UniTensor`-based code in Python, as the existing examples already do.

Changes:

- Delete `include/tn_algo/` and `src/tn_algo/`.
- Remove `src/tn_algo` from the CMake build.
- Remove `tn_algo` headers from the umbrella `cytnx.hpp` include.
- Remove stale `tn_algo` object rules from `developer_tools/Makefile`.
- Delete the unused archival `tests/OLDtest.cpp` and `tests/gpu/OLDtest.cpp` scratch files that referenced `tn_algo`.


pcchen:
## Review — clean removal, LGTM

For a pure-deletion PR the key question is whether anything still references the removed code. I traced build files, headers, bindings, tests, examples, and docs; the removal is clean:

- **No dangling build references.** The only remaining `tn_algo` mention in a build file is `src/backend/CMakeLists.txt:53`, which is already `#add_subdirectory(tn_algo)` (commented out) — no effect.
- **No residual symbol usage.** The only `tn_algo::` / `MPS`/`MPO`/`DMRG` C++ usages outside the deleted source dirs are in the two `OLDtest.cpp` files, both deleted by this PR. `pybind/` has zero references (bindings were removed earlier in #922).
- **Deleted test files were never built.** `tests/CMakeLists.txt` / `tests/gpu/CMakeLists.txt` use explicit source lists, and `OLDtest.cpp` isn't in them — nothing to update.
- **Header install unaffected.** `CMakeLists.txt:549` uses `install(DIRECTORY include/ ...)`, so the removed subdir simply disappears; no per-subdir rule to fix.
- **Docs/examples unaffected.** The DMRG/iDMRG docs and `example/` are Python/UniTensor-based; none reference the C++ `tn_algo` API.

### Optional, non-blocking nits (cosmetic, not introduced by this PR)
- `src/backend/CMakeLists.txt:53` — the now-doubly-dead `#add_subdirectory(tn_algo)` comment could be swept up, but it's outside this PR's scope and harmless.
- `misc_doc/version.log:139` — a historical changelog line mentions `tn_algo`; changelogs are normally left as-is.

### By design (not a defect)
This intentionally removes the public `cytnx::tn_algo::{MPS,MPO,DMRG}` C++ API — a breaking change for any external C++ consumer. It's documented in the PR description and pointed at UniTensor-based Python code. Worth a changelog/release-note entry if the project tracks API breaks, but nothing to change in the diff.

**Verdict: LGTM** — self-consistent removal with no leftover references.

---
*Posted by Claude Code on behalf of @pektiong*

pcchen (approved):

pcchen:
## Merge conflict resolved

Merged latest `master` into this branch (merge commit `b1d43381`, pushed to `ianmccul:remove-cpp-tn-algo` via maintainer edit). The PR is now **MERGEABLE** (was CONFLICTING).

**Cause:** all 9 conflicts were **modify/delete** — this PR deletes the `tn_algo` layer and the two `OLDtest.cpp` scratch files, while `master` had recently *modified* those same files (the `using namespace std` removal in #952, SparseUniTensor-related edits, etc.). Since the PR's purpose is to remove these files, the resolution for every conflict is to **keep the deletion**; master's edits to soon-to-be-deleted files are moot.

Resolved by `git rm` on the 7 `src/tn_algo/*.cpp` files plus `tests/OLDtest.cpp` and `tests/gpu/OLDtest.cpp`; `developer_tools/Makefile`, `src/CMakeLists.txt`, and `include/cytnx.hpp` auto-merged cleanly.

**Verified:**
- No unmerged paths / conflict markers remain.
- No dangling `tn_algo` references in active code (`cytnx.hpp`, `src/CMakeLists.txt`, `developer_tools/Makefile` all clean; the only matches left are a historical changelog line and an already-commented-out `#add_subdirectory(tn_algo)`).
- No build/test target references the deleted `OLDtest.cpp` files.
- Net diff vs `master` is exactly the intended scope — **17 files, 2251 deletions, 0 additions** — nothing extra pulled in.

CI is re-running on the merge commit; compilation/tests will be confirmed there.

---

Issue #944, opened 2026-06-26
[build] submodule missing


pcchen:
If I freshly git clone the repo and do pip install, I will run into submodule missing problem.

````
git clone https://github.com/Cytnx-dev/Cytnx.git
cd Cytnx
pip install .

CMake Error at CMakeLists.txt:17 (message):
morse_cmake submodule missing.  Run: git submodule update --init
--recursive
````
Manually do ``git submodule update --init`` will solve the issue. 

Can we modify the pip install script to fix this? Or we should document this on user guide?

pcchen:
@IvanaGyro can you comment on this?

IvanaGyro:
`pip install .` directly trigger cmake. I decided to prevent network access in cmake process before like removing downloading HPTT in cmake. 

We should document this on user guide for manually installation. The manual installation should also include the instruction about installing other system dependencies like blas, cuQuantum, cuTENSOR, cuda.

pcchen:
Proposing a layered solution that respects @IvanaGyro's "no network access during cmake" decision.

**Root cause.** `pip install .` runs cmake directly on the git working tree (not an sdist), and `CMakeLists.txt:15-18` requires `cmake_modules/morse_cmake/modules/MorseInit.cmake`, which only exists after `git submodule update --init`. A fresh clone doesn't have it, hence the fatal error. The submodule is hosted on `gitlab.inria.fr`, so it's also an extra non-GitHub network dependency and point of failure.

`morse_cmake` is used for exactly one thing here — `include(MorseInit)` to pull in its up-to-date `FindLAPACKE`. Three options, in order of preference:

### Option A (recommended): vendor the morse modules, drop the submodule

Since we only need `MorseInit` + `FindLAPACKE` (and their direct helper modules), copy that minimal subset into `cmake/Modules/` and remove the `cmake_modules/morse_cmake` submodule from `.gitmodules`.

- ✅ No network access in cmake — aligns with the decision already made for HPTT
- ✅ `pip install .` works from a fresh clone with **zero** extra steps
- ✅ Removes the fragile gitlab.inria.fr dependency
- ⚠️ We take on responsibility for occasionally syncing the vendored modules upstream (rare — `FindLAPACKE` is stable). Add a short `cmake/Modules/README` noting the upstream commit it was copied from + LICENSE.

This fully fixes the issue rather than working around it.

### Option B: keep the submodule but make the sdist self-contained + document

[...]

Issue #857, opened 2026-05-30
Test/suite names use underscores (violates GoogleTest naming rule; ~39+ generate reserved identifiers / UB)


IvanaGyro:
(Issue written by Claude on behalf of @IvanaGyro.)

## Summary

GoogleTest's own FAQ says **test suite names and test names must not contain underscores** ([googletest FAQ — "Why should test suite names and test names not contain underscore?"](https://google.github.io/googletest/faq.html#why-should-test-suite-names-and-test-names-not-contain-underscore)). Our test suite violates this very widely, and a subset of the violations produce **reserved C++ identifiers (undefined behavior)**, not just style problems.

## Why it matters (from the FAQ)

`TEST(SuiteName, TestName)` expands to a class named `SuiteName_TestName_Test` (the macro joins the parts with `_` and appends `_Test`). Two failure modes follow:

1. **Reserved identifiers (UB).** The C++ standard reserves *"any identifier that starts with `_` followed by an upper-case letter, and any identifier that contains two consecutive underscores (`__`) anywhere."* So a `TestName` that ends in `_` (→ `..._` + `_Test` = `..__Test`), or that already contains `__`, makes the generated class name a reserved identifier — undefined behavior.
2. **Name collisions.** Even when not reserved, different underscore placements collapse to the same class name. The FAQ's example:
   ```cpp
   TEST(Time,       Flies_Like_An_Arrow) { ... }
   TEST(Time_Flies, Like_An_Arrow)       { ... }
   ```
   both generate `Time_Flies_Like_An_Arrow_Test` → conflict.

## Scope in this repo (master @ 24e8cc6c)

Counting `TEST` / `TEST_F` / `TEST_P` / `TYPED_TEST` macros under `tests/`:

- **877 of 1020** test names contain an underscore.
- **123** suite/fixture names contain an underscore (e.g. `linalg_Test`).
- Of the test names, the **reserved-identifier (UB) subset**:
  - **~39** test names **end with `_`** → generated class has `__`. Examples: `Trace_`, `Trace__str`, `Transpose_`, `Pow_`, `Tensor_Gemm_`, `Tensor_Inv_`, `DenseUt_InvM_`, `contiguous_`, `get_block_`, `get_blocks_`, `gpu_Tensor_Gemm_`, `gpu_DenseUt_Pow_`, `Conj__`.
  - **~68** test names contain `__` directly.

[...]

PR #972, opened 2026-07-05
fix(tests): make the test suite build on macOS/AppleClang

yingjerkao:
## Summary

The `RUN_TESTS=ON` build only worked on Linux/GCC with a system-installed googletest; on macOS/AppleClang it failed to configure and compile. This PR fixes four portability problems in the test suite:

- **Narrowing errors (`-Wc++11-narrowing`)** — signed loop indices (`cytnx_int64`/`int`) fed braced initializer lists for `std::vector<cytnx_uint64>` APIs (`at()`, `elem_exists()`, `uniform()`, `zeros()`), which AppleClang rejects as a hard error. Converted the affected loop variables and shape parameters to `cytnx_uint64` in `BlockUniTensor_test.cpp` (CPU and GPU), `Lanczos_Gnd_test.cpp`, `Vectordot_test.cpp`, and `linalg_test.h`. Every loop body was audited first: none used the variables in signed-dependent contexts (the only `i-1` arithmetic was in commented-out code).
- **`fill(i + 1)` with `size_t i`** — on macOS `size_t` is `unsigned long`, distinct from every Cytnx integer type, so the `fill` dispatch fails. Cast to `cytnx_uint64` in both `BlockUniTensor_test.h` files.
- **`<bits/stdc++.h>`** — a GCC-internal header absent from libc++. Replaced with the concrete standard headers in `getNconParameter.h` and `generate_networks.cpp` (CPU and GPU).
- **googletest provisioning** — nothing provided the `gtest`/`gmock` link targets; the bare names only resolved where googletest was system-installed. `tests/CMakeLists.txt` now uses `find_package(GTest REQUIRED)` and additionally requires the `GTest::gmock` target (find_package can resolve gtest without gmock), erroring out with install guidance (brew / apt / conda) when it is missing; both test CMakeLists link the namespaced `GTest::` targets. GoogleTest is a documented **prerequisite** (`tests/README.md`) rather than fetched, keeping the CMake configuration free of network access per **#944**. Also drops the top-level `gtest_force_shared_crt` line, which was set *after* `add_subdirectory(tests)` and never took effect. _(Updated after review: the earlier `find_package(GTest QUIET)` + FetchContent(v1.14.0) fallback was dropped — it added network access against #944 and could hit an `add_library ... ALIAS GTest::gtest` collision.)_

Related: #857, #944.

## Verification

On macOS (AppleClang, Darwin 25.5.0):

```
git submodule update --init --recursive
cmake -S . -B build_t -DRUN_TESTS=ON -DBUILD_PYTHON=OFF -DUSE_MKL=OFF -DUSE_HPTT=OFF -DUSE_CUDA=OFF
cmake --build build_t --target test_main
```

- `test_main` builds cleanly (zero errors) with GoogleTest provided via `find_package(GTest)` (Homebrew's `googletest` on macOS; the distro/conda package elsewhere).
- The test suites in every touched file pass: `BlockUniTensorTest.*`, all Vectordot tests, `Lanczos_Ut.nx_not_match` — 73/73.
- The bulk edits were mechanically verified to contain only `cytnx_int64` -> `cytnx_uint64` swaps.
- On Linux/GCC CI the system googletest continues to be picked up by `find_package`, so behavior there is unchanged.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -2,23 +2,23 @@
 if(BACKEND_TORCH)
 else()
 
-# GoogleTest (gtest + gmock): prefer a system/toolchain install, otherwise
-# fetch a pinned release so the test suite builds on machines without one
-# (e.g. macOS). The GTest::gmock target only exists with a config-mode
-# install or CMake's FindGTest module from 3.23+, so key the fallback on
-# the target rather than GTest_FOUND.
-find_package(GTest QUIET)
+# GoogleTest (gtest + gmock) is a required prerequisite for the test suite and
+# must be provided by the toolchain / a config-mode install. It is deliberately
+# NOT fetched here so the CMake configuration stays free of network access
+# (see #944, which moved the other optional deps to submodules/finders).
+# find_package(GTest) can succeed with only gtest -- the GTest::gmock target
+# needs a config-mode install or CMake's FindGTest from 3.23+ -- so require the
+# gmock target explicitly and fail with actionable guidance when it is missing.
+find_package(GTest REQUIRED)
 if(NOT TARGET GTest::gmock)
-  include(FetchContent)
-  # Keep the MSVC runtime selection consistent with the parent project.
-  set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
-  set(INSTALL_GTEST OFF CACHE BOOL "" FORCE)
-  FetchContent_Declare(
-    googletest
-    URL https://github.com/google/googletest/archive/refs/tags/v1.14.0.tar.gz
-    URL_HASH SHA256=8ad598c73ad796e0d8280b082cebd82a630d73e73cd3c70057938a6501bba5d7
-  )
-  FetchContent_MakeAvailable(googletest)
+  message(FATAL_ERROR
+    "GoogleTest was found but the GTest::gmock target is missing, which the test "
+    "suite links against. Install a GoogleTest that also provides GMock (it is not "
+    "fetched automatically):\n"
+    "  macOS:         brew install googletest\n"
+    "  Ubuntu/Debian: apt install libgtest-dev libgmock-dev\n"
+    "  conda:         conda install -c conda-forge gtest gmock\n"
+    "or point CMake at such an install with -DCMAKE_PREFIX_PATH=<gtest-prefix>.")
 endif()
 
 # Define the path to the test data directory
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -2,6 +2,25 @@
 if(BACKEND_TORCH)
 else()
 
+# GoogleTest (gtest + gmock): prefer a system/toolchain install, otherwise
+# fetch a pinned release so the test suite builds on machines without one
+# (e.g. macOS). The GTest::gmock target only exists with a config-mode
+# install or CMake's FindGTest module from 3.23+, so key the fallback on
+# the target rather than GTest_FOUND.
+find_package(GTest QUIET)
+if(NOT TARGET GTest::gmock)
+  include(FetchContent)
+  # Keep the MSVC runtime selection consistent with the parent project.
+  set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
+  set(INSTALL_GTEST OFF CACHE BOOL "" FORCE)
+  FetchContent_Declare(
+    googletest
+    URL https://github.com/google/googletest/archive/refs/tags/v1.14.0.tar.gz
+    URL_HASH SHA256=8ad598c73ad796e0d8280b082cebd82a630d73e73cd3c70057938a6501bba5d7
+  )
+  FetchContent_MakeAvailable(googletest)
+endif()
+
 # Define the path to the test data directory
 set(CYTNX_TEST_DATA_DIR "${CMAKE_CURRENT_SOURCE_DIR}/test_data_base")
 
@@ -71,9 +90,9 @@ target_include_directories(
 )
 target_link_libraries(
   test_main
-  gtest_main
-  gmock
-  gtest
+  GTest::gtest_main
+  GTest::gmock
+  GTest::gtest
 )
 target_link_libraries(test_main cytnx)
 
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,26 +1,40 @@
 # Cytnx C++ test suite
 
+## Prerequisites
+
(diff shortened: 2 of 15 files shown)


yingjerkao:
Added `tests/README.md` (3acd3efd) documenting how to build/run the suite and the macOS `DYLD_LIBRARY_PATH` pitfall root-caused in #974 — with the env var set, dyld's case-insensitive leaf-name lookup substitutes conda libraries for Apple system ones (Accelerate BLAS/LAPACK, ImageIO codecs, libc++) and every ARPACK-backed Lanczos/Arnoldi test segfaults inside `dsaupd_`. The README points to `-DCMAKE_BUILD_RPATH`/`install_name_tool` as the correct alternatives. This was the docs follow-up suggested when closing #974.

tests/CMakeLists.txt
```diff
@@ -2,6 +2,25 @@
...
+find_package(GTest QUIET)
+if(NOT TARGET GTest::gmock)
+  include(FetchContent)
+  # Keep the MSVC runtime selection consistent with the parent project.
+  set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
+  set(INSTALL_GTEST OFF CACHE BOOL "" FORCE)
+  FetchContent_Declare(
+    googletest
+    URL https://github.com/google/googletest/archive/refs/tags/v1.14.0.tar.gz
+    URL_HASH SHA256=8ad598c73ad796e0d8280b082cebd82a630d73e73cd3c70057938a6501bba5d7
```
IvanaGyro: We have removed all Internet access in cmake file and make HPTT and LAPACKE finder as submodule. #944 

Listing gtest as a requirement for running test in the document avoids we add more dependencies' links in cmake script, which may developer hard to trace where is the dependencies actually come from.

tests/README.md
```diff
@@ -0,0 +1,43 @@
+# Cytnx C++ test suite
+
+## Building
+
+```bash
+git submodule update --init --recursive   # required once per fresh clone/worktree
+cmake -S . -B build -DRUN_TESTS=ON
+cmake --build build --target test_main
+```
```
IvanaGyro, line 25: Consider to put general build instruction in other places.

Manually building without using presets is not recommended. It's hard to maintain the configurations out of predefined presets. The build folder will not be `build` if using presets.

tests/CMakeLists.txt
yingjerkao: Agreed — dropped the FetchContent fallback in bdfce0e5 to keep the CMake config network-free per #944. `tests/CMakeLists.txt` now uses `find_package(GTest REQUIRED)` (plus an explicit `GTest::gmock` target check, since find_package can resolve gtest without gmock) and errors out with install guidance (brew / apt / conda) when it is missing. GoogleTest is now documented as a prerequisite in `tests/README.md` rather than fetched.

tests/README.md
yingjerkao: Updated in bdfce0e5 — `tests/README.md` no longer teaches a manual `-S/-B` build; it points at the CMake presets (the `debug-*` presets enable `RUN_TESTS`) and notes each preset builds into its own `build/<preset>/` directory. The test-specific bits (running suites, the macOS `DYLD_LIBRARY_PATH` warning) stay.

tests/CMakeLists.txt
```diff
@@ -2,6 +2,25 @@
...
+  )
+  FetchContent_MakeAvailable(googletest)
+endif()
```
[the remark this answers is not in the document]
yingjerkao: Good catch on the `add_library ... ALIAS GTest::gtest ... already exists` hazard. Rather than the `FIND_PACKAGE_ARGS` route, we dropped the FetchContent fallback entirely (project policy #944 keeps the CMake configuration free of network access), so the find_package + FetchContent combination that could collide no longer exists — `find_package(GTest REQUIRED)` with an explicit `GTest::gmock` check is used instead. Fixed in bdfce0e5.

pcchen:
## Code Review — LGTM

Four portability fixes so the `RUN_TESTS=ON` suite builds on macOS/AppleClang (was Linux/GCC-only): `int→uint` loop vars to avoid `-Wc++11-narrowing` in braced-init APIs, a `fill(i+1)` cast, `<bits/stdc++.h>`→concrete headers, and proper googletest provisioning via `find_package(GTest)`.

### Verified
- **The bulk `cytnx_int64 → cytnx_uint64` swaps are safe.** Every converted loop is a forward `for (uint i = 0; i < literal; i++)` with small literal bounds. I scanned all added *and* context lines for the unsigned hazards (`i >= 0`, `i--`, `i - 1`) and found none live: the only `at({i-1,...})` uses are **commented out** (matches the PR's audit claim), the `permute(a, -1)` hits are rowrank args, and the `per(i,a,n)` reverse-loop macro stays `int`. No underflow / infinite-loop risk. ✅
- **`fill(i+1)` → `fill(static_cast<cytnx_uint64>(i+1))`** — correct, resolves the macOS `size_t`-dispatch failure. ✅
- **CMake googletest change is correct and thoughtful:** `find_package(GTest REQUIRED)` + an explicit `if(NOT TARGET GTest::gmock) FATAL_ERROR` with brew/apt/conda guidance (closing the "gtest found without gmock" gap), namespaced `GTest::` link targets, and no FetchContent — respecting #944's no-network rule. Dropping the dead `gtest_force_shared_crt` line (set after `add_subdirectory` so it never took effect) is correct cleanup. ✅
- **`<bits/stdc++.h>` → concrete headers** is a reasonable set and build-verified. ✅
- **CI is green on all platforms including macOS ×2** (`BuildAndTest-macos-latest` + `-15-intel`) — the target platform now builds + tests. ✅
- Adds `tests/README.md` documenting GoogleTest as a prerequisite.

### Findings
None blocking — a mechanical, well-audited, CI-verified portability fix. (Slightly behind master but `MERGEABLE`.)

**LGTM.**

pcchen (approved):
Approving. Clean, well-audited macOS/AppleClang portability fix. The bulk `cytnx_int64 → cytnx_uint64` loop-var swaps are safe (all forward loops with literal bounds; the only `at({i-1,...})` uses are commented out; the `per` reverse-loop macro stays `int`), the `fill(i+1)` cast and `<bits/stdc++.h>`→concrete-headers changes are correct, and the CMake googletest provisioning is modern and #944-compliant (`find_package(GTest REQUIRED)` + explicit `GTest::gmock` guard with install guidance + namespaced targets, no FetchContent). CI is green on all platforms including both macOS runners — the target of the fix.

PR #973, opened 2026-07-05
fix(tests): compile the test suite warning-clean on AppleClang

yingjerkao:
## Summary

`cmake -DRUN_TESTS=ON` fails to build on macOS because AppleClang treats `-Wc++11-narrowing` as a hard error, while CI's GCC only warns for same-width signed→unsigned narrowing in braced initializer lists. This PR fixes all such sites in the test sources by using the correct unsigned index types (no behavior change — every converted variable counts up from 0 or 1 and is used only as a `vector<cytnx_uint64>` locator/shape element), plus two additional macOS-only compile blockers that surfaced once clang got past its error limit.

Relates to #857 (test-hygiene work).

## Changes

- **tests/BlockUniTensor_test.cpp** — ~99 index-loop counters (and the `offset` accumulator) `cytnx_int64` → `cytnx_uint64`; they feed `at()`, `elem_exists()`, and `_fx_locate_elem()` locators, all of which take `std::vector<cytnx_uint64>`.
- **tests/linalg_test/Lanczos_Gnd_test.cpp** — `CreateLRO`/`Create_UTinit` shape parameters and one local `D, d` pair `int` → `cytnx_uint64` (feed `UniTensor::uniform` shape lists).
- **tests/linalg_test/Vectordot_test.cpp** — `InitTensor` length parameter and its summation loop counter `int` → `cytnx_uint64`.
- **tests/linalg_test/linalg_test.h** — drop wrong-direction `(cytnx_int64)` casts in a `zeros()` shape list; `n` is already `cytnx_uint64` and `zeros` takes `vector<cytnx_uint64>`.
- **tests/BlockUniTensor_test.h** — cast the `fill(i + 1)` argument to `cytnx_double`: on macOS `size_t` is `unsigned long`, which matches none of the nine `fill()` overloads exactly, making the call ambiguous (on Linux `size_t` == `cytnx_uint64`, so CI never sees this).
- **tests/utils/getNconParameter.h** — replace the GCC-internal `<bits/stdc++.h>` umbrella header (absent in AppleClang/libc++) with the standard headers actually used.

## Verification (macOS arm64, AppleClang, `-DRUN_TESTS=ON -DBUILD_PYTHON=OFF -DUSE_MKL=OFF -DUSE_HPTT=OFF -DUSE_CUDA=OFF`)

- Full `test_main` build from a clean configure completes with zero warnings/errors in the test sources.
- Affected suites pass: BlockUniTensorTest 66/66, Vectordot 6/6, NconTest 3/3; full suite excluding ARPACK-dependent families: **982 passed / 4 failed / 23 skipped of 997**.
- The 4 failures (`linalg_Test.BkUt_{Svd,Gesvd}_truncate_return_err_*`) and the segfaults in every `linalg::Lanczos`/`Arnoldi` test (null deref inside ARPACK `dsaupd_`) are pre-existing macOS runtime issues on untouched code paths — previously invisible because the suite never compiled on macOS. They likely deserve their own issues.

### DIFF
--- a/tests/BlockUniTensor_test.cpp
+++ b/tests/BlockUniTensor_test.cpp
@@ -121,9 +121,9 @@ TEST_F(BlockUniTensorTest, is_blockform) {
 }
 TEST_F(BlockUniTensorTest, clone) {
   UniTensor cloned = UT_pB_ans.clone();
-  for (cytnx_int64 i = 0; i < 5; i++)
-    for (cytnx_int64 j = 0; j < 9; j++)
-      for (cytnx_int64 k = 1; k < 30; k++) {
+  for (cytnx_uint64 i = 0; i < 5; i++)
+    for (cytnx_uint64 j = 0; j < 9; j++)
+      for (cytnx_uint64 k = 1; k < 30; k++) {
         EXPECT_EQ(cloned.at({i, j, k}).exists(), UT_pB_ans.at({i, j, k}).exists());
         if (cloned.at({i, j, k}).exists()) EXPECT_EQ(cloned.at({i, j, k}), UT_pB_ans.at({i, j, k}));
       }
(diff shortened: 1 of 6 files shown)


yingjerkao:
Closing as superseded by #972, which contains the same fixes (narrowing conversions, the `fill()` overload ambiguity, `<bits/stdc++.h>`) and additionally covers the GPU test tree, `generate_networks.cpp`, and googletest provisioning via `find_package`/FetchContent — the latter properly fixes the `gtest_main` link failure this PR did not address.

The macOS verification results from this PR remain relevant to #972: with the test suite building on macOS arm64 (AppleClang, conda-forge OpenBLAS/ARPACK), 982 of 997 non-ARPACK tests pass; the pre-existing runtime failures are tracked in #974 (ARPACK `dsaupd_` segfaults) and #975 (Svd/Gesvd_truncate `return_err` degeneracy ties, fix in #977).

Issue #975, opened 2026-07-05
macOS: BkUt_{Svd,Gesvd}_truncate return_err tests fail — exact-degeneracy tie at the cut makes kept count platform-dependent


yingjerkao:
## Summary

With the C++ test suite now compiling on macOS (#973), four tests fail there that pass on Linux CI:

- `linalg_Test.BkUt_Svd_truncate_return_err_returns_discarded_values`
- `linalg_Test.BkUt_Gesvd_truncate_return_err_returns_discarded_values`
- `linalg_Test.BkUt_Svd_truncate_return_err_one_returns_first_discarded_value`
- `linalg_Test.BkUt_Gesvd_truncate_return_err_one_returns_first_discarded_value`

Failure mode (`tests/linalg_test/linalg_test.cpp:50`):

```
Expected equality of these values:
  trunc[0].shape()[0]
    Which is: 6
  5
```

`Svd_truncate(svd_T, keepdim=5, ...)` keeps **6** singular values on macOS arm64 (conda-forge OpenBLAS) instead of the asserted 5.

## Diagnosis

Keeping more than `keepdim` values is *documented intended behavior* when singular values at the cut are exactly degenerate — see the degeneracy note on the `min_blockdim` overload of `Svd_truncate` in `include/linalg.hpp` (block scans keep everything `>= Smin`; example given there: spectrum (1 2 2 2 2 3) with keepdim=3 keeps 5).

The `Svd_truncate1.cytnx` fixture is Float32 with a near-degenerate multiplet at the top of the spectrum; whether adjacent values round to bitwise-equal floats depends on the BLAS/platform. On macOS arm64 + OpenBLAS the tie occurs at the cut, triggering the documented enlargement; on Linux CI it does not. So:

1. The tests are over-strict: asserting an exact kept count (`ASSERT_EQ(trunc[0].shape()[0], 5)`) is platform-dependent under the documented degeneracy contract. They should assert `kept >= keepdim` plus tie/consistency conditions (discarded count == total − kept).
2. There is also a genuine off-by-tie inconsistency in the `return_err` discarded-values list when the enlargement triggers (the discarded tail overlaps the kept values).

A fix addressing both (test contract + `return_err` discarded-count consistency under ties in `src/linalg/Svd_truncate.cpp`, `Gesvd_truncate.cpp`, and `block_truncation_helpers.hpp`) is already drafted locally and should follow as a PR.

## Environment / reproduction

[...]

PR #976, opened 2026-07-05
fix(Tensor): harden reshape validation (multiple -1, 0-dim with -1, strong exception guarantee)

yingjerkao:
## Summary

Two commits hardening `Tensor_impl::reshape_` / `reshape` input validation:

- **Reject multiple `-1` dimensions.** The duplicate-`-1` guard tested `new_shape[i] != -1` (always false inside the negative branch) instead of `has_undetermine`, so `t.reshape({-1, -1})` was silently accepted and left a `-1` in the shape.
- **Guard division by zero for `{0, -1}`-style shapes.** With a `-1` present, the inferred dimension was computed as `storage.size() % new_N` / `/ new_N` where `new_N` is the product of the non-negative dims — a 0-sized dim made `new_N == 0`, i.e. integer division/modulo by zero (UB; SIGFPE on x86). Inferring `-1` alongside a 0-sized dimension is now rejected with a clear error.
- **Strong exception guarantee.** Validation previously ran after `_shape` was already overwritten, so a throwing `reshape_` left the tensor with inconsistent metadata (e.g. shape `{0, 18446744073709551615}` after `reshape_({0, -1})` while the storage still held 12 elements). All checks now run before any member is mutated; a failed reshape leaves the tensor unchanged.
- The duplicate-`-1` error message now says "reshape can have only one dimension specified as -1" instead of reusing the generic message.

## Tests

Three new gtests in `tests/Tensor_test.cpp` (all written red-first):

- `Tensor.ReshapeRejectsMultipleUnknownDims`
- `Tensor.ReshapeRejectsZeroDimWithUnknownDim`
- `Tensor.FailedReshapeLeavesShapeUnchanged`

Full suite on macOS/AppleClang: 1066 passed; the only failures are the 4 pre-existing `linalg_Test.*_return_err_*` SVD-truncate failures tracked in #975 (verified identical without this change).

### DIFF
--- a/include/backend/Tensor_impl.hpp
+++ b/include/backend/Tensor_impl.hpp
@@ -254,48 +254,48 @@ namespace cytnx {
     }
 
     void reshape_(const std::vector<cytnx_int64> &new_shape) {
-      if (!this->_contiguous) {
-        this->contiguous_();
-      }
-      // std::vector<cytnx_uint64> result_shape(new_shape.size());
+      // validate the whole request before mutating any member so a throwing
+      // reshape_ leaves the tensor unchanged
       cytnx_uint64 new_N = 1;
       bool has_undetermine = false;
-      unsigned int Udet_id = 0;
-      // this->_shape = vec_cast<cytnx_int64,cytnx_uint64>(new_shape);
-      this->_shape.resize(new_shape.size());
+      cytnx_uint64 Udet_id = 0;
       for (cytnx_uint64 i = 0; i < new_shape.size(); i++) {
-        this->_shape[i] = new_shape[i];
-      }
-      for (int i = 0; i < new_shape.size(); i++) {
         if (new_shape[i] < 0) {
           cytnx_error_msg(
             new_shape[i] != -1, "%s",
             "[ERROR] reshape can only have dimension > 0 and one undetermine rank specify as -1");
-          cytnx_error_msg(
-            has_undetermine, "%s",
-            "[ERROR] reshape can only have dimension > 0 and one undetermine rank specify as -1");
+          cytnx_error_msg(has_undetermine, "%s",
+                          "[ERROR] reshape can have only one dimension specified as -1");
           Udet_id = i;
           has_undetermine = true;
         } else {
           new_N *= new_shape[i];
-          // result_shape[i] = new_shape[i];
         }
       }
 
       if (has_undetermine) {
+        cytnx_error_msg(
+          new_N == 0, "%s",
+          "[ERROR] reshape cannot infer the -1 dimension when another dimension is 0");
         cytnx_error_msg(new_N > this->_storage.size(), "%s",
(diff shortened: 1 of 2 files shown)


pcchen:
## Code Review

Hardens `Tensor_impl::reshape_` validation — three real bugs, all fixed correctly.

### Verified correct (traced each)
1. **Duplicate `-1` now rejected.** The old guard `cytnx_error_msg(new_shape[i] != -1, …)` was always false inside the `new_shape[i] < 0` branch, so it never fired and `reshape({-1,-1})` silently left a `-1` (→ `18446…615`) in the shape. Now it's `cytnx_error_msg(has_undetermine, …)`, which correctly throws on the second `-1`. ✅
2. **Div-by-zero (`{0,-1}`) guarded.** `cytnx_error_msg(new_N == 0, …)` runs *before* the `storage.size() % new_N` / `/ new_N`, closing the old SIGFPE/UB path — for both `{0,-1}` and `{-1,0}` orderings. ✅
3. **Strong exception guarantee.** All validation now precedes any mutation (`_shape.resize`/fill, `contiguous_()`, `_mapper`). Moving `contiguous_()` after validation is safe — it doesn't change `storage.size()`, which is all the checks read. A failed reshape leaves the tensor untouched. ✅

### No collateral
The fix is targeted — a genuine 0-sized reshape *without* `-1` (e.g. `{0}` on an empty tensor) still passes the `new_N != storage.size()` path; only `0`-dim *with* `-1` is rejected. Valid `{3,-1}` → `{3,4}` preserved. And it's the single impl: `Tensor_impl::reshape` clones + calls `reshape_`, and all `Tensor::reshape_` overloads (int64/uint64/initializer_list) forward here, so every entry point is covered.

### Tests
Three red-first gtests covering multiple-`-1`, `0`+`-1` (both orders), and shape-unchanged-on-failure. Good coverage; CI green on all platforms (Linux + macOS ×2).

### Action item
🟠 The PR is **`CONFLICTING` with master and needs a rebase** before merge — its green CI is from the pre-conflict commit, so a fresh run on the rebased commit is worth confirming (the change is small and self-contained, so the rebase should be straightforward).

No findings on the code itself — clean, correct, well-tested hardening. **LGTM once rebased.**

pcchen (approved):
Approving after a fresh re-review of the rebased head (`2749ffdf`). The rewritten `reshape_` is a strict validate-then-mutate design: the duplicate-`-1` guard now actually fires (old guard was tautologically false), the `{0,-1}`/`{-1,0}` division-by-zero path is closed before any `%`/`/`, and the strong exception guarantee holds (all checks read only `new_shape` + `storage.size()`; every mutation, including `contiguous_()`, happens post-validation). No collateral to valid reshapes, and the single impl covers all reshape entry points. The 3 red-first tests map one-to-one onto the fixes, and fresh CI on this exact head is fully green (16 pass — BuildAndTest on Linux + both macOS, DownstreamFindPackage). No findings.

PR #978, opened 2026-07-05
fix(Storage): check dtype in at<T> unconditionally (#965)

### LINKED ISSUE #965: Storage::at<T>() and Tensor::item<T>() silently reinterpret mismatched dtype in normal builds
## Problem

`Tensor::item<T>()` and `Storage::at<T>()` can silently reinterpret storage as the wrong C++ type in normal builds.

For example, if a tensor has dtype `Type.Float`, then this should be a type error:

```cpp
auto t = cytnx::zeros({1}, cytnx::Type.Float);
t(0) = 1.0;
auto x = t.item<double>();
```

Instead, in a normal build it can return garbage because the implementation eventually does:

```cpp
return static_cast<double *>(this->data())[idx];
```

on a `Float` storage buffer.

The dtype mismatch checks in `Storage_base::at<T>()` are guarded by `cytnx::User_debug`, so they are disabled in ordinary builds. This contradicts the `Tensor::item<T>()` documentation, which says a type mismatch will issue an error.

## Why this matters

This is very easy to trip over in tests and C++ code. A one-element `Float` tensor read with `item<double>()` can produce meaningless values rather than failing loudly. The same issue applies to other typed `at<T>()` specializations where the requested C++ type does not match the storage dtype.

## Expected behavior

Typed raw accessors should check that the requested C++ type matches the storage dtype unconditionally, not only when `User_debug` is enabled.

Unchecked raw reinterpretation should not be the behavior of `Tensor::item<T>()` or `Storage::at<T>()`; if an unchecked path is needed internally, it should be explicitly named as such and kept out of public/user-facing typed access.

yingjerkao:
## Summary

Fixes #965.

- **Unconditional dtype check in `Storage_base::at<T>`.** Every `at<T>` specialization wrapped its dtype-mismatch check in `if (cytnx::User_debug)` (default `false`), so e.g. `Tensor({1}, Type.Float).item<double>()` reinterpreted a float buffer as double and returned garbage in normal builds. The checks now run unconditionally, matching `data<T>()` and `back<T>()`, which already checked unconditionally. Mismatched `at<T>`/`item<T>` calls now throw `std::logic_error` instead of silently type-punning.
- **Review follow-up:** the out-of-bound error in the same `at<T>` specializations formatted `cytnx_uint64` index and `unsigned long long` size through `%d` (garbage for values ≥ 2³¹, UB in the formatter) — now `%llu` with explicit casts.

## Test fallout fixed (call sites relying on the old silent behavior)

- `tests/linalg_test/Arnoldi_Ut_test.cpp`: `linalg::Max(block).item<double>()` on `linalg::Eig` output (ComplexDouble) — now reads `.item().real()`.
- `tests/DenseUniTensor_test.cpp` (`at` test): the const `cut` tensor was constructed once with default dtype but read as every dtype in the loop — now cloned per-dtype.

## Tests

- `Storage.AtDtypeMismatchThrows`, `Tensor.ItemDtypeMismatchThrows` — written red-first against the silent-garbage behavior.
- `StorageTest.AtDtypeMismatchThrowsForAllDtypes` — exhaustive 11×11 dtype/T matrix, iterating the shared `TestTools::dtype_list`.
- `StorageTest.AtOutOfBoundMessageReportsIndexAndSize` — pins the bounds-error message content.

Full suite on macOS/AppleClang: 1067 passed; the only failures are the 4 pre-existing `linalg_Test.*_return_err_*` SVD-truncate failures tracked in #975 (verified identical without this change).

A follow-up branch (in progress) consolidates the 11 copy-pasted `at<T>`/`back<T>`/`data<T>` specializations into one template body via `cy_typeid_v<T>` so no future specialization can drop the check again.

### DIFF
--- a/src/backend/Storage_base.cpp
+++ b/src/backend/Storage_base.cpp
@@ -648,11 +648,9 @@ namespace cytnx {
   //====================================================
   template <>
   float &Storage_base::at<float>(const cytnx_uint64 &idx) const {
-    if (cytnx::User_debug) {
-      cytnx_error_msg(this->dtype() != Type.Float,
-                      "[ERROR] type mismatch. try to get <float> type from raw data of type %s",
-                      Type.getname(this->dtype()).c_str());
-    }
+    cytnx_error_msg(this->dtype() != Type.Float,
+                    "[ERROR] type mismatch. try to get <float> type from raw data of type %s",
+                    Type.getname(this->dtype()).c_str());
     if (idx >= this->size())
       cytnx_error_msg(true, "[ERROR] index [%d] out of bound [%d]\n", idx, this->size());
 
@@ -664,11 +662,9 @@ namespace cytnx {
 
   template <>
   double &Storage_base::at<double>(const cytnx_uint64 &idx) const {
-    if (cytnx::User_debug) {
-      cytnx_error_msg(this->dtype() != Type.Double,
-                      "[ERROR] type mismatch. try to get <double> type from raw data of type %s",
-                      Type.getname(this->dtype()).c_str());
-    }
+    cytnx_error_msg(this->dtype() != Type.Double,
+                    "[ERROR] type mismatch. try to get <double> type from raw data of type %s",
+                    Type.getname(this->dtype()).c_str());
     if (idx >= this->size())
       cytnx_error_msg(true, "[ERROR] index [%d] out of bound [%d]\n", idx, this->size());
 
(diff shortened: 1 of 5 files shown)


yingjerkao:
Heads-up: [#979](https://github.com/Cytnx-dev/Cytnx/pull/979) is stacked on this branch — it refactors the 11 copy-pasted `at<T>`/`back<T>`/`data<T>` specializations in `Storage_base.cpp` into single template bodies keyed on `cy_typeid_v<T>`, so the dtype check this PR fixes can't drift per-type again. It should be retargeted to `master` and merged after this one.

IvanaGyro (approved):

Issue #906, opened 2026-06-16
Scalar in-place Tensor arithmetic detaches storage-sharing views


ianmccul:
Opened by Codex on behalf of @ianmccul.

## Summary

`Tensor` has Python-like shallow assignment semantics: copying or assigning a `Tensor` copies the `Tensor_impl` intrusive pointer, while `clone()` is the explicit deep-copy operation. At the same time, several operations behave like views by creating a new `Tensor_impl` that shares the same underlying `Storage`, notably `permute()` and contiguous `reshape()`.

However, scalar in-place arithmetic does not preserve this storage sharing. For example, `a += 1.0` computes a fresh result tensor and replaces `a._impl->storage()` with the result storage. Any other `Tensor` view that shared the old storage continues to point at the old storage, so it silently detaches from `a`. The same applies when the RHS is `cytnx::Scalar` or `Scalar::Sproxy`.

This is inconsistent with Tensor-vs-Tensor in-place arithmetic, which uses the `iAdd`/`iSub`/`iMul`/`iDiv` kernels and mutates the existing storage in place.

## Why this is a bug

If `permute()` and contiguous `reshape()` are intended to be storage-sharing views, then in-place mutation should mutate the shared storage. In-place operations should not silently break the relationship between view-like tensors.

The current behavior means that these two operations can have different aliasing semantics:

```cpp
auto x = cytnx::ones({1}, cytnx::Type.Double);
a += x;    // shape-{1} rank-1 Tensor RHS: mutates shared storage
a += 1.0;  // scalar RHS: replaces a's Storage and detaches storage-sharing views
```

The same issue applies to scalar `-=`, `*=`, and `/=`. In this context, "scalar" includes built-in numeric scalar types, `cytnx::Scalar`, and `Scalar::Sproxy`.

## Minimal example

In C++:

```cpp
auto a = cytnx::arange(0, 6, 1, cytnx::Type.Double).reshape({2, 3});
auto b = a.permute({1, 0});

// b is a different Tensor_impl but shares a's Storage.
// This should preserve the storage-sharing view relationship.
a += 1.0;

// Expected: b observes the updated values.
// Actual: b still points at the old Storage, while a now points at a new Storage.
```

The same behavior should be visible through the Python API:

```python
a = cytnx.arange(0, 6, 1, dtype=cytnx.Type.Double).reshape(2, 3)
b = a.permute(1, 0)

a += 1.0

# Expected: b observes the updated values.
[...]

Issue #935, opened 2026-06-23
Replace `Scalar` raw-pointer polymorphism with value semantics


ianmccul:
## On the topic of removing technical debt

`cytnx::Scalar` is currently implemented as a raw owning pointer to a runtime-polymorphic `Scalar_base` hierarchy:

```c++
class Scalar {
 public:
  Scalar_base *_impl;
  ...
};
```

Copies are deep copies:

```c++
Scalar(const Scalar &rhs) : _impl(new Scalar_base()) {
  if (this->_impl != nullptr) delete this->_impl;
  this->_impl = rhs._impl->copy();
}

Scalar &operator=(const Scalar &rhs) {
  if (this->_impl != nullptr) delete this->_impl;
  this->_impl = rhs._impl->copy();
  return *this;
}
```

and destruction is manual:

```c++
~Scalar() {
  if (this->_impl != nullptr) delete this->_impl;
}
```

This is not Python-style reference binding semantics. It is also not reference-counted. A Python `cytnx.Scalar` object owns a C++ `Scalar`, and that C++ `Scalar` owns a unique heap-allocated `Scalar_base`. Copying a `Scalar` copies the scalar value.

This design is effectively a hand-written `std::variant`, but with heap allocation, virtual dispatch, duplicated conversion methods, a dtype-indexed factory table, and manual memory management.

## Related but separate: `Sproxy`

There is a separate proxy type:

```c++
struct Scalar::Sproxy {
  boost::intrusive_ptr<Storage_base> _insimpl;
  cytnx_uint64 _loc;
  ...
};
```

That proxy really is a reference into tensor storage. It must remain a proxy or view-like object, because assignment through it mutates storage. But this does not imply that standalone `Scalar` needs pointer identity or reference-counted semantics. `Sproxy` and `Scalar` have different jobs:

- `Sproxy`: a reference-like proxy to an element in `Storage`.
- `Scalar`: an owning scalar value.

The current implementation conflates the value case with a polymorphic heap object, even though standalone `Scalar` behaves like a value.

## Consequences

The current design causes unnecessary complexity and risk:

- Every scalar construction allocates on the heap.
- Copy construction and assignment allocate and dispatch through virtual `copy()`.
- `Scalar` manually manages ownership with `new` and `delete`.
- `Scalar_init_interface` exists only to map runtime dtype integers to heap factories.
[...]

PR #980, opened 2026-07-05
fix(Tensor): scalar in-place arithmetic mutates storage in place

yingjerkao:
## Problem

Part of #906. `t += 1.0` (and `-=`, `*=`, `/=` with any scalar or `cytnx::Scalar`) replaced the tensor's Storage with a freshly computed one:

```cpp
this->_impl->storage() = cytnx::linalg::Add(*this, rc)._impl->storage();
```

Any view sharing that storage (from `permute()`, contiguous `reshape()`, `from_storage()`) silently detached, while `t += tensor` mutated in place — the two paths disagreed on aliasing semantics.

## Fix

Scalar RHS now routes through the same `iAdd/iSub/iMul/iDiv` kernels via a shape-`{1}` wrapper tensor (the kernels already broadcast a length-1 RHS and mutate LHS storage in place). All 48 scalar specializations rewritten; the `Tensor`/`Tproxy`/`Sproxy` specializations are untouched.

## Deliberate behavior changes

- Views sharing storage stay attached (the bug fix).
- LHS dtype is preserved: `Float tensor += 1.0` stays `Float` (previously promoted to `Double`); integer LHS truncates fractional scalars — both matching the existing tensor-RHS in-place behavior and numpy's scalar rules.
- `real_tensor ⊙= complex_scalar` now throws instead of silently promoting the tensor to complex.

The GPU `iAdd` path can still rebind on promotion; that remains tracked under #906.

## Testing

Seven new gtests: storage-sharing preserved for all four ops (including through a real non-contiguous `permute()` view), dtype preservation, integer truncation pinned, `real ⊙= complex` throws for all four ops, and `cytnx::Scalar` RHS. Red-state verified against the old code (a shared view read `0` after `+=` — update lost). Full suite passes at baseline.

### DIFF
--- a/src/Tensor.cpp
+++ b/src/Tensor.cpp
@@ -546,6 +546,25 @@ namespace cytnx {
     return out;
   }
 
+  namespace {
+    // Wrap a scalar as a shape-{1} Tensor on `device` so scalar in-place
+    // arithmetic reuses the iAdd/iSub/iMul/iDiv kernels (which broadcast a
+    // length-1 RHS and mutate LHS storage in place). See #906.
+    template <class T>
+    Tensor _scalar_as_rank1_tensor(const T &rc, const int device) {
+      Tensor s({1}, Type.cy_typeid(rc), Device.cpu);
+      s.storage().at<T>(0) = rc;
+      if (device != Device.cpu) s = s.to(device);
+      return s;
+    }
+    Tensor _scalar_as_rank1_tensor(const Scalar &rc, const int device) {
+      Tensor s({1}, rc.dtype(), Device.cpu);
+      s.item() = rc;  // Sproxy assignment
+      if (device != Device.cpu) s = s.to(device);
+      return s;
+    }
+  }  // namespace
+
   ///@cond
   // +=
   template <>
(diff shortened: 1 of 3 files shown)


src/Tensor.cpp
```diff
@@ -546,6 +546,25 @@ namespace cytnx {
     return out;
   }
 
+  namespace {
+    // Wrap a scalar as a shape-{1} Tensor on `device` so scalar in-place
+    // arithmetic reuses the iAdd/iSub/iMul/iDiv kernels (which broadcast a
+    // length-1 RHS and mutate LHS storage in place). See #906.
+    template <class T>
+    Tensor _scalar_as_rank1_tensor(const T &rc, const int device) {
```
IvanaGyro, line 554: This can be moved to linalg to benefit the linalg functions. If the naming style in the current file is messy, the name of the new function should follow the coding style and avoid prefixing with `_`.

IvanaGyro (dismissed):

src/Tensor.cpp
ianmccul: Perhaps a good opportunity make scalar as rank 0 tensors?  It would make operations such as trace much more consistent.

pcchen:
## Code Review

Fixes the real #906 bug: `t += 1.0` did `storage() = Add(*this, rc).storage()` — computing out-of-place at a *promoted* dtype and swapping the Storage handle, silently detaching every view sharing that storage — while `t += tensor` mutated in place. All 48 scalar specializations (+4 `Scalar` overloads) now route through the in-place `iAdd/iSub/iMul/iDiv` kernels via a shape-`{1}` wrapper. **LGTM once rebased.**

### Correctness — verified
- **The `iOp` + shape-`{1}` route is sound:** `iDiv` et al. explicitly special-case a length-1 RHS (`iDiv.cpp:17` skips the shape check for `{1}`) and dispatch in-place on LHS storage — contiguous and non-contiguous paths — with an aliasing guard (irrelevant here; the wrapper is always fresh). Same established path `DenseUniTensor::normalize_` uses. ✅
- **`_scalar_as_rank1_tensor`** is correct for CPU + GPU (build on CPU, set element, `.to(device)`); `Tensor::from_storage` used by the tests exists (`Tensor.hpp:558`). ✅
- **Perf is a side win:** the old path allocated a full N-element result per scalar op; the new one allocates one element.
- **Tests are strong:** storage-sharing pinned via `from_storage` (with a sharp comment on why copy-ctor aliasing couldn't observe the bug) *and* through a real non-contiguous `permute()` view; dtype preservation; integer truncation; `real⊙=complex` throws ×4; `Scalar` RHS; corrected error text. Red-state verified; CI on the pre-conflict commit: 15 pass, only advisory codecov red. ✅

### Findings
1. **🟠 `CONFLICTING` — trivial rebase.** Checked the merge-base: `src/Tensor.cpp` and `iArithmetic_visit.hpp` have **no** master-side changes; the only collision is the trailing test-append in `tests/Tensor_test.cpp` (vs. #976's reshape tests). I'll resolve it (keep both), as with #976/#981.
2. **🟡 Changelog entry needed** for the deliberate behavior changes: dtype no longer promotes (`Float += 1.0` stays Float; `Int64 += 2.7` truncates) and `real ⊙= complex` throws instead of promoting. numpy-consistent and tensor-RHS-consistent, but downstream code relying on silent promotion will change results.
3. **🟢 Positive cross-PR interactions:** making `operator/=<double>`/`<Scalar>` dtype-preserving fixes the `/=`-half of the Float→Double Lanczos regression flagged on #1000 (the out-of-place `clone()/norm()` sites there still promote), and gives Block/BlockFermionic `normalize_` (`block /= Scalar`) the dtype-preservation guarantee that previously only Dense had.

### On the three inline comments
- **@IvanaGyro** (move helper to linalg; avoid the `_` prefix): both fair, non-blocking. The helper is genuinely reusable (Ger's implicit alpha, the `normalize_` divisors); relocating it to linalg forces the rename anyway. Fine here or as a fast-follow.
- **@ianmccul** (rank-0 scalar tensors): good design question, bigger than this PR — spun off as **#1016** with the options laid out (true rank-0 / document the `{1}` convention / fold into #935). This PR's `{1}` wrapper is forward-compatible either way — it's the single switch site if rank-0 lands.
- **@gemini-code-assist** (GPU H2D-per-op bottleneck): valid observation, wrong severity — the *old* path allocated a full **N-element** GPU tensor + ran an out-of-place kernel per scalar op; the new path allocates **one element** + one H2D copy, strictly cheaper for any nontrivial tensor. The per-call GPU overhead is already tracked in **#988**, and the real fix (dedicated scalar-argument kernels) is #906 follow-up territory. **One caution on the caching suggestion:** a pre-allocated/cached scalar tensor would be mutable shared state — a data race now that #987 releases the GIL around these calls. Any caching must be thread-local or immutable-per-value; better to not cache and go straight to scalar kernels.

src/Tensor.cpp
IvanaGyro: That will make Scalar heavier than a simple std::variant and we may have to triage scalar in all operations consuming tensors

ianmccul: No, I meant rank 0 tensor should exist and be "scalar + shared storage + device"; see comment https://github.com/Cytnx-dev/Cytnx/issues/1016#issuecomment-4910913995

- Scalar is similar to to Python builtin number
- Rank-0 tensor is similar to numpy rank 0 ndarray

pcchen: Proposal: defer this to the outcome of **#1016** rather than changing it here. Where the scalar-wrapper helper should live (and what it's named) is entangled with the "rank-0 tensor vs. shape-`{1}` convention" design question now tracked there — #1016's option B explicitly includes exactly the relocation you suggested (a shared `linalg`-level `scalar_tensor()` helper as the single implementation point, no `_` prefix), while option A would make this helper obsolete altogether. Two stacked PRs also touch this same helper (#992 makes it CPU-resident to fix a GPU UB; #1015's typed-storage direction may reshape it again), so moving it now would just multiply churn.

If that's acceptable, could you re-review/dismiss the requested-changes gate so #980 + #992 (the CUDA use-after-free / UB fix stacked on this) can land? The rename+relocate then happens once, in its final form, per the #1016 decision.

pcchen: Good concrete argument — captured in **#1016** (the design-question issue for rank-0 vs. the shape-`{1}` convention), where it weighs for option B/C (keep `Scalar` a lightweight variant, as #1011 builds) against option A (true rank-0 tensors). Let's keep the design debate there so it accumulates in one place; no change on this PR.

pcchen (approved):
Approving. The fix is verified: all 48 scalar specializations (+4 `Scalar` overloads) now route through the in-place `iAdd/iSub/iMul/iDiv` kernels via a shape-`{1}` wrapper — the kernels explicitly special-case a length-1 RHS and mutate LHS storage in place, so storage-sharing views stay attached (the #906 bug), and the old full-N-element allocation per scalar op is gone. Dtype preservation, integer truncation, and the `real ⊙= complex` rejection are all pinned by red-state-verified tests, including through a genuine non-contiguous `permute()` view. CI is functionally green on the rebased head (only advisory codecov red).

Two coordination notes, both agreed in the threads:
1. **Land #992 immediately after this** — on CUDA, this PR's `.to(device)` wrapper is read host-side by the GPU kernels (UB); #992 (stacked on this branch, hardware-validated) fixes that plus the non-contiguous GPU scalar regression. The pair should merge back-to-back.
2. **The helper relocation/rename requested in review is deferred to #1016** (the rank-0-vs-`{1}`-convention design issue) rather than churned here — it's exactly option B's "single implementation point" helper there, and #992/#1015 touch the same code. The rename+move happens once, in its final form, per that decision.

Behavior changes (dtype no longer promotes; `real ⊙= complex` throws) are deliberate, numpy-consistent, and need a changelog entry alongside the other in-place-semantics rulings.

PR #981, opened 2026-07-05
fix: drop C complex.h from pybind TUs; rename macro-colliding template param

### LINKED ISSUE #951: Macro 'I' in complex.hpp overwrites static_assert(I ...) in type.hpp
The `complex.h` header defines a macro `I` (`#define I _Complex_I`), which then overwrites `Type.hpp`'s template `<std::size_t I, ...>`. So `complex.h` must be included last, after `Type.hpp` or `cytnx.hpp`.

Suggestion: rename `I` into something more meaningful in `Type.hpp`.

It seems that `complex.h` is not needed anyways since we use our own files. The only place where it seems relevant is `src/backend/linalg_internal_cpu/Det_internal.cpp`, which uses `<complex.h>`: it references `double _Complex` / `float _Complex` (lines 43, 87) in the non-MKL LAPACK path. In `lapack_wrapper.hpp` the corresponding lines are commented out.

yingjerkao:
## Problem

Fixes #951. All 14 pybind TUs carried a vestigial `#include "complex.h"` — the C header that defines the macro `I` (`#define I _Complex_I`). It collides with `template <std::size_t I>` in `include/Type.hpp` and compiled only by include-order luck; any user TU including `<complex.h>` before cytnx headers broke the same way. (macOS masks the bug — libc++ shims `complex.h` to `<complex>` in C++ mode — but Linux/GCC reproduces it exactly.)

## Fix

- Removed the include from all 14 pybind TUs (none used `_Complex_I`/`creal`/`cimag`).
- Renamed the sole colliding template parameter in `include/Type.hpp` to `Idx`, hardening the public header against macro-`I` from any include order.
- `Det_internal.cpp`'s TU-local `<complex.h>` use is deliberate and untouched.

## Testing

Compile-time fix: the C++ library, full test suite (baseline pass), and the full python module (`BUILD_PYTHON=ON`) all build cleanly. The hazard was reproduced pre-fix with the real glibc macro expansion (hard parse error at the template line) and compiles clean post-fix.

Follow-ups noted during review (not in this PR): `index_in_tuple_helper` appears to be dead code, and `Det_internal.cpp:3`'s `<complex.h>` include also looks vestigial.

### DIFF
--- a/tests/include_order_canary_test.cpp
+++ b/tests/include_order_canary_test.cpp
@@ -1,0 +1,16 @@
+// Regression guard for #951: cytnx public headers must stay usable in a TU
+// that sees the C complex.h macro `I` before including any cytnx header.
+#include <complex.h>
+#ifndef I
+  // libc++'s complex.h does not leak the C macro `I` in C++ mode; define it
+  // explicitly so this guard bites on every platform, not just glibc.
+  #define I (18973)
+#endif
+
+#include "cytnx.hpp"
+
+#undef I
+
+#include "gtest/gtest.h"
+
+TEST(IncludeOrderCanary, HeadersSurviveComplexHMacroI) { SUCCEED(); }
--- a/include/Type.hpp
+++ b/include/Type.hpp
@@ -83,13 +83,13 @@ namespace cytnx {
     struct is_complex_floating_point_impl<cuda::std::complex<T>> : std::is_floating_point<T> {};
 #endif
 
-    template <std::size_t I, typename T, typename Tuple>
+    template <std::size_t Idx, typename T, typename Tuple>
     constexpr std::size_t index_in_tuple_helper() {
-      static_assert(I < std::tuple_size_v<Tuple>, "Type not found!");
-      if constexpr (std::is_same_v<T, std::tuple_element_t<I, Tuple>>) {
-        return I;
+      static_assert(Idx < std::tuple_size_v<Tuple>, "Type not found!");
+      if constexpr (std::is_same_v<T, std::tuple_element_t<Idx, Tuple>>) {
+        return Idx;
       } else {
-        return index_in_tuple_helper<I + 1, T, Tuple>();
+        return index_in_tuple_helper<Idx + 1, T, Tuple>();
       }
     }
 
--- a/pybind/algo_py.cpp
+++ b/pybind/algo_py.cpp
@@ -12,7 +12,6 @@
 
 #include "cytnx.hpp"
 // #include "../include/cytnx_error.hpp"
-#include "complex.h"
 
 namespace py = pybind11;
 using namespace pybind11::literals;
--- a/pybind/bond_py.cpp
+++ b/pybind/bond_py.cpp
@@ -12,7 +12,6 @@
 
 #include "cytnx.hpp"
 // #include "../include/cytnx_error.hpp"
-#include "complex.h"
 
 namespace py = pybind11;
 using namespace pybind11::literals;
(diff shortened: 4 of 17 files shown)


tests/include_order_canary_test.cpp
```diff
@@ -0,0 +1,16 @@
...
+  #define I (18973)
+#endif
+
+#include "cytnx.hpp"
+
+#undef I
+
+#include "gtest/gtest.h"
+
+TEST(IncludeOrderCanary, HeadersSurviveComplexHMacroI) { SUCCEED(); }
```
IvanaGyro, line 16: This test is not really useful. If this test is to avoid including `<complex.h>`, we should use other approaches.
And the comment of the test is not really correct. We only run tests on Ubuntu which uses glibc by default. The test doesn't really guard for all platforms.

IvanaGyro: Except the test, LGTM

pcchen:
## Code Review

Fixes #951 (the C `complex.h` macro-`I` collision). The fix is correct and well-designed — the only blocker is a merge conflict (which is what's making CI look red on macOS).

### The fix is correct
- **`Type.hpp` `I`→`Idx` rename** — all 5 uses updated consistently. This is the real robustness fix: it protects the public header against a macro-`I` from *any* include order or consumer, not just cytnx's own TUs. ✅
- **14 pybind `#include "complex.h"` removals** — vestigial (none use `_Complex_I`/`creal`/`cimag`); removes the internal macro source. ✅
- **Canary test** is a clever compile-time regression guard: includes `<complex.h>`, force-defines `I (18973)` even where libc++ doesn't leak it, includes `cytnx.hpp` under the active macro, then `#undef I` before gtest. The file compiling *is* the test, and it covers the whole public header tree. Confirmed no other bare-`I` identifier remains in public headers. ✅
- `DownstreamFindPackage` passes (no C++20 surprises here). ✅

### 🔴 Blocker — merge conflict (the actual cause of the red macOS CI)
The 4 failing macOS/wheel jobs fail at the **"Merge with latest target branch"** step (~30s), not at compile — the branch conflicts with master in `include/Type.hpp` and `tests/CMakeLists.txt`, so the CI's pre-build merge fails. It's misleading; it is not a code error. **Rebasing onto master** clears it.

### ⚠️ Residual to verify after rebase
The canary's **macOS** compilation is currently unverified — macOS CI never got past the merge step, so it never actually compiled the canary. The canary deliberately force-`#define I` before `#include "cytnx.hpp"`, which recreates the exact condition libc++ *avoids* (libc++ shims `complex.h` to not leak `I` in C++ mode precisely so its own headers can use `I`). On Linux/libstdc++ it compiled fine; on macOS/libc++, force-defining `I` before pulling libc++ headers via `cytnx.hpp` *could* break a libc++-internal use of `I`. Please confirm macOS `BuildAndTest` actually compiles the canary after the rebase — if it breaks, the canary may need to include cytnx's system deps before force-defining `I`, or scope the macro more narrowly.

Follow-ups noted (dead `index_in_tuple_helper`, `Det_internal.cpp`'s `complex.h`) are honestly out of scope. Fine.

**Recommendation:** good fix — needs a rebase onto master (which also un-hides whether the canary compiles on macOS).

pcchen (approved):
Approving. Correct fix for the `complex.h` macro-`I` collision (#951): the `I`→`Idx` rename hardens the public header against a macro-`I` from any include order, the 14 vestigial pybind `complex.h` includes are removed, and the include-order canary is a solid compile-time regression guard. Post-rebase CI is fully green on all platforms — importantly the macOS runners now pass, which resolves my earlier residual concern (the canary's force-`#define I` before `#include "cytnx.hpp"` compiles cleanly on macOS/libc++, not just Linux). No findings.

PR #982, opened 2026-07-05
fix(Type): promote mixed complex/real dtypes by precision; add to_real/to_complex

### LINKED ISSUE #858: dtype promotion to complex uses type ordering
Magic tricks such as ` in.dtype() <= 2 ? in.dtype() + 2 : in.dtype()` promote `Double` and `Float` to their corresponding complex data types. This should better be handled with a helper in `Type.hpp` that does not rely on the ordering of dtypes and works for other types as well.

These files are possibly affected (maybe others depend on the ordering as well in different ways):
```
src/linalg/Lstsq.cpp:53
src/linalg/Eigh.cpp:25
src/linalg/Tridiag.cpp:26       (used in error check, slightly different)
src/linalg/Gesvd.cpp:30
src/linalg/Rsvd.cpp:41
src/linalg/Gesvd_truncate.cpp:52,67
src/linalg/Svd.cpp:30
src/linalg/Rsvd_truncate.cpp:65,83
src/linalg/Lanczos_ER.cpp:558   (uses Type.is_complex() form — same intent)
```

yingjerkao:
## Problem

Fixes #858; addresses the core promotion rule behind #692. `Type_class::type_promote` picked the lower enum index, and the dtype enum interleaves complexness and precision — so `ComplexFloat + Double → ComplexFloat`, silently discarding double precision (numpy/torch: `ComplexDouble`). Separately, nine-plus linalg call sites computed "the real counterpart dtype" with the enum-layout hack `dtype() <= 2 ? dtype() + 2 : dtype()`.

## Fix

- `type_promote`: mixed complex/real pairs now promote their real counterparts and re-complexify (`ComplexFloat + Double → ComplexDouble`). Same-kind promotion is unchanged. Still `constexpr`, so the compile-time kernel instantiation (`type_promote_t`) and runtime output-dtype selection stay coherent automatically.
- New `Type.to_real` / `Type.to_complex` helpers (layout-free, `constexpr`); all enum-ordering hacks across 11 linalg files replaced (`Svd`, `Gesvd`, `Gesvd_truncate`, `Rsvd`, `Lstsq`, `Eigh`, `Tridiag`, `Lanczos_Gnd_Ut`, `Lanczos_ER`, `Conj`, `Conj_`).
- `Vectordot` had its own hand-rolled lower-index dtype selection that never used `type_promote`; it now routes through `type_promote` so its output dtype agrees with the elementwise ops. Sibling functions (`Dot`, `Matmul`, `Gemm`, `Tensordot`, …) carry the same hand-rolled pattern without test coverage and are tracked separately.

## User-visible changes

- Mixed `ComplexFloat ⊙ Double` results are now `ComplexDouble` everywhere (Tensor ops, UniTensor contractions).
- `Vectordot` integer-pair outputs follow `type_promote`'s signed/unsigned adjustment (e.g. `Vectordot(Uint64, Int32)` now yields `Int64`).

## Testing

New gtests pin both directions of mixed promotion at the `Type` and `Tensor` levels, the unchanged same-kind rules including the signed/unsigned adjustment path, `to_real`/`to_complex` (plus compile-time `static_assert` pins), and the dtype adjacency the adjustment relies on. `Vectordot.alltype` — which caught the Vectordot inconsistency — passes. Full suite passes at baseline.

### DIFF
--- a/include/Type.hpp
+++ b/include/Type.hpp
@@ -338,18 +338,44 @@ namespace cytnx {
       return cy_typeid_v<T>;
     }
 
+    // Real counterpart of a dtype: ComplexDouble -> Double, ComplexFloat -> Float,
+    // anything else unchanged. Replaces the "dtype <= 2 ? dtype + 2 : dtype" idiom
+    // (without depending on the enum layout).
+    static constexpr unsigned int to_real(unsigned int type_id) {
+      check_type(type_id);
+      if (type_id == ComplexDouble) return Double;
+      if (type_id == ComplexFloat) return Float;
+      return type_id;
+    }
+
+    // Complex counterpart of a dtype: Double -> ComplexDouble, Float -> ComplexFloat,
+    // complex types unchanged, Void unchanged, integral/bool -> ComplexDouble.
+    static constexpr unsigned int to_complex(unsigned int type_id) {
+      check_type(type_id);
+      if (is_complex(type_id)) return type_id;
+      if (type_id == Double) return ComplexDouble;
+      if (type_id == Float) return ComplexFloat;
+      if (type_id == Void) return Void;
+      return ComplexDouble;
+    }
+
     // Find a common type for typeL and typeR
     static constexpr unsigned int type_promote(unsigned int typeL, unsigned int typeR) {
+      if (typeL == Void || typeR == Void) return Void;
+      // Mixed complex/real: promote the real counterparts, then re-complexify.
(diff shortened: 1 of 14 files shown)


ianmccul:
This is very ambitious, I was thinking this was basically unfixable since there are so many places that assume the output type and cast from `void*`.  This is very dangerous, until there is a test for every possible dtype-promoting function...

PR #985, opened 2026-07-06
fix(tests): match blocks by qindices in apply() pybind test helper

yingjerkao:
## Summary

Fixes the two pre-existing failures in `tests/test_apply_pybind.py`:

- `TestApplyBlockFermionicUniTensor::test_apply_applies_signflips`
- `TestApplyBlockFermionicUniTensor::test_apply_inplace_applies_signflips`

**The backend is correct — the test helper was the bug.** `T.apply()` output was verified element-wise against the hand-built `BFUT3PERM` reference over the full tensor shape: 0 mismatches, including the expected sign-flipped entries (−5, −6, −7, −8), matching the C++ `LinAlgElementwise` expectation in `tests/BlockFermionicUniTensor_test.cpp`.

## Root cause

`_are_nearly_eq` zipped `get_blocks_()` by list position, but block order is not canonical: `permute().contiguous()` keeps the source tensor's block enumeration order, while a freshly constructed reference tensor enumerates blocks in canonical qnum order. Positional pairing therefore compared unrelated blocks. The C++ helper `AreNearlyEqUniTensor` (`tests/test_tools.cpp`) explicitly matches blocks by quantum numbers ("they might be not in the same order!"), which is why the C++ test passes.

## Changes

- Rewrite `_are_nearly_eq` to match blocks by `get_qindices()` and require `signflip()` flags to agree before comparing raw block data. Deliberately no sign-folding, so the tests stay sensitive to whether `apply()` actually baked the pending signs in (verified: unapplied `T` vs reference → `False`, applied → `True`).
- Use the public `Bond.redirect_()` instead of the `c_redirect_` shadow binding (equivalent on master via the `Bond_conti.py` glue wrapper; avoids a conflict with the in-progress shadow-API removal branch).

## Test plan

- `tests/test_apply_pybind.py`: 9 passed (was 7 passed / 2 failed)
- Full `pytests/` suite: 28 passed (matches baseline)
- Test-only change; no C++ touched


pcchen:
## Code Review

Rewrites `_are_nearly_eq` in `tests/test_apply_pybind.py` to match blocks by quantum-number sector (`get_qindices`) instead of list position, requiring `signflip()` agreement before comparing block data; plus `c_redirect_` → `redirect_`.

### Correct as far as it goes
- The `_are_nearly_eq` sector-matching + signflip-agreement logic is right — it matches the C++ `AreNearlyEqUniTensor` (blocks are not in canonical order between `permute().contiguous()` and a freshly-built reference), and deliberately doesn't sign-fold, so it stays sensitive to whether `apply()` actually baked the pending signs in. ✅
- `redirect_()` is correct on current master (post-#986, which binds it directly). ✅

### 🟠 Finding 1 — the fix isn't actually enforced by CI
CI runs `pytest pytests/` (`ci-cmake_tests.yml:140`), and this file lives under `tests/`, which pytest does **not** discover there. So `tests/test_apply_pybind.py` has never been run by CI — that's why the "2 failures" only appear in manual runs, and why #985's green checks don't include them. This PR fixes the helper but leaves the file where CI won't run it, so the now-passing tests **remain silently skipped**. To make it a real fix, also `git mv tests/test_apply_pybind.py pytests/apply_test.py` so CI enforces it.

### 🟠 Finding 2 — duplicates #1001
#1001 (still open, `CONFLICTING`) does the **identical** `_are_nearly_eq` rewrite **and** the `pytests/` file move **and** the `redirect_` change, as part of its larger Bond-immutable work. #985 and #1001 modify the same lines, so they'll conflict — whichever merges second must drop this. Given #1001 is currently stuck (changes-requested + conflicts + gateway-narrowing follow-up), landing #985 — **made complete by adding the file move** — is a pragmatic way to fix these tests now, after which #1001 drops its copy.

### Recommendation
Good, correct fix — but please add the **`pytests/` relocation** (otherwise it's a no-op in CI), and coordinate with #1001 so the duplicate `_are_nearly_eq` change doesn't collide.

yingjerkao:
Addressed the review findings:

**Finding 1 (fix not enforced by CI) — resolved.** Moved `tests/test_apply_pybind.py` → `pytests/apply_test.py` (commit 52091dcb, a 100% rename) so CI's `pytest pytests/` discovers and runs it. The branch was also brought up to date with `master` (168 commits behind → current) so the relocation is verified against today's code, not a stale tree:
- `pytests/apply_test.py` → **9 passed** (both previously-failing signflip tests included)
- full `pytests/` → **135 passed, 1 skipped** — no collection breakage

**Finding 2 (duplicates #1001) — acknowledged.** #1001 is still `OPEN` / `CONFLICTING` / `DIRTY`, so landing #985 (now made complete by the file move) remains the pragmatic way to fix these tests. Once #985 merges, #1001 should drop its duplicate `_are_nearly_eq` rewrite + `pytests/` move to avoid the collision.

PR #986, opened 2026-07-06
refactor(pybind): bind in-place methods directly, drop the c__* shadow API

### LINKED ISSUE #336: Always return reference for inplace methods
Some inplace methods of UniTensor do not return a reference to the UniTensor. It would be good to have this consistent, and is especially important if we want to suggest users to use train-like initialization of tensors. In the paper we will use the non-underscore versions, so this is fine. But users might use the underscore version as well.
These methods should return a reference:
```
permute_
reshape_
normal_
uniform_
to_dense_
```

For others I think this is probably not so important. The methods below return `void`. I am not sure if it should also better be a reference to the UniTensor.
```
contiguous_
group_basis_
put_block(_)
set
combineBonds
```

### LINKED ISSUE #779: Python bindings: chaining schemantics use inconsistent return self.method()
Chaining requires all methods to return a pointer to the object itself. Currently, the pybindings (for example in `unitensor_py.cpp`) use different constructions for this:
1) Returning a new object
```
return self.permute(mapper,rowrank);
```
2) Changing the object and returning with `&`
```
  return &self.permute_(mapper,rowrank);
```
3) Changing the object and not returning with `&`
```
return self.get_block_(labels,qnum, force);
```
This is inconsistent. It should be changed to a common way, and it is best to avoid copying the object.

yingjerkao:
Fixes #779, fixes #336.

## What this does

In-place dunders and `_`-suffixed methods were bound under shadow names (`c__iadd__`, `cConj_`, `c_relabel_`, …) and re-exported by monkey-patching wrappers in `cytnx/*_conti.py` that called the shadow binding and returned `self`. That left three inconsistent return conventions, a wasted Python-object copy on every in-place op, and the wrong API surface in `help()`/stubs.

**Commit 1 (`efa8ea43`)** binds the real method names directly with lambdas that take and return the *same* `py::object`, so Python object identity is preserved and chaining (`t.Conj_().Abs_()`, `ut.set_name("x").relabel_(["a","b"])`) works without the delegation layer:

```cpp
.def("Conj_", [](py::object self) {
  self.cast<Tensor &>().Conj_();
  return self;
})
```

- `tensor_py.cpp`: `__iadd__/__isub__/__imul__/__itruediv__/__ifloordiv__/__ipow__/__imatmul__`, `Conj_/Exp_/InvM_/Inv_/Abs_/Pow_`
- `unitensor_py.cpp`: `Conj_/Trace_/Transpose_/normalize_/Dagger_/tag/__ipow__/Pow_/truncate_/set_name/set_label(s)/relabel_/relabels_/set_rowrank_/convert_from`
- `bond_py.cpp`: `redirect_`; `getDegeneracy(qnum, return_indices=)` and `group_duplicates()` now return their aux indices directly from C++
- `storage_py.cpp`: the 11 `c_pylist_<T>` bindings collapse into one `pylist()` that switches on `self.dtype()`
- The `c__*`/`c*` shadow bindings and their `*_conti.py` wrappers are removed; genuinely Python-side features (iterators, `Network.Diagram`, `Qs`, the `UniTensor.at` proxy) remain.

**Commit 2 (`a3367f99`)** fixes a bug this conversion surfaced: the string-label overload of `UniTensor.truncate_` called the non-mutating `truncate()` and discarded the result, so `ut.truncate_("label", dim)` was a silent no-op. This is pre-existing on master (the old `ctruncate_` shadow binding had the same call), not introduced by the refactor. It now calls the in-place C++ `truncate_(label, dim)`, matching the `bond_idx` overload.

## Testing

- New `pytests/binding_inplace_test.py`: object-identity (`r is t`) and chaining guarantees, `Storage.pylist` round-trip, `Bond.group_duplicates` tuple return, and a grep-style guard that no shadow names survive on the Python classes.
- New regression tests in `pytests/unitensor_test.py` for the `truncate_` fix: string-label truncation actually shrinks the bond in place (was red before commit 2 — shape stayed `[3, 4]`), and the string-label/bond-index overloads agree.
- Full pytest suite: **37 passed, 0 failed** (master baseline 28 + this branch's 9 new tests).

## Notes for reviewers

- Behavior change: in-place methods now return the *same* Python object (`r is t` is `True`); previously some paths returned a fresh wrapper around the same C++ impl.
- `ut.truncate_("label", dim)` now actually truncates — code that accidentally relied on the no-op will see the bond shrink.
- Follow-up binding-hygiene work (collapsing per-dtype operator overloads, `__ne__`/`__bool__`) is planned as separate PRs stacking on this one.


IvanaGyro:
How will this interact with #912? It seems `getDegeneracy` may conflict against each other.

yingjerkao:
@IvanaGyro good catch — they do touch the same surface, but it's a stub-vs-binding staleness rather than a hard code conflict. Both branches edit `unitensor_py.cpp`, but in disjoint regions (#912: the `put_block` `in`→`Tin` rename at ~L664–789; this PR: the in-place bindings elsewhere), so `git merge-tree` reports a clean merge with no conflicts — the only real interaction is the stub regeneration below.

Concretely, what this PR does to that surface in `bond_py.cpp`:
- `getDegeneracy(qnum)` + `c_getDegeneracy_refarg(qnum, indices)` → a single `getDegeneracy(qnum, return_indices=False)` that returns the degeneracy, or `(degeneracy, indices)` when `return_indices=True`.
- `group_duplicates()` + `c_group_duplicates_refarg(mapper)` → `group_duplicates()` returning `(Bond, mapper)` directly.
- `c_redirect_` → folded into `redirect_`.

The interaction is that #912 ships **generated, committed** stubs, and its `cytnx/cytnx/__init__.pyi` currently contains the shadow methods this PR removes:

```
def c_getDegeneracy_refarg(self, qnum, indices: list) -> int: ...
def c_group_duplicates_refarg(self, mapper: list) -> Bond: ...
def c_redirect_(self) -> Bond: ...
def getDegeneracy(self, *args, **kwargs): ...
def group_duplicates(self, *args, **kwargs): ...
```

So once this PR lands, those `c_*` stub entries describe methods that no longer exist on the extension, and the real `getDegeneracy(qnum, return_indices=False)` / `group_duplicates()` signatures aren't captured. Since #912's stubs are `pybind11-stubgen` output (not hand-written), that's a mechanical regeneration, not a manual reconciliation.

Suggested ordering: land this PR first — it deletes the entire `c__*`/`c*` shadow surface, so regenerating #912's committed stubs afterward is strictly better (they drop the dead `c_*` entries and pick up the cleaned-up signatures). If #912 lands first, we should regenerate the committed stubs as part of rebasing this PR onto it. Either way the fix is a stub regen against the post-refactor extension; happy to sequence them whichever way you prefer.

pcchen:
## Code Review

Removes the `c__*`/`c*` shadow bindings and their `cytnx/*_conti.py` monkey-patch wrappers, binding the real method/dunder names directly with the identity-preserving idiom (`py::object self` -> `self.cast<T&>().mutate(); return self`). Commit 2 fixes a real pre-existing bug where `UniTensor.truncate_("label", dim)` called the non-mutating `truncate()` and was a silent no-op.

### Correctness — verified
- **The idiom is right** — canonical pybind11 pattern for in-place ops with exact object identity; overload dispatch still keys off the second arg. ✅
- **Refactor is complete — no dangling references.** Grepped the PR head tree: every `c__*`/`c*` name is gone from the Python package, C++ bindings, examples, and docs; the only survivors are inside the test that asserts they're gone. ✅
- **`truncate_(string, dim)` C++ overload exists** (`UniTensor.hpp:385/1071/5571`), so commit 2's fix compiles and is real, and it's covered by a good regression test. ✅
- **No overloads dropped** in the mechanical conversion (each operator keeps Tensor + 11 scalars + Scalar + 11 numpy_scalars), and **all build/test/wheel CI is green on every platform**. ✅
- **`__imatmul__` assignment is sound**: `self.cast<Tensor&>() = Dot(self.cast<Tensor&>(), rhs)` — Dot's result is computed before assignment, no aliasing hazard, identity preserved. ✅

### Findings

**1. `@=` semantics genuinely change — untested and not called out (low, but worth a note).**
On master the wrapper was misspelled `def __imatmul` (missing the trailing `__`), so `Tensor.__imatmul__` was never actually a special method — Python's `a @= b` fell back to `a = a @ b`, producing a **new** object. This PR binds `__imatmul__` correctly, so `@=` now mutates in place and preserves identity (`c = a; a @= b` now makes `c` observe the change). That's a correct fix, but unlike the other in-place ops (whose identity was already preserved by the old wrappers) it's a genuine behavioral change. It has **no test** and isn't mentioned in "Notes for reviewers." Suggest adding an `@=` identity test and a changelog line.

**2. `codecov/patch` fails — the new mechanical lines are largely uncovered.** The many `numpy_scalar` overloads, `Storage.pylist`'s `default` branch, and `__imatmul__` aren't exercised. Not a blocker for a mechanical rename (behavior unchanged on those paths), but a couple of representative tests (one numpy-scalar in-place op, plus `@=`) would clear the gate and cover the one path that actually changed.

**3. `Storage.pylist` invalid-dtype now raises differently (nit).** Old code raised Python `ValueError`; the new `default` calls `cytnx_error_msg` -> `std::logic_error` -> `RuntimeError` (or `CytnxError` once #989 lands). Essentially unreachable edge case, but a minor exception-type change if anything caught `ValueError`.

**4. `getDegeneracy` return type varies by flag (nit).** Scalar when `return_indices=False`, a `(deg, indices)` tuple otherwise — preserved from the old wrapper, but a dynamic return type is awkward for stubs/typing. Fine to keep for compatibility.

**5. Verbosity (nit, acknowledged).** The per-dtype overload duplication is large; already flagged as planned follow-up.

### Verdict
Strong, careful refactor: the mechanical conversion is correct and complete, it deletes a real maintenance wart, fixes a genuine `truncate_` no-op bug, and passes full CI on all platforms. The one thing I'd ask for before merge is a small test + release note for the **`@=` identity change** (which also clears `codecov/patch`); everything else is nits. **Approve-with-nits.**

pcchen (approved):
Approving. The refactor is correct and complete: the identity-preserving `py::object self` idiom is applied consistently, no shadow-name call sites dangle anywhere in the tree (verified against the head tree), no operator overloads were dropped, the `truncate_(string, dim)` C++ overload exists so commit 2's no-op fix is real, and full build/test/wheel CI is green on every platform. It also deletes a real maintenance wart and fixes a genuine silent no-op bug.

Non-blocking follow-ups (fine to do here or in a fast-follow):
- Add an `@=` identity test + a changelog line. `__imatmul__` was previously misspelled `def __imatmul` in the Python wrapper, so `a @= b` fell back to `a = a @ b` (new object); it now truly mutates in place and preserves identity. This is the one genuine behavioral change and is currently untested.
- That same test (plus one numpy-scalar in-place op) would also clear the `codecov/patch` gate.

pytests/binding_inplace_test.py
```diff
@@ -0,0 +1,98 @@
...
+def test_bond_redirect__chains():
+    b = cytnx.Bond(2, cytnx.BD_KET)
+    bid = id(b)
+    r = b.redirect_()
+    assert r is b
+    assert id(r) == bid
+    assert b.type() == cytnx.BD_BRA
+
+
+def test_shadow_bindings_are_gone():
```
IvanaGyro, line 70: This test is useless for the future code. This should be clean up.

yingjerkao: Agreed — it's a one-time migration guard with no value for future code (and the last reference to those dead `c_*` names). Since #986 is already merged, I've removed it in a follow-up: #1009. That PR also adds a real `@=` identity test in its place (the one genuine behavioral change from this refactor, which was untested), so the file keeps only the meaningful contract tests.

Issue #974, opened 2026-07-05
macOS: ARPACK-backed linalg::Lanczos / Arnoldi tests segfault in dsaupd_ (null dereference)


yingjerkao:
## Summary

Now that the C++ test suite compiles on macOS (#973), every test that reaches the ARPACK-backed Krylov path crashes with SIGSEGV (`EXC_BAD_ACCESS`, `KERN_INVALID_ADDRESS at 0x0`) inside ARPACK's `dsaupd_`. These tests had no macOS baseline before #973 because the suite never built there, so this is newly *visible*, not newly *introduced* — the crash also hits test files #973 did not touch (e.g. `tests/linalg_test/Lanczos_test.cpp`).

## Observed crashes

Directly observed (each run terminates the whole test binary):

- `Lanczos.which_LM_test`
- `Lanczos.fermionic_LanczosArpackFermionic`
- `Lanczos_Ut.which_LM_test`
- `Lanczos_Gnd.Bk_Lanczos_test`
- `Arnoldi.which_LM_test`

The sibling `which_*` / `ExcuteTest`-based tests in the same suites funnel through the same `linalg::Lanczos(..., which, maxiter, cvg_crit, k, ...)` / `linalg::Arnoldi` entry points and are presumed affected. Non-ARPACK tests are fine: excluding `*Lanczos*` and `Arnoldi*`, the suite runs 997 tests with 982 passing.

Notably `Lanczos_Gnd.Lanczos_Gnd_test` and `Lanczos_Gnd.Bk_Lanczos_Gnd_test` (the non-ARPACK Gnd path) pass.

## Crash stack

From the macOS crash report (`~/Library/Logs/DiagnosticReports/test_main-*.ips`):

```
exception: EXC_BAD_ACCESS (SIGSEGV), KERN_INVALID_ADDRESS at 0x0000000000000000

libarpack.2.1.0.dylib   dsaupd_ + 1792
test_main               std::function<void (int*, char*, int*, char*, int*, double*, double*, int*, double*, int*, int*, int*, double*, double*, int*, int*)>::operator()(...) + 164
test_main               cytnx::linalg::_Lanczos_internal<double, cytnx::UniTensor>(...) + 1036
test_main               cytnx::linalg::_Lanczos(...) + 1656
[...]

PR #990, opened 2026-07-06
fix(arpack): pass Fortran hidden character-length arguments in ARPACK calls

yingjerkao:
## Summary

The `extern "C"` prototypes in `include/backend/arpack_wrapper.hpp` omitted the hidden length argument that Fortran compilers append for every `CHARACTER` dummy (`bmat`, `which`, `howmny`). Calling through such prototypes is undefined behavior. It happens to be benign for ARPACK's fixed-length dummies under today's gfortran ABI — and it was explicitly **not** the cause of the macOS segfaults investigated in #974 (those were environment-induced; see that issue) — but it is exactly the class of ABI mismatch that surfaces as platform-specific corruption, and was flagged as a cleanup follow-up when closing #974.

## Changes

- **include/backend/arpack_wrapper.hpp** — all twelve `x{s,n}aupd_` / `x{s,n}eupd_` prototypes gain trailing `std::size_t` length parameters, one per `CHARACTER` argument, with a comment documenting the convention. `std::size_t` matches gfortran ≥ 8 and LAPACK's modern `FORTRAN_STRLEN` convention; on LP64 little-endian targets it is also register/stack-slot compatible with the `int` lengths used by older compilers.
- **src/linalg/Lanczos.cpp, src/linalg/Arnoldi.cpp** — the `std::function` dispatch signatures extended to match, and every call site now passes the actual lengths (`bmat`: 1, `which`: 2, `howmny`: 1).

No functional change is expected on any platform where the code currently works: the callee already expected these arguments; we now actually pass them.

## Verification

On macOS arm64 (AppleClang, conda-forge arpack-ng 2.1.0 + OpenBLAS), with the #972 test compile fixes applied locally (not part of this branch):

- full `test_main` build succeeds;
- all **81 tests in `*Lanczos*` + `Arnoldi*` pass** (run without `DYLD_LIBRARY_PATH`, per the pitfall documented in #974/#972) — covering the d/s symmetric, d/s non-symmetric, and z/c complex driver pairs.

Linux/GCC CI exercises the same call paths and should show no change.

Related: #974 (root-cause investigation that flagged this), #626 (original ARPACK integration), #893 (previous wrapper cleanup, exonerated in #974).

### DIFF
--- a/include/backend/arpack_wrapper.hpp
+++ b/include/backend/arpack_wrapper.hpp
@@ -3,76 +3,89 @@
 
 #include "lapack_wrapper.hpp"
 #include <complex>
+#include <cstddef>
 
 namespace arpack {
+
+  // Fortran compilers pass a hidden length argument, appended after the declared
+  // arguments, for every CHARACTER dummy (gfortran >= 8 passes it as size_t; older
+  // gfortran and other Unix compilers as int -- passing size_t is compatible with
+  // both on LP64 little-endian targets, matching LAPACK's modern FORTRAN_STRLEN
+  // convention). Omitting them from the C prototypes is undefined behavior, so the
+  // declarations below carry one trailing length per CHARACTER argument and call
+  // sites must pass the actual lengths (bmat: 1, which: 2, howmny: 1).
 
   extern "C" {
 
   // === Real Symmetric (double precision) ===
   void dsaupd_(int *ido, char *bmat, int *n, char *which, int *nev, double *tol, double *resid,
                int *ncv, double *v, int *ldv, int *iparam, int *ipntr, double *workd, double *workl,
-               int *lworkl, int *info);
+               int *lworkl, int *info, std::size_t bmat_len, std::size_t which_len);
 
   void dseupd_(int *rvec, char *howmny, int *select, double *d, double *z, int *ldz, double *sigma,
                char *bmat, int *n, char *which, int *nev, double *tol, double *resid, int *ncv,
                double *v, int *ldv, int *iparam, int *ipntr, double *workd, double *workl,
-               int *lworkl, int *info);
+               int *lworkl, int *info, std::size_t howmny_len, std::size_t bmat_len,
+               std::size_t which_len);
 
   // === Real Symmetric (single precision) ===
   void ssaupd_(int *ido, char *bmat, int *n, char *which, int *nev, float *tol, float *resid,
                int *ncv, float *v, int *ldv, int *iparam, int *ipntr, float *workd, float *workl,
-               int *lworkl, int *info);
+               int *lworkl, int *info, std::size_t bmat_len, std::size_t which_len);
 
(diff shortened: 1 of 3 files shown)


pcchen:
## Code Review

Adds the Fortran hidden `CHARACTER`-length arguments (which every Fortran compiler appends for `CHARACTER` dummies) to the 12 ARPACK `extern "C"` prototypes and their call sites, fixing a genuine undefined-behavior class — calling through prototypes that omit the hidden lengths reads uninitialized register/stack slots. **Correct and clean.**

### Verified correct
- **Arg count & order per function** — `*aupd_` (chars `bmat`, `which`) gets 2 trailing lengths; `*eupd_` (chars `howmny`, `bmat`, `which`, in that source order) gets 3, ordered `howmny_len, bmat_len, which_len` to match the order the `CHARACTER` dummies appear. That's the correct gfortran convention. ✅
- **Values passed** — `bmat=1`, `which=2`, `howmny=1` at every call site (`(..., 1, 2)` for aupd, `(..., 1, 1, 2)` for eupd). ✅
- **`std::function` dispatch signatures** in `Lanczos.cpp`/`Arnoldi.cpp` extended consistently so `func_xsaupd = dsaupd_` etc. stay assignable — confirmed by the green compile. ✅
- **`std::size_t` is ABI-safe** — on LP64 little-endian it's compatible with both the modern (gfortran ≥ 8 `size_t`) and legacy (`int`) hidden-length conventions; the small values fit in the low bytes. Verified on macOS arm64 (81 Lanczos/Arnoldi tests) and now green on Linux + macOS CI. ✅
- **Strictly a safety improvement, no behavior change** — passing the args is defined whether or not the linked ARPACK reads them, whereas the old code was UB when it did. The PR is honest that this was *not* the #974 macOS-segfault cause (environment-induced), just the cleanup flagged when closing it.

### Notes (non-blocking)
- No new test — reasonable, since this is a benign-under-current-ABI UB cleanup with no observable behavior change; the existing Lanczos/Arnoldi suite (green in CI) covers the affected call paths.
- Behind master but conflict-free; a fresh CI run is optional (the current green is valid — not the stale-#1007 pattern).

No findings requiring changes — a well-reasoned, well-documented ABI-correctness fix. **LGTM.**

pcchen (approved):
Approving. Correct, clean Fortran/C ABI fix: the 12 ARPACK `extern "C"` prototypes and their call sites now carry the hidden `CHARACTER`-length arguments — right count and order per function (`aupd`: `bmat,which` → `1,2`; `eupd`: `howmny,bmat,which` → `1,1,2`), consistent `std::function` dispatch signatures, and `std::size_t` chosen for ABI-compatibility on LP64 little-endian with both the modern (gfortran ≥8) and legacy (`int`) hidden-length conventions. Strictly a UB-safety improvement with no behavior change, and CI is green on all platforms including the Arnoldi/Lanczos suites that exercise these calls.

Issue #928, opened 2026-06-22
Fix stub-quality issues in the pybind layer instead of post-processing in generate_stubs.py


IvanaGyro:
Sub-issue of #585.

## Background

`tools/generate_stubs.py` (added in #912) runs `pybind11-stubgen` and then **post-processes** the output in `sanitize()` to paper over signatures that pybind11 emits in a form static type checkers reject. Post-processing is lossy and fragile: it rewrites information that the binding *could* have produced correctly in the first place. The durable approach — demonstrated by #912 (binding order) and #915 (per-Python-type overloads) — is to fix the **binding** so `pybind11-stubgen` emits valid, precise types directly, shrinking or eliminating the sanitizer.

This issue tracks moving the remaining workarounds from the generator into the pybind layer.

## Currently masked by `generate_stubs.py` post-processing

`sanitize()` rewrites stubgen output with these rules:

- `-> ...:`  →  `-> typing.Any:`
- `: ...` (parameter annotation)  →  `: typing.Any`
- `[...]` (subscript element)  →  `[typing.Any]`

Each `...` is a type pybind11-stubgen could not parse, almost always a **raw C++ type leaking into a binding's docstring/signature** (e.g. `cytnx::Tensor`, `boost::intrusive_ptr<...>`). The fix is at the binding: stop emitting raw C++ types, so stubgen produces the real Python type and the rewrite becomes a no-op.

(The trailing-whitespace / end-of-file normalization in `sanitize()` is just pre-commit alignment and can stay.)

### Already fixed at the binding layer (templates to follow)

- **#912** — `generator_binding` was registered before `tensor_binding`, so the `Tensor` class was not yet registered when `zeros`/`ones`/`arange`/`eye`/`identity`/`linspace`/`_from_numpy` were bound; pybind11 fell back to a raw C++ return type that sanitized to `typing.Any`. Reordering the bindings makes them type as `-> Tensor` directly.
- **#915** — `linalg.ExpH`/`ExpM` were bound once per C++ scalar dtype, collapsing to duplicate `@typing.overload` signatures (`overload-cannot-match`). Binding one overload per Python-distinguishable dtype removes the duplicates at the source.

## Remaining work (move to the pybind layer)

[...]

PR #991, opened 2026-07-06
fix(pybind): collapse operator overloads per Python type; add __ne__/__bool__; fix numpy dtype loss

yingjerkao:
## Problem

Addresses #928/#916 and finishes the core of #692. Every Tensor/UniTensor arithmetic operator was bound ~24× (one per C++ scalar type + numpy scalars). Beyond the stub blowup this caused **live dtype bugs**: pybind11's arithmetic caster accepts any `__index__`-bearing object even in the no-convert pass, so plain fixed-width int overloads registered ahead of the numpy-scalar ones silently ate `np.int32/uint64/…` and produced Int64 results; `t[0] = np.float32(x)` raised a hard RuntimeError (Tensor `__setitem__` had no numpy coverage and `complex128` registered first). Separately, `t1 != t2` returned a single always-`False` bool for non-empty tensors (Python's default `__ne__` negated the elementwise `__eq__` Tensor via `__len__`), and `bool(t)` fell through to `__len__`.

## Fix

- Every binary/reverse/in-place operator on Tensor and UniTensor collapsed to a per-Python-distinguishable-type keep-set (24→14 / 23→13): Tensor operand, numpy scalars (32-bit + integer dtypes preserved), arbitrary-precision `py::int_` with int64/uint64 runtime dispatch, `double`, `complex128`, `Scalar` — ordering is load-bearing and documented once in the new shared `pybind/pyint_dispatch.hpp` ("KEEP-SET ORDERING" note, cross-referencing PR #915 whose ExpH/ExpM work pioneered the pattern).
- `__setitem__`/`set_elem` numpy coverage + ordering fixed on both classes; `__pow__`/`__ipow__` accept ints and numpy scalars.
- New elementwise `__ne__` (composed from existing kernels, Bool result, mirrors `__eq__` including the `Scalar` operand) and numpy-semantics `__bool__` (ValueError for multi-element or Void; truthiness of the single element otherwise). **Behavior change:** `if tensor:` previously meant `len(tensor) > 0`.

## Testing

`pytests/binding_dtype_test.py` (20 tests) with per-test baseline red/green documented — 10 genuinely red pre-fix (dtype loss, setitem error, `__ne__`/`__bool__` traps). Full pytest 55 passed. Known pre-existing gap (numpy scalar on the LEFT of Tensor hits `__iter__`) documented, out of scope.

## Stacking

**Base branch is `fix/pybind-inplace-return-self` (PR #986).** Merge #986 first, then retarget to `master`.

### DIFF
--- a/cytnx/Bond_conti.py
+++ b/cytnx/Bond_conti.py
@@ -1,9 +1,8 @@
 from .utils import *
 from cytnx import *
-# from typing import List
-# Use beartype to check the type of arguments
-from beartype.typing import List
 
 # All of Bond's Python-side helpers (redirect_, getDegeneracy, group_duplicates)
 # have been folded directly into the C++ bindings in pybind/bond_py.cpp; this
-# module intentionally has no remaining delegation wrappers.
+# module intentionally has no remaining delegation wrappers. The `List` import
+# that used to support type-hinted wrapper signatures here is gone now that
+# there is nothing left to type-hint against.
--- a/pybind/pyint_dispatch.hpp
+++ b/pybind/pyint_dispatch.hpp
@@ -1,0 +1,87 @@
+#ifndef PYBIND_PYINT_DISPATCH_HPP_
+#define PYBIND_PYINT_DISPATCH_HPP_
+
+#include <pybind11/pybind11.h>
+
+#include "Type.hpp"
+#include "cytnx_error.hpp"
+
+// =========================================================================
+// KEEP-SET ORDERING (canonical rationale -- per-operator comment blocks in
+// tensor_py.cpp / unitensor_py.cpp point here)
+// =========================================================================
+//
+// Every scalar-accepting Tensor/UniTensor operator is bound once per
+// Python-DISTINGUISHABLE dtype (the "keep-set"), not once per C++ dtype,
+// and the overloads MUST be registered in this order:
+//
+//   1. Tensor / UniTensor (elementwise operand)
+//   2. py::numpy_scalar<float>, py::numpy_scalar<std::complex<float>>
+//   3. py::numpy_scalar<int64/uint64/int32/uint32/int16/uint16/bool>
+//   4. py::int_ (arbitrary precision; dispatch_pyint below picks the
+//      int64 or uint64 kernel by magnitude)
+//   5. cytnx_double  (absorbs Python float and np.float64, a float subclass)
+//   6. cytnx_complex128  (absorbs Python complex and np.complex128)
+//   7. cytnx::Scalar  (widest implicit-conversion net, always last)
+//
+// WHY THE ORDER IS LOAD-BEARING (not just stub cosmetics): pybind11
+// resolves an overloaded call in two passes -- every overload is tried
+// WITHOUT implicit conversion first, then every overload is retried WITH
+// conversion -- and within a pass the first-registered match wins. Two
+// traps follow:
+//
+//   * The __index__ no-convert trap: pybind11's plain arithmetic
+//     type_caster for integral C++ types accepts ANY object satisfying the
+//     __index__ protocol even in the no-convert pass (pybind11 3.x cast.h,
+//     arithmetic type_caster). Every numpy integer scalar implements
+//     __index__, so a cytnx_int64-style overload registered before
+//     numpy_scalar<int32_t> greedily consumes np.int32 in the FIRST pass
+//     and silently rewrites its dtype to Int64; the numpy_scalar overload
+//     becomes unreachable. numpy scalars must therefore precede any plain
+//     integral (or py::int_-absorbing) overload.
+//
+//   * The __float__-fallback trap: Python's complex() builtin falls back
+//     to __float__ when __complex__ is absent, so the cytnx_complex128
+//     caster's convert pass accepts any float-like object (np.float32
+//     included). A complex128 overload registered before the float/double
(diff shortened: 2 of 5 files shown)


pcchen:
## Code Review

Collapses every Tensor/UniTensor arithmetic operator from ~24 per-C++-dtype overloads to a ~14-entry per-Python-distinguishable-type "keep-set" with a load-bearing registration order (centralized in the new `pybind/pyint_dispatch.hpp`), fixes real numpy dtype-loss bugs, and adds elementwise `__ne__` + numpy-semantics `__bool__` on Tensor.

### Correctness — verified
- **`dispatch_pyint`** is sound: `PyLong_AsLongLongAndOverflow` -> int64 when it fits (covers all negatives), else uint64, else a clean out-of-range error. `Cpr(rhs)` is literally `*this == rhs`, so `__eq__` and `__ne__` share one kernel. ✅
- **The keep-set ordering rationale is accurate.** Both documented pybind11 traps are real: the `__index__` no-convert trap (numpy ints greedily eaten by a plain integral overload) and the `__float__`-fallback trap (`complex128` capturing `np.float32`). The `__add__`/`__eq__` orderings match the documented sequence, and dropping `numpy_scalar<double>`/`<complex128>` is correct — `np.float64`/`np.complex128` are float/complex subclasses absorbed by the `double`/`complex128` casters in the no-convert pass. ✅
- **`__ne__`** = `(1 - self.Cpr(rhs)).astype(Bool)`, composed from existing kernels, mirrors `__eq__` (incl. the Scalar operand); fixes the old always-`False` trap. ✅
- **`__bool__`** raises `ValueError` on Void/multi-element (numpy semantics), else the single element's truthiness (with a proper complex-zero check). ✅
- **UniTensor** gets the same collapse but **intentionally no** `__eq__`/`__ne__`/`__bool__` — verified it has zero `__eq__` bindings, so `!=`/`bool()` fall back to safe Python identity defaults. Sound scoping (unlike Tensor, where an elementwise `__eq__` made the default `!=` actively wrong). ✅
- **Tests (20)** are excellent: dtype preservation for every numpy scalar type, large-int -> uint64, `__ne__` (not-silently-wrong / elementwise / Scalar), `__bool__` (raises + single-element), UniTensor coverage — plus `test_imatmul_preserves_identity`, which also closes the `@=` identity gap noted on #986. ✅
- **Compilation is green on all wheel platforms** (macOS x2, Ubuntu x2), which also confirms the re-enabled `operator==(Tensor, Scalar)` path builds. ✅

### Findings (all low / nits — none blocking)

1. **Documented behavior change — `bool(t)`.** `if tensor:` no longer means `len(tensor) > 0`; multi-element/Void now raise `ValueError`. Well flagged in the PR/code/tests, but it's a real API break — worth a **changelog/release-note** entry since downstream `if tensor:` idioms will start raising.
2. **Undocumented minor change — Python `bool` operands.** Dropping the C++ `cytnx_bool` overload routes Python `True/False` through `py::int_` (int64 path) instead of Bool. numpy-consistent and harmless for float tensors, but for a small-int-dtype tensor it can shift the result dtype (`int16 + True` -> Int64 vs Int16). Untested edge; worth a one-line note.
3. **`dispatch_pyint` return-type coupling (latent).** Both branches must deduce the same type from the generic `fn` — true for all current call sites, but a footgun for future callers; a short comment or `static_assert` would harden it.
4. **Nit:** the `pyint_dispatch.hpp` error message still uses the pre-#983 `"...range.%s", "\n"` idiom (zero-vararg calls are legal now).
5. **Nit (already acknowledged in-code):** `dispatch_pyint` duplicates #915's two-arg helper; the comment already plans to fold them once #915 merges.
6. **CI:** wheels compile green, but confirm a fresh `BuildAndTest`/pytest run on the current master-merged state before merge (the visible test run predates the retarget).

The left-of-Tensor numpy-scalar gap (`np.float32(1.0) + t` hits `__iter__`) is pre-existing, explicitly documented, and out of scope — reasonable.

### Verdict
Strong, unusually well-documented PR that fixes several genuine dtype bugs and centralizes a subtle, easy-to-regress ordering invariant with excellent tests. Findings are all notes, not blockers. **Approve-with-nits**, pending a fresh full-test CI run.

pcchen (approved):
Approving. This fixes several genuine numpy dtype-loss bugs (silent Int64/complex coercion, the `t[0] = np.float32(x)` hard error), replaces the always-`False` `!=` with a correct elementwise `__ne__`, and adds numpy-semantics `__bool__` — while collapsing the ~24x per-dtype overloads behind a well-documented, centralized keep-set ordering. The ordering rationale (the `__index__` no-convert and `__float__`-fallback traps) is accurate, `__ne__`/`__eq__` share one kernel, UniTensor's omission of `__eq__`/`__ne__`/`__bool__` is a sound identity-fallback choice, and the 20-test suite is thorough (including the `@=` identity test that also covers the #986 gap). Compilation is green on all wheel platforms.

Non-blocking follow-ups (fine here or in a fast-follow):
- Add a changelog/release-note line for the `bool(t)` behavior change (`if tensor:` now raises ValueError for multi-element/Void instead of meaning `len > 0`).
- Optional: note the Python-`bool`-operand dtype shift (now routed through the int64 `py::int_` path), and add a short comment/`static_assert` on `dispatch_pyint`'s same-return-type requirement.

Please confirm a fresh full `BuildAndTest`/pytest run is green on the current master-merged state before merging (the visible test run predates the retarget).

Issue #583, opened 2025-03-18
Run GitHub action with all allowed configurations and cache the build result


IvanaGyro:
Some bugs failing build proccesses are detected lately, e.g., #581, #582. The reason is that we only build the project in CI with only one configuration. We should build all accepted configurations with GitHub flow to catch build errors as soon as possible. Below are the combinations I suggest.

CPU:

HPTT | BLAS
--- | ---
enabled | OpenBLAS
enabled | MKL
disabled | OpenBLAS
disabled | MKL

CUDA:

| CUTT       | cuTENSOR   | cuQuantum  |
|------------|------------|------------|
| **enabled** | **enabled** | **enabled** |
| **enabled** | **enabled** | disabled    |
| **enabled** | disabled    | **enabled** |
| **enabled** | disabled    | disabled    |
| disabled    | **enabled** | **enabled** |
| disabled    | **enabled** | disabled    |
| disabled    | disabled    | **enabled** |
| disabled    | disabled    | disabled    |

Some CUTT features are hidden behind the cuTENSOR's or cuQuantum's feature flag, so we can't separate CUTT from the combinations above. In addition, we have different feature flags for cuTENSOR and cuQuantum in the codebase, so we can't expect the feature flag of cuTENSOR is always on when enabling cuQuanum's feature flag, even cuTENSOR is a dependency of cuQuantum.

pcchen:
If cuTENSOR is a dependency of cuQuantum, they don't we enforce to enable cuTENSOR when we enable cuTENSOR? 

Or we only have the option of cuQuantum. Use cannot enable/disalbe cuTENSOR.

yingjerkao:
With CUTT removed, I think we are converging to have both cuTENSOR and cuQuantum as dependencies. I am trying to set up a self-hosted GPU runner to test out the GPU CI/CD.

pcchen:
[...]

Issue #538, opened 2024-12-14
Build CI CD for gpu.


hunghaoti:
Currently, we only test cpu version for every commit. We may need to add for gpu version (maybe running test on local machine.)

kaihsin:
There is a way to do GPU testing for CI but I don't think we have bandwidth to do that

yingjerkao:
We are bringing in some help.

jysh1214:
Hi, I would like to contribute this part.

The related PR will use `nvidia/cuda` docker image.
What's the CUDA version we intent to use in CI?

jysh1214:
I think we are using ubuntu 20 or 22, right?

kaihsin:
22 would be nice. Also some DevOps help on setting up GitHub CI actions of
that would be great.

jysh1214:
@kaihsin 
I try to build and test in the nvidia container (`nvidia/cuda:11.8.0-devel-ubuntu22.04`).

Here is my config:
```
cmake -B $CYTNX_BUILD -G Ninja \
  -D CMAKE_BUILD_TYPE=Debug \
  -D Python_EXECUTABLE="$(which python3.11)" \
  -D Python_INTERPRETER="$(which python3.11)" \
  -D Python3_EXECUTABLE="$(which python3.11)" \
  -D Python3_INTERPRETER="$(which python3.11)" \
  -D BLAS_LIBRARIES="$OPENBLAS_BLAS_LIB" \
  -D LAPACK_LIBRARIES="$OPENBLAS_LAPACKE_LIB" \
  -D USE_MKL=OFF \
  -D USE_CUDA=ON \
  -D USE_CUTENSOR=OFF \
  -D USE_CUQUANTUM=OFF \
  -D USE_CUTT=OFF \
  -D USE_MAGMA=OFF \
  -D RUN_TESTS=ON
```

The 43 tests failed. It's a expected?
Can the tests pass with only `USE_CUDA` enabled?
```
95% tests passed, 43 tests failed out of 891

Total Test time (real) = 413.69 sec

The following tests FAILED:
        640 - NetworkTest.gpu_Network_dense_no_order (Failed)
        641 - NetworkTest.gpu_Network_dense_find_optimal (Failed)
        642 - NetworkTest.gpu_Network_dense_order_line (Failed)
        643 - NetworkTest.gpu_Network_dense_specified_order (Failed)
        644 - NetworkTest.gpu_Network_dense_reuse (Failed)
[...]

PR #992, opened 2026-07-06
fix(linalg): make GPU scalar in-place ops correct on CUDA (#988)

### LINKED ISSUE #988: GPU: scalar in-place ops throw on non-contiguous tensors for *= and /=, and do a per-call H2D copy
Follow-up to #906 / PR #980, from the code review of `fix/scalar-inplace-ops-preserve-storage`. Two confirmed GPU-side gaps that could not be fixed on that branch because the review machine had no CUDA (its `build_t` is `USE_CUDA=OFF`). Needs a CUDA machine.

## 1. Regression: GPU non-contiguous LHS now throws for `*=` and `/=`

`src/linalg/iMul.cpp` and `src/linalg/iDiv.cpp` (~lines 64–67) throw

```
[iMul][on GPU/CUDA] error two tensors must be contiguous.
```

for a non-contiguous LHS, while `src/linalg/iAdd.cpp` and `src/linalg/iSub.cpp` (~lines 64–74) handle non-contiguous LHS by passing the mappers into the GPU kernel. Since PR #980 routes scalar in-place ops through these kernels, `gpu_tensor.permute(...); gpu_tensor *= 2.0;` now throws where the old scalar path (out-of-place `Mul`/`Div` + storage swap) succeeded.

**Fix:** make the iMul/iDiv GPU branches handle non-contiguous LHS the same way iAdd/iSub already do — the mapper-passing pattern exists in those two files as a template.

## 2. Inefficiency: per-call CPU→GPU copy for the scalar wrapper

`_scalar_as_rank1_tensor` in `src/Tensor.cpp` builds the shape-`{1}` scalar tensor on CPU and then calls `s.to(device)`, so every scalar in-place op on a GPU tensor does a small H2D transfer. The old path passed a CPU-resident 1-element `Storage` directly into the cuda dispatch (see `src/linalg/Add.cpp` ~line 21, `Storage Cnst(1, lhs_dtype)`). Consider a kernel-level scalar path, or letting the in-place kernels accept a CPU scalar storage for a GPU LHS, to avoid the per-call copy.

## Setup

- Base on `fix/scalar-inplace-ops-preserve-storage` (PR #980) until it merges, then master.
- Build: `cmake -S . -B build_gpu -DUSE_CUDA=ON -DRUN_TESTS=ON && cmake --build build_gpu -j --target test_main`
- Add GPU gtests under `tests/gpu/` covering: non-contiguous GPU tensor `*=`/`/=` scalar (the regression), and the existing scalar in-place semantics (storage sharing, dtype preservation, real⊙=complex throws) on CUDA tensors.
- gtest naming convention: no underscores in test names (#857).

yingjerkao:
Fixes #988. Follow-up to #906 / #980 (based on `fix/scalar-inplace-ops-preserve-storage` until it merges, then this should retarget `master`).

Validated on a CUDA device (RTX 4070 Ti, CUDA 13) which the #980 review lacked.

## What #988 asked for
1. **Regression:** non-contiguous GPU LHS threw for `*=` / `/=` (scalar), while `iAdd`/`iSub` handled it.
2. **Efficiency:** the shape-`{1}` scalar wrapper did a per-call H2D copy.

## Changes (`src/linalg/i{Add,Sub,Mul,Div}.cpp`, `src/Tensor.cpp`)

1. **Non-contiguous scalar regression.** `iMul`/`iDiv` now route a broadcast scalar on a non-contiguous GPU LHS through the kernel — the `rconst` path scales every physical element in place, so the layout mappers are irrelevant and the result is correct. Because `cuMul`/`cuDiv` (unlike `cuAdd`/`cuSub`) do **not** consume the layout mappers, a genuine non-contiguous `tensor*=tensor` / `tensor/=tensor` now **throws loudly** instead of silently pairing mismatched elements.

2. **No per-call H2D copy.** `_scalar_as_rank1_tensor` keeps the wrapper **CPU-resident**. For a length-1 RHS the GPU kernels read the scalar with a host-side dereference and pass it in by value, so no copy is needed; the device check permits a host-resident length-1 scalar against a GPU LHS, and `cudaSetDevice` targets `Lt.device()`. (This also removes a latent UB: #980 moved the scalar *to* the GPU, making that `_Rin[0]` host dereference read a device pointer.)

3. **Safety — `real op= complex` corrupted memory.** The GPU path reinterpreted the real output buffer as complex (out-of-bounds write). Added a **device-independent** guard so it throws like the CPU path.

4. **Safety — narrow LHS corrupted memory.** A LHS narrower than the promoted dtype (e.g. `float_gpu *= 2.0`) had the promoted-width result written into the narrower buffer (OOB; `float += 1.0` produced `0`). Now computed in the promoted dtype and truncated back, matching the CPU element-wise semantics.

## Tests (`tests/gpu/Tensor_test.cpp`, `Gpu…` — no underscores per #857)
Non-contiguous `*=`/`/=` regression, host-scalar contiguous values, storage sharing, dtype preservation, `real op= complex` throws, and the non-contiguous tensor⊗tensor guard.

## Verification (RTX 4070 Ti, CUDA 13)
- 7/7 new GPU tests pass.
- 8/8 existing CPU scalar in-place tests still pass (guard doesn't change CPU behavior).
- 224 GPU arithmetic tests pass, 0 failures — including every existing `iSub`/`iMul`/`iDiv` all-types in-place case.

## Known limitation (follow-up)
General non-contiguous `tensor⊗tensor` mul/div on GPU still throws; the real fix is adding non-contiguous kernels to `cuMul`/`cuDiv` like `cuAdd`/`cuSub` have. Left for a separate PR.

### DIFF
--- a/src/Tensor.cpp
+++ b/src/Tensor.cpp
@@ -547,20 +547,20 @@ namespace cytnx {
   }
 
   namespace {
-    // Wrap a scalar as a shape-{1} Tensor on `device` so scalar in-place
-    // arithmetic reuses the iAdd/iSub/iMul/iDiv kernels (which broadcast a
-    // length-1 RHS and mutate LHS storage in place). See #906.
+    // Wrap a scalar as a host-resident shape-{1} Tensor so scalar in-place arithmetic reuses the
+    // iAdd/iSub/iMul/iDiv kernels (which broadcast a length-1 RHS and mutate LHS storage in
+    // place). See #906. The wrapper deliberately stays on the CPU even when the LHS is on a GPU:
+    // for a length-1 RHS the GPU kernels read the scalar with a host-side dereference and pass it
+    // in by value, so keeping it on the host avoids a per-call H2D copy. See #988.
     template <class T>
-    Tensor _scalar_as_rank1_tensor(const T &rc, const int device) {
+    Tensor _scalar_as_rank1_tensor(const T &rc) {
       Tensor s({1}, Type.cy_typeid(rc), Device.cpu);
       s.storage().at<T>(0) = rc;
-      if (device != Device.cpu) s = s.to(device);
       return s;
     }
-    Tensor _scalar_as_rank1_tensor(const Scalar &rc, const int device) {
+    Tensor _scalar_as_rank1_tensor(const Scalar &rc) {
       Tensor s({1}, rc.dtype(), Device.cpu);
       s.item() = rc;  // Sproxy assignment
-      if (device != Device.cpu) s = s.to(device);
       return s;
     }
   }  // namespace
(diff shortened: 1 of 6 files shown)


pcchen (commented):
Reviewed the diff and **built the branch + ran it on a CUDA device (Tesla V100, CUDA 12.9)** to check the GPU-only paths — note CI has no CUDA runner, so none of this code (incl. the 7 new GPU tests) is exercised by CI; correctness rests on local hardware runs.

**Overall: solid, well-motivated fixes. Two issues below, both reproduced on the V100.**

### What's good
- `cudaSetDevice(Rt.device()) → Lt.device())` is required now that the scalar stays host-resident (`Rt.device()==cpu==-1` would be an invalid device). ✅
- Dropping the per-call H2D scalar copy (and the latent device-pointer-deref UB it removes). ✅
- Device-independent `real op= complex` guard — throws early on both paths instead of corrupting the real buffer. ✅
- The non-contiguous scalar regression fix (route size-1 RHS through the `rconst` kernel, throw loudly on genuine non-contiguous `tensor⊗tensor`) is the right call, and the comments are genuinely excellent.

---

### ⚠️ 1. Narrow-LHS scalar path silently breaks storage sharing on GPU (CPU/GPU divergence)
The narrow-LHS guard does `Lt = promoted.astype(Lt.dtype())`, which **rebinds `Lt` to fresh storage**, so an aliased handle stops tracking the update. The CPU path applies `lhs[idx] op= rhs[idx]` straight into the existing buffer and preserves sharing. Measured:

```
CPU  shared_after=1  b[0]=2   # aliased handle sees a *= 2.0
GPU  shared_after=0  b[0]=1   # aliased handle does NOT
```
(`a = ones({4}, Float, dev); b = from_storage(a.storage()); a *= 2.0;`)

This diverges from the "preserve storage sharing" invariant of the parent chain (#906/#980). The current tests miss it: `GpuScalarInplaceKeepsStorageSharing` uses a `Double` LHS (non-narrow) and `GpuScalarInplacePreserveDtype` uses a `Float` LHS but never checks aliasing. Suggest copying the truncated result back into `Lt`'s existing storage in place rather than rebinding, plus a narrow+aliased test.

### ⚠️ 2. The narrow-LHS memory-corruption fix only covers *host scalars*, not a GPU-resident RHS
The guard is gated on `rhs_is_host_scalar`, but the real danger condition is `Lt.dtype() > Rt.dtype()` on GPU. The shared contiguous branch still does `tmpo = Lt.clone()` (narrow) while the kernel writes promoted width, so a same-shape mixed-dtype op corrupts — `float_gpu *= double_gpu` (RHS all 3.0):

```
no throw; a = 0 2.125 0 2.125    # expected 3 3 3 3  → out-of-bounds garbage
```

This is **pre-existing** (the `clone()` predates this PR — not a regression) and technically outside #988's scalar scope, but it's the exact corruption class the PR's point 4 describes, just via a non-host RHS. Fixing at the source — `tmpo = Lt.astype(type_promote(Lt, Rt))` in the shared branch — would close all paths at once and let the special-case guard be dropped. At minimum worth tracking, since `a *= b` with mixed dtype on GPU silently corrupts.

---

### Minor
- `iAdd(gpu_tensor, cpu_shape{1}_tensor)` now silently succeeds (treated as a broadcast scalar) instead of erroring on device mismatch — intended for the wrapper, but a slight public-API broadening worth a doc note.
- Heavy verbatim duplication of the ~14-line guard block across all four files (and the non-contig scalar kernel block between iMul/iDiv); a shared helper would reduce divergence risk.
- Base is the unmerged `fix/scalar-inplace-ops-preserve-storage` — needs retargeting to `master` before merge (already noted in the description).

<sub>🤖 Reviewed with Claude Code; GPU findings reproduced on a Tesla V100 / CUDA 12.9 build of this branch.</sub>

pcchen:
## Code Review

**LGTM on substance — and this PR materially matters for #980:** it fixes a real UB that #980 introduces on CUDA. #980's `{1}`-wrapper does `.to(device)`, but the GPU kernels read the length-1 scalar with a **host-side `_Rin[0]` dereference** — dereferencing a device pointer on the host. There is no GPU CI, so checks can't see it; this PR was validated on real hardware (RTX 4070 Ti, 224 GPU tests green). **Consequence: #980 and #992 should land together (or back-to-back)** — #980 alone leaves CUDA scalar in-place broken.

### Verified in the diff (all four i-ops)
- **Host-resident wrapper** (drops `.to(device)`) + `rhs_is_host_scalar` device-check exemption + `cudaSetDevice(Lt.device())` (the old code targeted `Rt.device()` — wrong once the RHS is host-resident). ✅
- **`real ⊕= complex` device-independent guard** — the CPU visit already threw, but the GPU path bypassed it and reinterpreted a real output buffer as complex (memory corruption). Now rejected uniformly, matching #980's CPU semantics. ✅
- **Narrow-LHS OOB fix:** the GPU in-place kernels write the *promoted-width* result into the LHS buffer → out-of-bounds write when the LHS is narrower. The detour computes at the promoted dtype then truncates back — matching the CPU/#980 value semantics — and the recursion terminates (the promoted LHS is no longer narrower). ✅
- **Non-contiguous:** a scalar RHS on a non-contiguous GPU LHS now routes through the rconst path (layout mappers are irrelevant for scale-every-element — the #988 regression fix); genuine non-contiguous `tensor⊗tensor` mul/div still throws, now with an accurate message (cuMul/cuDiv ignore the mappers; routing would silently mispair elements). The known-limitation note (add real non-contig kernels later) is honest. ✅

### Notes (non-blocking)
1. **🟡 Residual, pre-known (#906-tracked):** the narrow-LHS GPU path does `Lt = promoted.astype(...)` — a rebind, so storage-sharing views detach in that specific case (CPU preserves sharing). This extends the already-documented "GPU can rebind on promotion" limitation rather than introducing a new class — but deserves a one-liner in the PR body.
2. **🟠 Process:** the branch still contains the **pre-rebase #980 commits** (the base branch was force-pushed during its conflict resolution), so this needs a cherry-pick of its own commit onto the current `fix/scalar-inplace-ops-preserve-storage` head — happy to do that.
3. GPU tests can't run in CI (#538) — local-hardware evidence only; same standing caveat as the MKL paths (#583).

pcchen (approved):
Approving. Verified across all four i-ops: the host-resident `{1}` wrapper removes the CUDA host-deref UB that #980 alone would have shipped (plus the per-call H2D copy), the `rhs_is_host_scalar` device-check exemption and `cudaSetDevice(Lt.device())` fix are correct, the device-independent `real ⊕= complex` guard closes a GPU memory-corruption path, the narrow-LHS promote-compute-truncate-back detour eliminates the OOB write while matching CPU value semantics, and the non-contiguous scalar regression is fixed via the mapper-irrelevant rconst path (with genuine non-contiguous tensor⊗tensor mul/div now failing loudly instead of silently mispairing — honest follow-up noted for real kernels). Hardware-validated on RTX 4070 Ti (224 GPU tests); CPU-side CI fully green on master base including BuildAndTest on all platforms — the only red is advisory codecov/patch, structural for GPU-only lines with no GPU CI (#538/#583). The residual narrow-LHS rebind (views detach in that GPU case) is the pre-known #906-class limitation.

Issue #946, opened 2026-06-27
[build] FindCUTENSOR/FindCUQUANTUM hard-code an outdated directory layout; tarball (2.x) and apt installs both break


pcchen:
## Summary

`cmake/Modules/FindCUTENSOR.cmake` hard-codes a version-numbered library subdirectory (`$CUTENSOR_ROOT/lib/<cudaver>/`, e.g. `lib/12/`) derived from the CUDA major/minor version (`FindCUTENSOR.cmake:33-41`). This layout only matches **old cuTENSOR 1.x tarballs**. Both the **current 2.x tarball** and **apt** installs use different structures, so the finder fails to locate the libraries.

## The three layouts

**1. Old tarball (cuTENSOR 1.x) — what the finder assumes**
```
$CUTENSOR_ROOT/
  include/
  lib/11/libcutensor.so
  lib/12/libcutensor.so
```

**2. New tarball (cuTENSOR 2.x) — version subdir is gone**
```console
$ ls -l ~/src/libcutensor-linux-x86_64-2.7.0.5_cuda12-archive
include
lib
LICENSE
```
Libraries now live directly under `lib/` (no `12/` subdirectory). The finder's `set(CUTNLIB_DIR "${CUTNLIB_DIR}12")` builds a path of `lib/12/` that **does not exist**, so `find_library` returns `NOTFOUND`.

**3. apt (Ubuntu 22.04) — multiarch path, split runtime/dev packages**
```console
$ dpkg -L libcutensor2-cuda-12
/usr/lib/x86_64-linux-gnu/libcutensor/12/libcutensor.so.2.7.0
/usr/lib/x86_64-linux-gnu/libcutensor/12/libcutensorMg.so.2.7.0
/usr/lib/x86_64-linux-gnu/libcutensor/12/libcutensorMp.so.2.7.0
/usr/lib/x86_64-linux-gnu/libcutensor/12/libcutensor.so.2
...
```
- Libraries under `/usr/lib/x86_64-linux-gnu/libcutensor/12/` (multiarch triplet + `libcutensor/` component).
- Headers ship in a separate `-dev` package (`/usr/include/`).
- The unversioned `libcutensor.so` symlink exists **only** in the `-dev` package; the runtime package ships only `libcutensor.so.2` / `.so.2.7.0`, which `find_library(NAMES "cutensor")` won't resolve by default.
- Ships an extra `libcutensorMp` (multi-process), in addition to `libcutensorMg`.

[...]

Issue #945, opened 2026-06-27
[build] Clarify compiler/CUDA/cuTENSOR requirements and fix FindCUTENSOR.cmake


pcchen:
## Summary

Three inter-related build correctness issues surfaced during the v1.1.0 documentation update, plus an open question on SM 7.0 / CUDA 11 support going forward.

---

### 1. gcc >= 13 is a hard requirement (not just a recommendation)

`pybind/symmetry_py.cpp` and `pybind/unitensor_py.cpp` both include `<format>`, which is only available in C++20 with gcc >= 13 (or equivalent clang). The current docs say:

> gcc >= 13 (recommended latest or equivalent clang)

This should be stated as a hard minimum, not a recommendation, to avoid silent build failures on gcc 11/12 where `<format>` is absent or incomplete.

**Fix:** change "recommend" → "required" in `adv_install.rst` and `installation.rst`.

---

### 2. cuTENSOR v2.3.0 drops SM 7.0 (V100) and CUDA 11 support

cuTENSOR 2.3.0 release notes state that SM 7.0 (Volta / V100) and CUDA Toolkit 11 are no longer supported. Our default `CMAKE_CUDA_ARCHITECTURES` currently includes `70-real` (`CMakeLists.txt:210`), and `FindCUTENSOR.cmake` has subdirectory logic for CUDA 10 and 11.

**Decision needed:**

**Option A — Drop SM 7.0 and CUDA 11, require cuTENSOR >= 2.3.0**
- Remove `70-real` from `CMAKE_CUDA_ARCHITECTURES`
- Remove CUDA 10/11 subdirectory branches from `FindCUTENSOR.cmake`
- Minimum: CUDA 12 + cuTENSOR >= 2.3.0

**Option B — Keep SM 7.0 / CUDA 11, pin cuTENSOR < 2.3.0**
- Add an upper-bound check in `FindCUTENSOR.cmake` and document the constraint
- Keeps V100 users working but prevents adopting newer cuTENSOR features

Recommended: **Option A**. CUDA 11 is EOL and V100 adoption is declining. cuTENSOR >= 2.3.0 with CUDA 12 is the practical baseline for new deployments.

---

### 3. `FindCUTENSOR.cmake` causes build failure when `libcutensor`/`libcutensorMg` are not found

[...]

Issue #962, opened 2026-07-01
[build] Decide supported CUDA toolkit under cuTENSOR >= 2.0 + cuQuantum >= 24.0 (CUDA 11? CUDA 13?)


pcchen:
## Summary

Decide (and document) which CUDA Toolkit version(s) Cytnx supports for GPU builds, given the intended dependency floors **cuTENSOR ≥ 2.0** and **cuQuantum ≥ 24.0**. The short answer is **CUDA 12 is the only universally-supported toolkit**; CUDA 11 and CUDA 13 each require *changing* one or both floors. This issue lays out the matrix so we can make an explicit call.

Focused follow-up to the "SM 7.0 / CUDA 11 going forward" open question in #945; related: #950, #580.

## What each dependency supports (from NVIDIA release notes)

**cuTENSOR 2.x → CUDA Toolkit** (latest 2.7.0):

| cuTENSOR | CUDA | Min GPU |
| --- | --- | --- |
| 2.0.0 – 2.2.x | 11 + 12 | SM 7.0 (Volta) |
| **2.3.0**+ (→ 2.7.x) | **12 + 13** (CUDA 11 **removed**) | SM 7.5 (Turing; SM 7.0 removed) |

**cuQuantum 24.x/25.x → CUDA Toolkit** (SDK `YY.MM`; cuStateVec / cuTensorNet):

| cuQuantum SDK | CUDA |
| --- | --- |
| 24.03 / 24.08 / 24.11 | **11 + 12** |
| 25.03 | 12 |
| ~25.09 → 25.11 (latest) | **12 + 13** |

(cuTensorNet 2.4.0 in cuQuantum 24.03 itself requires cuTENSOR ≥ 2.0.1, so the two floors are mutually consistent.)

## Intersection under `cuTENSOR ≥ 2.0` + `cuQuantum ≥ 24.0`

| CUDA | cuTENSOR ≥ 2.0 | cuQuantum ≥ 24.0 | Usable? |
| --- | --- | --- | --- |
| **11** | only 2.0.0–2.2.x (dropped in 2.3.0) | only 24.x (dropped in 25.x) | ⚠️ only at the low end of *both* ranges |
| **12** | all of 2.x ✅ | all 24.x + 25.x ✅ | ✅ **yes — for any qualifying versions** |
| **13** | only cuTENSOR ≥ 2.3.0 | only cuQuantum ≈ ≥ 25.09 | ❌ not with the floor versions |

[...]

Issue #948, opened 2026-06-28
[build] Installed CUDA extension is not rpath-pinned to the build toolkit; runtime loads system/apt CUDA runtime


pcchen:
## Summary

`CMAKE_CUDA_COMPILER` only controls **link-time** toolkit selection. On Linux, the installed `pycytnx` extension gets **no rpath to the CUDA toolkit**, so at runtime the dynamic loader resolves the CUDA runtime libraries (`libcudart.so.12`, `libcublas.so.12`, …) from the system/`ldconfig` cache. On a machine with CUDA installed via **both apt and a tarball** (e.g. `/usr/local/cuda-12`), a `pip install` built against the tarball will silently load the **apt-installed** CUDA runtime — even when the user passes `-DCMAKE_CUDA_COMPILER=/usr/local/cuda-12/bin/nvcc`.

## Reproduction scenario

- CUDA toolkit present both ways: apt (`/usr/lib/x86_64-linux-gnu/libcudart.so.12`, in the `ldconfig` cache) and tarball (`/usr/local/cuda-12/lib64`).
- Build:
  ```
  pip install . --config-settings=cmake.args=-DCMAKE_CUDA_COMPILER=/usr/local/cuda-12/bin/nvcc ...
  ```
- Expectation: cytnx uses the tarball CUDA runtime at runtime.
- Actual: the installed module loads the apt CUDA runtime.

### Link-time is correct
`enable_language(CUDA)` (`CytnxBKNDCMakeLists.cmake:137`) runs **before** `find_package(CUDAToolkit)` (`CytnxBKNDCMakeLists.cmake:138`). Because the CUDA language is enabled, `find_package(CUDAToolkit)` derives the toolkit root from `CMAKE_CUDA_COMPILER`, so the `CUDA::cudart/cublas/cusolver/...` imported targets (`CytnxBKNDCMakeLists.cmake:189-190`) resolve to `/usr/local/cuda-12/lib64`. ✅

But linking against `/usr/local/cuda-12/lib64/libcudart.so` only records the **SONAME** (`libcudart.so.12`) as a `DT_NEEDED` entry — not an absolute path. Runtime resolution is then governed by rpath / `LD_LIBRARY_PATH` / `ldconfig`.

### Runtime is not pinned (Linux)
`CMakeLists.txt:428-441`:
```cmake
if(APPLE)
    set_target_properties(pycytnx PROPERTIES
      BUILD_WITH_INSTALL_RPATH TRUE
[...]

PR #993, opened 2026-07-06
fix(cmake): make cuTENSOR/cuQuantum finders honest and version-aware

pcchen:
First split of #950 (the finder-correctness portion). No CMakeLists.txt or
compiler/version-floor changes — those remain separate follow-up splits.

## Changes
- **Honest `*_FOUND`.** `CUTENSOR_FOUND`/`CUQUANTUM_FOUND` are now derived from
  the actual `find_library` results via `find_package_handle_standard_args`,
  instead of `set(... TRUE)` unconditionally — which let a `NOTFOUND` silently
  pass a caller's `REQUIRED` check and link an empty / `-NOTFOUND` entry.
- **Version-aware cuTENSOR layout.** Search both the cuTENSOR 2.x flat `lib/`
  and the legacy per-CUDA-major `lib/<major>` via `PATH_SUFFIXES`, dropping the
  dead `lib/10.2`/`lib/11` branches; report the directory actually found.
- **cuTENSOR ≥ 2.0 enforced** by reading `CUTENSOR_MAJOR/MINOR` from the headers
  and failing early on the 1.x API.
- **FindCUQUANTUM** builds `CUQUANTUM_LIBRARIES` conditionally (parity).

Addresses the finder portions of #945 / #946.

## Verification
Extracted verbatim from the already-reviewed #950 diff; both modules parse
cleanly. These finders only run under `USE_CUTENSOR`/`USE_CUQUANTUM` (i.e. a
CUDA build), which can't be exercised on the author's macOS machine — a
reviewer with a CUDA box please sanity-check a `-DUSE_CUTENSOR=ON`
(and `-DUSE_CUQUANTUM=ON`) configure.


pcchen:
Thanks — addressed all three in `6650adc`:

- **cuTENSOR CUDA-11 rejection (codex, P2):** dropped the `CUDA >= 12` `FATAL_ERROR`. It's a version-floor *policy* that belongs in the separate `CMakeLists.txt`/#962 split, and as written it rejected valid CUDA 11 + cuTENSOR 2.x installs. The generic `lib/${CUDAToolkit_VERSION_MAJOR}` suffix already covers 11/12/13, so this finder now stays policy-free; the CUDA-floor decision stays in #962.
- **Malformed `CUTENSOR_VERSION` (gemini):** `CUTENSOR_MINOR` now defaults to `0` when the header line is absent/unmatched, so the `VERSION_VAR` comparison can't see a `"2."`-style version.
- **`CUQUANTUM_LIBRARY_DIRS` (gemini):** now derived from the found library via `get_filename_component`, mirroring FindCUTENSOR, instead of the earlier guessed `${CUQUANTUM_ROOT}/${CUTNLIB_DIR}`.

Note this PR (first split of #950) intentionally keeps only finder correctness + layout + the cuTENSOR ≥ 2.0 API requirement — no CMakeLists.txt / CUDA-version-floor changes. As before, it can't be built on a CUDA-less machine; a reviewer with a CUDA box please sanity-check a `-DUSE_CUTENSOR=ON` configure.

cmake/Modules/FindCUTENSOR.cmake
```diff
@@ -27,22 +27,56 @@ else()
...
-elseif((${CUDAToolkit_VERSION_MAJOR} GREATER_EQUAL 11) AND (${CUDAToolkit_VERSION_MAJOR} LESS 12) AND (${CUDAToolkit_VERSION_MINOR} LESS_EQUAL 0))
-  set(CUTNLIB_DIR "${CUTNLIB_DIR}11.0")
-elseif((${CUDAToolkit_VERSION_MAJOR} GREATER_EQUAL 11) AND (${CUDAToolkit_VERSION_MAJOR} LESS 12) AND (${CUDAToolkit_VERSION_MINOR} GREATER_EQUAL 1))
-  set(CUTNLIB_DIR "${CUTNLIB_DIR}11")
-elseif((${CUDAToolkit_VERSION_MAJOR} GREATER_EQUAL 12))
-  set(CUTNLIB_DIR "${CUTNLIB_DIR}12")
-endif()
+# Search both cuTENSOR library layouts: 2.x tarballs place the libraries
+# directly under lib/, while older tarballs and apt use a per-CUDA-major
+# subdir (lib/<cuda-major>, e.g. lib/11, lib/12, lib/13). Listing both as
```
IvanaGyro, line 32: This layout was changed since 2.3 not 2.x

cmake/Modules/FindCUTENSOR.cmake
```diff
@@ -27,22 +27,56 @@ else()
...
+# find_library PATH_SUFFIXES resolves either layout for any CUDA major. The
+# older minor-specific lib/10.2 and lib/11.0 special-cases were removed; the
```
IvanaGyro, line 34: This comment is not useful for the future. It can be put in the commit message.

cmake/Modules/FindCUTENSOR.cmake
```diff
@@ -27,22 +27,56 @@ else()
...
+# generic lib/<major> covers them (apt multiarch paths remain, issue #946).
+# The CUDA-version floor is a separate policy decision (issue #962), enforced
+# in CMakeLists.txt rather than gated here, so this finder stays policy-free.
+set(CUTNLIB_DIR lib lib/${CUDAToolkit_VERSION_MAJOR})
```
IvanaGyro, line 38: `CUTNLIB_DIR` is not used in the new implementation.

yingjerkao:
@pcchen can you update the PR based the review comments? /

yingjerkao:
Verified on a CUDA box, as the PR description asked for.

**Environment**

- NVIDIA RTX 4070 Ti SUPER (`sm_89`), CUDA **13.0** (`nvcc` V13.0.88), CMake 4.0, GCC 13.3
- cuTENSOR **2.4.1** via apt (`/usr/lib/x86_64-linux-gnu/libcutensor/{12,13}/`) and **2.2.0** via conda
- cuQuantum: cuTensorNet **2.10** / cuStateVec via apt, and 2.7.0 / 1.8.0 via conda

## 1. Finder matrix — isolated

A minimal probe project calling `find_package(CUTENSOR|CUQUANTUM REQUIRED)`, with `CMAKE_MODULE_PATH`
pointed at either **current `master`**'s modules or this PR's, nothing else varied. (Both roots
passed as cache vars with `CUTENSOR_ROOT`/`CUQUANTUM_ROOT` cleared from the environment — see the
note at the bottom.)

| layout | `master` | this PR |
| --- | --- | --- |
| cuTENSOR 2.x flat `lib/` | libs found, but `CUTENSOR_LIBRARY_DIRS` reports `<root>/lib/12` — **a directory that does not exist** | found; `LIBRARY_DIRS` = the real `<root>/lib`; version 2.4 detected |
| cuTENSOR legacy `lib/13` | **`CUTENSOR_FOUND=TRUE` with `CUTENSOR_LIB-NOTFOUND`** | found correctly at `<root>/lib/13` |
| cuTENSOR apt multiarch (`CUTENSOR_ROOT=/usr`) | **`CUTENSOR_FOUND=TRUE` with `-NOTFOUND`** | fails cleanly: `Could NOT find CUTENSOR (missing: CUTENSOR_LIB) (found version "2.4")` |
| cuQuantum apt multiarch (`CUQUANTUM_ROOT=/usr`) | **`CUQUANTUM_FOUND=TRUE` with `-NOTFOUND`** | fails cleanly: `Could NOT find CUQUANTUM (missing: CUQUANTUM_TENSORNET_LIB ...)` |
| cuQuantum conda root | found, correct | found, correct |

**The PR improves every case and regresses none.** Two things worth adding to the description:

**(a) The silent-`-NOTFOUND` bug reproduces exactly as predicted.** On `master`, `CUTENSOR_LIBRARIES`
comes back as the literal string `CUTENSOR_LIB-NOTFOUND;CUTENSORMg_LIB-NOTFOUND` while
`CUTENSOR_FOUND=TRUE` and the caller's `REQUIRED` passes — so `-NOTFOUND` goes onto the link line.
Confirmed, not theoretical.

**(b) The `lib/<major>` change fixes a CUDA-13 break the description undersells.** Current `master`
hardcodes `lib/12` for *any* CUDA ≥ 12:

```cmake
elseif((${CUDAToolkit_VERSION_MAJOR} GREATER_EQUAL 12))
  set(CUTNLIB_DIR "${CUTNLIB_DIR}12")
```

so on a CUDA 13 host the legacy per-CUDA-major tarball layout (`lib/13/`) is never searched and
silently yields `-NOTFOUND`. This PR's `lib/${CUDAToolkit_VERSION_MAJOR}` is what fixes it. Given
the repo's GPU CI now runs CUDA 13 (#1044), that seems worth calling out explicitly.

## 2. End-to-end build

Because this branch is ~392 commits behind `master`, I did not build it as-is. Instead I applied
**only the two module files from this PR onto current `master` (23180260)** — which is what merging
will actually produce — and did a full CUDA build with both optional deps enabled:

```
cmake -S . -B build_full -DUSE_CUDA=ON \
  -DUSE_CUTENSOR=ON  -DCUTENSOR_ROOT=$CONDA_PREFIX \
  -DUSE_CUQUANTUM=ON -DCUQUANTUM_ROOT=$CONDA_PREFIX \
  -DCMAKE_CUDA_ARCHITECTURES=89 -DCMAKE_BUILD_TYPE=RelWithDebInfo -DRUN_TESTS=ON
```

Configure output:

```
-- cuTENSOR version: 2.2 (from .../include/cutensor.h)
-- Found CUTENSOR: .../lib/libcutensor.so (found version "2.2")
-- Build with CuTensor: YES
-- CuTensor: libdir:.../lib incdir:.../include libs:.../libcutensor.so;.../libcutensorMg.so
-- Found CUQUANTUM: .../lib/libcutensornet.so
-- Build with CuQuantum: YES
```

Both `CUTENSOR_LIBRARY_DIRS` and `CUQUANTUM_LIBRARY_DIRS` now name the directory the libraries were
actually found in, and the version gate reports a real number instead of being silent.

The compile + `gpu_test_main` run on that configuration is still in progress as I post this; I will
follow up in this thread with the result. Nothing about it affects section 1, which is where the
finder behaviour is actually pinned.

## 3. Unrelated pre-existing bug I hit while testing (not a blocker for this PR)

In both `master` and this PR, the finders prefer the **environment** variable over an explicit cache
variable:

```cmake
if(NOT DEFINED ENV{CUTENSOR_ROOT} AND NOT DEFINED CUTENSOR_ROOT)
  message(FATAL_ERROR "CUTENSOR_ROOT not set!")
else()
  if(DEFINED ENV{CUTENSOR_ROOT})
    set(CUTENSOR_ROOT "$ENV{CUTENSOR_ROOT}")   # <-- clobbers -DCUTENSOR_ROOT=...
  endif()
```

So `-DCUQUANTUM_ROOT=/some/path` is silently ignored whenever `CUQUANTUM_ROOT` happens to be
exported — which conda/miniforge environments do by default. This cost me a while to spot (I
concluded cuQuantum was unconfigurable on this box when in fact my own `-D` was being discarded).
CMake convention is the reverse: the cache variable should win, with the environment as fallback.
Happy to open a separate issue if you'd like it tracked rather than folded in here.

---

Net: **LGTM on the finder behavior** from a real CUDA 13 host. Suggest adding (b) to the PR
description since it turns this from a tidy-up into a CUDA-13 correctness fix.

yingjerkao:
Follow-up: the end-to-end build finished. **Green.**

Recall the setup — this PR's two module files applied onto current `master` (23180260), since the branch itself is ~392 commits behind, with `-DUSE_CUTENSOR=ON -DUSE_CUQUANTUM=ON` against a conda cuTENSOR 2.2.0 / cuTensorNet 2.7.0 on a CUDA 13.0 host (sm_89).

- `libcytnx.a`, `test_main` and `gpu_test_main` all build and link cleanly.
- GPU suite, `linalg_Test.*` + `Rsvd.*`: **25 passed, 1 skipped, 0 failed** (the skip is the unrelated pre-existing `GpuBkUtExpM`, which skips because `Eig` has no CUDA backend).

So the finder changes hold up in a real cuTENSOR + cuQuantum build, not just in isolation.

Two incidental observations from that run, neither caused by this PR:

**1. A cuQuantum build does not execute the #1102 failure path at all.** `Rsvd.GpuFlagCombinations` and `GesvdTruncate.GpuFlagCombinations` abort with `CUSOLVER_STATUS_EXECUTION_FAILED` from `cusolverDnDgesvdj` on a **non**-cuQuantum CUDA 13 build (that is #1102), but both pass here. The reason is that `src/linalg/Gesvd_truncate.cpp` has a `#ifdef UNI_CUQUANTUM` branch (lines 19 and 59) routing the GPU path through `cuQuantumGeSvd_internal` (cuTensorNet), so `cuGeSvd_internal.cu:244` is never reached — cuQuantum bypasses the bug rather than fixing anything. Full diagnosis, including the alternative explanations I ruled out (it is *not* a cuSOLVER version difference — `LD_DEBUG=bindings` shows both builds binding to the same CUDA 13 `libcusolver.so.12`), is in #1102.

*(Amended: this item originally read "cuQuantum flips two GPU failures to passes", which wrongly implied cuQuantum fixes something. It bypasses the call site. Corrected after tracing the mechanism.)*

**2. A conda cuQuantum drags in mismatched CUDA majors.** The link emits:

```
/usr/bin/ld: warning: libcublas.so.12, needed by .../libcutensornet.so, may conflict with libcublas.so.13
/usr/bin/ld: warning: libcusolver.so.11, needed by .../libcutensornet.so, may conflict with libcusolver.so.12
```

i.e. the conda-packaged cuTensorNet was built against CUDA 12 while the toolkit here is 13. It links and the tests pass, but it is exactly the class of runtime-library mixing #948 is about. Worth a version compatibility check somewhere, though clearly out of scope for this PR.

I have also split the environment-variable precedence bug I mentioned earlier into its own issue, #1129, rather than expand this PR's scope.

Verification summary unchanged: **LGTM.**

cmake/Modules/FindCUTENSOR.cmake
yingjerkao: `CUTNLIB_DIR` is still used — it is the `PATH_SUFFIXES` argument for both `find_library` calls on the current head (6650adc):

```
38:set(CUTNLIB_DIR lib lib/${CUDAToolkit_VERSION_MAJOR})
85:    PATH_SUFFIXES ${CUTNLIB_DIR}
92:    PATH_SUFFIXES ${CUTNLIB_DIR}
```

It changed from a single string to a two-element list, which is what makes both layouts searchable in one pass. I think this note predates that revision. Verified working on a CUDA 13 host: with the legacy `lib/13` layout the finder resolves via the `lib/${CUDAToolkit_VERSION_MAJOR}` element, and with a flat `lib/` it resolves via the first. Resolving — please reopen if I have misread you.

cmake/Modules/FindCUTENSOR.cmake
yingjerkao: Fair — I could not confirm the exact release from the installs on my test box (conda 2.2.0 ships a flat `lib/`, apt 2.4.1 uses `<multiarch>/libcutensor/13/`, but both are repackagings rather than NVIDIA tarballs), so I will defer to you on 2.3 being the boundary.

Worth noting this is a comment-accuracy point only: the finder lists **both** suffixes unconditionally, so behaviour is identical regardless of which release moved the layout. Leaving the thread open for @pcchen to correct the wording — flagging it as non-blocking rather than resolving it myself, since it is your comment.

cmake/Modules/FindCUTENSOR.cmake
yingjerkao: Agreed, and I have applied it literally: the rationale is now in the squash commit message for the merge rather than only in the source. The in-file comment can be trimmed in a follow-up — leaving this open for @pcchen rather than resolving it myself.

cmake/Modules/FindCUTENSOR.cmake
```diff
@@ -27,22 +27,54 @@ else()
...
+      break()
+    endif()
+  endif()
+endforeach()
+
+if(_cutensor_version_header)
+  file(STRINGS "${_cutensor_version_header}" _cutensor_minor_line REGEX "^#define[ \t]+CUTENSOR_MINOR[ \t]+[0-9]+")
+  string(REGEX REPLACE ".*CUTENSOR_MAJOR[ \t]+([0-9]+).*" "\\1" CUTENSOR_VERSION_MAJOR "${_cutensor_major_line}")
+  string(REGEX REPLACE ".*CUTENSOR_MINOR[ \t]+([0-9]+).*" "\\1" CUTENSOR_VERSION_MINOR "${_cutensor_minor_line}")
+  set(CUTENSOR_VERSION "${CUTENSOR_VERSION_MAJOR}.${CUTENSOR_VERSION_MINOR}")
```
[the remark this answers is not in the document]
yingjerkao: Already addressed in 6650adc — `CUTENSOR_VERSION_MINOR` now defaults to `0` when the regex does not match, so `CUTENSOR_VERSION` cannot come out malformed:

```cmake
if(_cutensor_minor_line)
  string(REGEX REPLACE ... CUTENSOR_VERSION_MINOR "${_cutensor_minor_line}")
else()
  set(CUTENSOR_VERSION_MINOR "0")
endif()
```

Confirmed on a real install: `-- cuTENSOR version: 2.4 (from /usr/include/cutensor.h)`. Resolving.

cmake/Modules/FindCUQUANTUM.cmake
```diff
@@ -49,6 +49,18 @@ find_library(
...
-set(CUQUANTUM_LIBRARIES "${CUQUANTUM_TENSORNET_LIB};${CUQUANTUM_CUSTATEVEC_LIB}")
-message(STATUS "ok")
-set(CUQUANTUM_FOUND TRUE)
+set(CUQUANTUM_LIBRARIES "")
+if(CUQUANTUM_TENSORNET_LIB)
+    list(APPEND CUQUANTUM_LIBRARIES "${CUQUANTUM_TENSORNET_LIB}")
+endif()
+if(CUQUANTUM_CUSTATEVEC_LIB)
+    list(APPEND CUQUANTUM_LIBRARIES "${CUQUANTUM_CUSTATEVEC_LIB}")
+endif()
```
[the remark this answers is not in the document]
yingjerkao: Already implemented on the current head — `FindCUQUANTUM.cmake` derives the directory from the found library rather than assuming `lib`:

```cmake
if(CUQUANTUM_TENSORNET_LIB)
  get_filename_component(CUQUANTUM_LIBRARY_DIRS "${CUQUANTUM_TENSORNET_LIB}" DIRECTORY)
endif()
```

Verified on a CUDA 13 host: `CuQuantum: libdir:/home/.../lib incdir:/home/.../include` now names the directory the library was actually found in. Resolving.

cmake/Modules/FindCUTENSOR.cmake
```diff
@@ -27,22 +27,56 @@ else()
...
+  message(STATUS "cuTENSOR version: ${CUTENSOR_VERSION} (from ${_cutensor_version_header})")
+  if(CUTENSOR_VERSION_MAJOR LESS 2)
+    message(FATAL_ERROR
+      "cuTENSOR >= 2.0 is required, but found ${CUTENSOR_VERSION} in "
+      "${CUTENSOR_ROOT}. Install cuTENSOR 2.x and point CUTENSOR_ROOT at it.")
+  endif()
+else()
+  message(WARNING
+    "Could not determine the cuTENSOR version from headers under "
+    "${CUTENSOR_INCLUDE_DIRS}; Cytnx requires cuTENSOR >= 2.0.")
```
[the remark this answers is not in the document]
yingjerkao: Still open, and I agree it is a real gap: when the headers do not expose `CUTENSOR_MAJOR`, the `else()` branch only warns and `find_package_handle_standard_args` can still succeed with `CUTENSOR_VERSION` unset.

Judging it non-blocking for this PR: it requires a cuTENSOR install whose libraries are present but whose headers are missing or unrecognised, which is a broken install rather than a supported layout — and the current behaviour (warn, proceed) is no worse than the pre-PR behaviour (no version check at all). Better handled together with the version-floor policy work in #950, which owns the "what do we require" decision. Leaving open as a tracked follow-up rather than resolving.

yingjerkao (approved):
Approving on the strength of the hardware verification above (RTX 4070 Ti SUPER, CUDA 13.0, cuTENSOR 2.4.1 apt / 2.2.0 conda, cuTensorNet 2.7.0):

- Finder matrix across five install layouts, current `master` vs this PR: **improves every case, regresses none.**
- End-to-end build of this PR's two modules applied onto current `master` with `-DUSE_CUTENSOR=ON -DUSE_CUQUANTUM=ON`: library, `test_main` and `gpu_test_main` all build; GPU suite 25 passed / 1 skipped / 0 failed.
- The silent-`-NOTFOUND` bug this PR fixes is confirmed real on `master`, not theoretical: `CUTENSOR_FOUND=TRUE` alongside `CUTENSOR_LIB-NOTFOUND`, passing a caller's `REQUIRED` check.
- The `lib/${CUDAToolkit_VERSION_MAJOR}` change additionally fixes a **CUDA-13 break** on `master`, which hardcodes `lib/12` for any CUDA >= 12 — relevant now that GPU CI runs CUDA 13 (#1044).

Three threads are left open deliberately rather than resolved:

- @IvanaGyro's two notes on the block comment (2.3 vs 2.x, and moving rationale to the commit message). Comment-accuracy only — the finder searches both suffixes unconditionally, so behaviour is unaffected. Left for @pcchen to trim; I have put the rationale in the squash commit message as requested.
- @chatgpt-codex-connector on a missing/unrecognised version header only warning. Real but narrow (a broken install), no worse than the pre-PR behaviour of no version check at all; belongs with the version-floor policy in #950.

Separately filed #1129 for the `$ENV{*_ROOT}`-clobbers-`-D` precedence bug in these same two finders, to keep it out of this PR's scope.

PR #995, opened 2026-07-06
fix(linalg): compute GPU Kron product in the promoted output dtype

### LINKED ISSUE #994: GPU build broken: cuKron_kernel fails for mixed ComplexFloat×Double (regression from #858)
## Summary

The GPU build of `master` is broken. Compiling `src/linalg/Kron.cpp` fails with a type error in the CUDA Kron kernel for mixed complex/real instantiations:

```
src/backend/linalg_internal_gpu/cuKron_internal.cuh(49): error: no operator "=" matches these operands
            operand types are: cuDoubleComplex = cuFloatComplex
          out[blockIdx.x * blockDim.x + threadIdx.x] = Lin[x] * Rin[y];

  detected during:
    instantiation of "void cytnx::linalg_internal::cuKron_kernel(TO *, ...) [with TO=cuDoubleComplex, TL=cuComplex, TR=cytnx::cytnx_double]"
    instantiation of "void cytnx::linalg_internal::cuKron_general(...) [with TO=cuDoubleComplex, TL=cuComplex, TR=cytnx::cytnx_double]" at line 90 of src/linalg/Kron.cpp
gmake[3]: *** [.../Kron.cpp.o] Error 1
```

## Root cause

Regression introduced by #858 (commit `004e433a`, "fix(Type): promote mixed complex/real dtypes by precision"). That change made mixed complex/real pairs promote by precision:

> ComplexFloat + Double now promotes to **ComplexDouble** (previously ComplexFloat).

`cuKron_general`/`cuKron_kernel` derive the output type via `type_promote_from_gpu_pointer_t<TL,TR>`, so for `(TL=ComplexFloat, TR=Double)` the output is now `cuDoubleComplex`. But the kernel computes `Lin[x] * Rin[y]` using the `operator*` overloads in `src/utils/cucomplex_arithmetic.cu`, where `cuFloatComplex * cytnx_double` returns **`cuFloatComplex`** (follows the left/complex operand's precision). Assigning that `cuFloatComplex` product into the `cuDoubleComplex` output has no matching `operator=`, hence the compile error.

| | `out_type` for (ComplexFloat, Double) | product `Lin*Rin` | result |
|---|---|---|---|
| before #858 | ComplexFloat | cuFloatComplex | compiles |
| after #858  | ComplexDouble | cuFloatComplex | **error** |

The broken instantiations are `(cuFloatComplex, cytnx_double)` and `(cytnx_double, cuFloatComplex)` (either operand order).

## Why CI didn't catch it

CI does not build the CUDA/GPU path (it needs a GPU + cuTENSOR), so the GPU-only template instantiation of Kron is never compiled. The CPU `Kron_general` path handles the conversion and compiles fine, so nothing flagged the regression when #858 merged.

Scope: only `src/linalg/Kron.cpp` uses `type_promote_from_gpu_pointer_t`, so this specific defect is isolated to the GPU Kron kernel. (Other GPU elementwise ops such as cuDiv already compute in the output type via a dispatch helper and are unaffected.)

## Reproduce

Configure with CUDA + cuTENSOR and build the `cytnx` target:

```
cmake -S . -B build -DUSE_CUDA=ON -DUSE_CUTENSOR=ON \
      -DCMAKE_CUDA_ARCHITECTURES=70   # any real arch
cmake --build build --target cytnx
```

The build fails at `Kron.cpp.o`.

## Suggested fix

Compute the product in the output type before assignment, mirroring the existing `cuDiv_dispatch.cu` pattern (`CuToComplexDouble`/`CuToComplexFloat` + `if constexpr` dispatch). i.e. replace

```cpp
out[...] = Lin[x] * Rin[y];
```

with a helper `cuKron_mul<TO>(Lin[x], Rin[y])` that promotes both operands to `TO` (`cuCmul`/`cuCmulf` for complex `TO`, `static_cast<TO>` otherwise). This makes the result assignable for every promoted `(TO, TL, TR)` triple.

## Environment

- CUDA 12.9, cuTENSOR 2.7.0, sm_70 (V100)
- master at the time of report; regression traces to #858 / `004e433a`

pcchen:
## What

Fix a GPU build break in `cuKron_kernel`: compute the Kron product in the **promoted output dtype** instead of relying on the raw `operator*` result type.

## Why

`cuKron_kernel` did `out[i] = Lin[x] * Rin[y]`. The cucomplex `operator*` overloads follow the left operand's precision, so `cuFloatComplex * double -> cuFloatComplex`. After #858 made mixed complex/real pairs promote by precision (`ComplexFloat + Double -> ComplexDouble`), the output type is `cuDoubleComplex`, and assigning the `cuFloatComplex` product to it fails to compile:

```
src/backend/linalg_internal_gpu/cuKron_internal.cuh:49: error: no operator "=" matches these operands
  operand types are: cuDoubleComplex = cuFloatComplex
```

This broke the CUDA build for the `(cuFloatComplex, cytnx_double)` instantiation (and its mirror). CI didn't catch it because it does not build the GPU path; the CPU `Kron_general` handles the conversion and compiled fine.

## How

Introduce a `cuKron_mul<TO>` helper that promotes both operands to `TO` before multiplying (`cuCmul`/`cuCmulf` for complex `TO`, `static_cast<TO>` otherwise), mirroring the existing `cuDiv_dispatch.cu` pattern. This makes the result assignable for every promoted `(TO, TL, TR)` triple.

## Testing

Built the full `cytnx` target with CUDA + cuTENSOR + cuQuantum (CUDA 12.9, sm_70 / V100) — compiles and links. Ran `Kron(ComplexFloat, Double)` on the GPU: output dtype is `ComplexDouble` and the values match the CPU reference and hand-computed expected results.

Note: this class of regression would be caught by a minimal GPU **compile** job in CI (no GPU hardware needed to compile the templates). Worth considering as a follow-up.

Fixes #994

### DIFF
--- a/src/backend/linalg_internal_gpu/cuKron_internal.cuh
+++ b/src/backend/linalg_internal_gpu/cuKron_internal.cuh
@@ -9,7 +9,9 @@
 #include "cytnx_error.hpp"
 #include "Type.hpp"
 #include "backend/lapack_wrapper.hpp"
+#include <cuComplex.h>
 #include <iostream>
+#include <type_traits>
 
 #ifndef __CUDACC__
   #error This file requires a CUDA compiler
@@ -18,6 +20,42 @@
 namespace cytnx {
 
   namespace linalg_internal {
+
+    // Promote an operand to the Kron output type before multiplying. The raw
+    // `operator*` overloads follow the left operand's precision (e.g.
+    // cuFloatComplex * double -> cuFloatComplex), which no longer matches the
+    // promoted output dtype after mixed complex/real pairs promote by precision
+    // (ComplexFloat + Double -> ComplexDouble, see #858). Computing the product
+    // in TO keeps the result assignable to `out`, mirroring cuDiv_dispatch.
+    template <typename T>
+    __device__ inline cuDoubleComplex cuKron_to_cd(const T &v) {
+      return make_cuDoubleComplex(static_cast<cytnx_double>(v), 0.0);
+    }
+    __device__ inline cuDoubleComplex cuKron_to_cd(const cuDoubleComplex &v) { return v; }
+    __device__ inline cuDoubleComplex cuKron_to_cd(const cuComplex &v) {
+      return cuComplexFloatToDouble(v);
+    }
+
+    template <typename T>
+    __device__ inline cuComplex cuKron_to_cf(const T &v) {
+      return make_cuFloatComplex(static_cast<cytnx_float>(v), 0.0f);
+    }
+    __device__ inline cuComplex cuKron_to_cf(const cuComplex &v) { return v; }
+    __device__ inline cuComplex cuKron_to_cf(const cuDoubleComplex &v) {
+      return make_cuFloatComplex(static_cast<cytnx_float>(cuCreal(v)),
+                                 static_cast<cytnx_float>(cuCimag(v)));
+    }
+
+    template <typename TO, typename TL, typename TR>
+    __device__ inline TO cuKron_mul(const TL &lhs, const TR &rhs) {
+      if constexpr (std::is_same_v<TO, cuDoubleComplex>) {
+        return cuCmul(cuKron_to_cd(lhs), cuKron_to_cd(rhs));
+      } else if constexpr (std::is_same_v<TO, cuComplex>) {
+        return cuCmulf(cuKron_to_cf(lhs), cuKron_to_cf(rhs));
+      } else {
+        return static_cast<TO>(lhs) * static_cast<TO>(rhs);
+      }
+    }
 
     template <class TO, class TL, class TR>
     __global__ void cuKron_kernel(TO *out, const TL *Lin, const TR *Rin, cytnx_uint64 *meta_infos,
@@ -46,7 +84,7 @@ namespace cytnx {
           x += cytnx_uint64(tmp2 / info[offset2 + j]) * info[len_nsa + j];
           y += cytnx_uint64(tmp2 % info[offset2 + j]) * info[offset1 + j];
         }
-        out[blockIdx.x * blockDim.x + threadIdx.x] = Lin[x] * Rin[y];
+        out[blockIdx.x * blockDim.x + threadIdx.x] = cuKron_mul<TO>(Lin[x], Rin[y]);
       }
     }
 


pcchen:
**Priority note:** `master`'s GPU build currently does **not compile** without this — `Kron.cpp` fails to build for the `(cuFloatComplex, cytnx_double)` instantiation after #858 changed mixed complex/real promotion (see #994). So this effectively blocks building/testing *any* GPU code on `master`, including the GPU paths in the scalar-in-place cluster (#980 / #992 / #937). Suggest prioritizing it.

Independently verified: built the full `cytnx` target with CUDA (CUDA 12.9, `sm_70` / Tesla V100) and ran `Kron(ComplexFloat, Double)` on-device — output dtype is `ComplexDouble` and values match the CPU reference. Also confirmed the pre-fix build fails at `Kron.cpp.o` and this change makes it compile + link + run.

Note that CI has no CUDA runner, so none of this is exercised by CI — worth keeping in mind for the whole GPU cluster.

ianmccul:
From discussion in #999, the actual root cause here is the Cytnx declarations in `include/utils/cucomplex_arithmetic.hpp,` definitions in `src/utils/cucomplex_arithmetic.cu`, that are incorrect with respect to any sane promotion rules:
```c++
 __host__ __device__
cuFloatComplex operator*(const cuFloatComplex &ln, const cytnx_double &rn) {
  return cuCmulf(make_cuFloatComplex(rn, 0), ln);
}

__host__ __device__
cuFloatComplex operator*(const cytnx_double &rn, const cuFloatComplex &ln) {
  return cuCmulf(make_cuFloatComplex(rn, 0), ln);
}
```
It is better to fix these, rather than patch `Kron` only, since they are used elsewhere; `cuOuter`, `iMul`, etc.

Even better would be to replace the old `cuXXXX` types with `cuda::std::complex`, which works with CUDA >= 11.3.

yingjerkao:
Superseded — closing. master already computes the Kron product in the promoted output dtype via `out[i] = TO(Lin[x]) * TO(Rin[y])` (landed in #1004, commit 7bbd8141, as part of the `cuda::std::complex` migration), and #1019 removed the `cucomplex_arithmetic` overloads this change worked around. The build break this fixed (#994/#999) is resolved on master, so this standalone fix is no longer needed. Thanks!

PR #996, opened 2026-07-06
chore(gitignore): ignore /.deps/ for local cuTENSOR/cuQuantum installs

pcchen:
Local, no-root installs of the CUDA math libs (e.g. NVIDIA pip wheels extracted under `.deps/cutensor` and `.deps/cuquantum`, pointed at via `CUTENSOR_ROOT`/`CUQUANTUM_ROOT`) drop hundreds of MB into the tree. Ignore `/.deps/` so they never show up in `git status`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,6 @@ GEMINI.local.md
 .antigravity/
 # Cursor
 .cursor/
+
+# Local CUDA math libs installed for testing (cuTENSOR / cuQuantum)
+/.deps/


pcchen:
Closing: `.deps/` isn't a project convention — it only exists from a local-install choice. Better handled per-developer via `.git/info/exclude` (or a global gitignore) than by adding a workflow preference to the shared .gitignore.

Issue #934, opened 2026-06-23
[BUG] `UniTensor` exposes raw element-wise array operations that are not meaningful tensor-network operations


ianmccul:
`UniTensor` should represent a tensor-network object with legs, bonds, rowrank, symmetry sectors, and possibly fermionic sign structure. But the current API exposes a large part of the raw `Tensor` element-wise array API directly on `UniTensor`.

For dense tensors this is already questionable. For symmetric, block-sparse, or fermionic `UniTensor`, many of these operations are not merely unusual; they are not well-defined tensor-network operations. ***The fact that these operations exist is a bug***, and not in the sense of "deprecate this API in the next release, remove these functions the release after", they are a **serious correctness issue**. Any user code that uses these functions is a bugged. There are no legitimate uses. If anyone really wants this operation then they are not working with a UniTensor, but some other kind of data that is incorrectly wrapped in a UniTensor.

## The worst offenders

The clearest example is scalar addition:

```c++
virtual void Add_(const Scalar &rhs);
```

What should `UniTensor + scalar` mean? It is not adding a scalar multiple of an identity tensor. It is not a tensor-network contraction. It is not a tensor product. It is just raw element-wise array arithmetic.

For dense `UniTensor`, Cytnx literally adds the scalar to every stored element:

```c++
void DenseUniTensor::Add_(const Scalar &rhs) {
  this->_block += rhs;
}
```

For block and fermionic block `UniTensor`, the implementation correctly refuses the operation:

```c++
void Add_(const Scalar &rhs) {
  cytnx_error_msg(
    true,
    "[ERROR] Cannot perform elementwise arithmetic '+' between Scalar and BlockUniTensor.\n %s "
    "\n",
    "This operation would destroy the block structure. [Suggest] Avoid or use get/put_block(s) "
    "to do operation on blocks.");
}
```

That error message is right. The problem is that the API still exposes this as a normal `UniTensor` operation, and dense `UniTensor` performs it. This is exactly the conceptual bug: `UniTensor` is being treated as a raw array when it should be treated as a structured tensor-network object.

Some arithmetic operations are legitimate. Scalar multiplication is fine. Tensor addition/subtraction is fine if the tensor spaces/bonds match exactly. But many of the exposed element-wise operations should not exist on `UniTensor`.

## Element-wise multiplication of two `UniTensor`s

```c++
virtual void Mul_(const boost::intrusive_ptr<UniTensor_base> &rhs);
```

[...]

Issue #753, opened 2026-03-26
Missing labels after + and -


chiamin:
In A=B+C, the label of A will be reset to ['0','1',...] regardless of the labels of B and C. Is this the correct behavior we want? I propose that we should require the labels to be the same for B and C, and set the same labels for A.

Here is a minimal example.
[bug_report_label_arithmetic.py](https://github.com/user-attachments/files/26262409/bug_report_label_arithmetic.py)

Issue #675, opened 2025-09-16
Copy labels in elementwise arithmetics


manuschneider:
[Current]
Add/Sub/Mul/Div (non-inline versions) between two UniTensors return a new UniTensor with labels set to '0','1','2',....

[Suggestion]
1) Check if the labels of the two input tensors are the same, throw an error otherwise (should be done in the inline versions for consistency)
2) Set the output labels to these input labels (or remove the line `out.set_labels(vec_range<std::string>(Lt.rank()));`)

[Improvement]
1) Additional safety, since input tensors need to have same labels and not just compatible bonds
2) Avoid unexpected behavior (output labels are not expected to be '0','1','2')
3) Create consistency between inline and non-inline versions

[Affects]
```
/src/linalg/Add.cpp - cytnx::UniTensor Add(const cytnx::UniTensor &Lt, const cytnx::UniTensor &Rt)
/src/linalg/Sub.cpp - cytnx::UniTensor Sub(const cytnx::UniTensor &Lt, const cytnx::UniTensor &Rt)
/src/linalg/Mul.cpp - cytnx::UniTensor Mul(const cytnx::UniTensor &Lt, const cytnx::UniTensor &Rt)
/src/linalg/Div.cpp - cytnx::UniTensor Div(const cytnx::UniTensor &Lt, const cytnx::UniTensor &Rt)
/src/DenseUniTensor.cpp - void DenseUniTensor::Add_(const boost::intrusive_ptr<UniTensor_base> &rhs) - similarly Sub_ Mul_ Div_
/src/BlockUniTensor.cpp - void BlockUniTensor::Add_(const boost::intrusive_ptr<UniTensor_base> &rhs) - similarly Sub_ Mul_ Div_
/src/BlockFermionicUniTensor.cpp - void BlockFermionicUniTensor::Add_(const boost::intrusive_ptr<UniTensor_base> &rhs) - similarly Sub_ Mul_ Div_
```

Issue #941, opened 2026-06-25
Its time: Typed Storage Dispatch


ianmccul:
This is a project that would be very difficult without a LLM to assist, but now I expect takes @yingjerkao a day or two....  Basically, the task is to point an LLM at this issue and let it go. Maybe get it to review existing Cytnx codebase first.

# Typed Storage Dispatch

Cytnx already has typed storage objects:

```cpp
template <typename DType>
class StorageImplementation : public Storage_base { ... };
```

but most internal APIs immediately erase them to `boost::intrusive_ptr<Storage_base>`, recover the dtype through an integer `dtype()`, and then cast `void *data()` by hand. This is the root cause of the mixed-dtype arithmetic bugs: a dispatch table can select a kernel whose C++ output type is `double`, while the actual output storage object is still `StorageImplementation<int16_t>`.

The goal of this refactor is simple:

> Once an operation has dispatched on dtype, arithmetic kernels should receive typed storage objects. It should be impossible to pass Int16 storage to a kernel that writes Double.

Concretely, backend kernels that are currently shaped like:

```cpp
void Kernel(boost::intrusive_ptr<Storage_base> &out,
            boost::intrusive_ptr<Storage_base> &lhs,
            boost::intrusive_ptr<Storage_base> &rhs,
            ...);
```

should become templates over typed storage pointers:

```cpp
template <typename TO, typename TL, typename TR>
void Kernel(boost::intrusive_ptr<StorageImplementation<TO>> out,
            boost::intrusive_ptr<StorageImplementation<TL>> lhs,
            boost::intrusive_ptr<StorageImplementation<TR>> rhs,
            ...);
```

The dispatch wrapper owns dtype selection and output-buffer preparation. The backend kernel owns arithmetic only.

This is an internal design change. The public `Storage`, `Tensor`, and Python APIs should remain essentially unchanged except for bug fixes and in some cases (such as integer division) aligning with the intended Python behavior.

Against the current open issue list, this refactor would directly close about four issues, likely close or substantially resolve about five more, and provide the necessary infrastructure for roughly another dozen. That makes it unusually high-leverage technical-debt work: the same `Storage_base + void* + dtype int` pattern is behind several apparently separate bugs.

[...]

PR #997, opened 2026-07-06
refactor(pybind)!: guard UniTensor +/- on matching metadata; drop * / (#934)

yingjerkao:
## Decision

Maintainer ruling (yingjerkao, 2026-07-06, **amended**) resolving #934, #753, and #675: elementwise `UniTensor ⊙ UniTensor` arithmetic splits by whether the operation is well defined.

- **`+` `-` `+=` `-=` are kept, but guarded.** Addition/subtraction of two UniTensors is a meaningful vector-space operation *when the operands describe the same tensor slot*. The Python dunders now require matching metadata and preserve the shared labels; on a mismatch they raise a guidance `TypeError` instead of silently discarding labels.
- **`*` `/` `*=` `/=` are removed.** The elementwise (Hadamard) product/quotient of two UniTensors is basis-dependent and has no tensor-network meaning. Their dunders raise `TypeError`.

(This amends the earlier ruling, which removed all eight operators from the Python surface.)

## What changed

- The four **add/sub** dunders check that the two operands match on `uten_type`, rank, labels (in order), rowrank, `is_diag`, and bonds (`unitensor_addsub_metadata_mismatch`). On a match they perform the op and **restore the shared labels/name** — `linalg::Add`/`Sub` reset labels to a plain range and clear the name, which is exactly the #934 label-discarding complaint. On a mismatch they raise a `TypeError` naming the specific difference and suggesting `permute()`/`relabel_()` to align.
- The four **mul/div** dunders raise `TypeError` with guidance pointing at `Contract()`/`Kron()`/`ut.get_block()`.
- **Scalar ↔ UniTensor arithmetic** (both directions, out-of-place and in-place) is fully retained and dtype-preserving.
- **C++ `operator+`/`operator-` are untouched** — the Krylov solvers (`Lanczos_Gnd_Ut.cpp`, `Lanczos_Exp.cpp`) depend on them as vector-space (axpy) operations, and they carry the canonical internal/advanced doxygen note added earlier in this PR.

## Semantics of "matching metadata"

Order-sensitive: same `uten_type`, rank, labels (same order), rowrank, `is_diag`, and bonds (via `Bond::operator==`). `a["i","j"] + a["j","i"]` raises rather than auto-permuting — align the operands first. This keeps the sum/difference unambiguous and label-preserving. (For symmetric `BlockUniTensor`, the C++ `Add_` already enforces matching bonds/`is_diag`; the Python guard adds the label check and a uniform `TypeError`.)

## BREAKING CHANGE (Python users)

- `ut1 * ut2` / `ut1 / ut2` (and `*=` / `/=`) now raise `TypeError`.
- `ut1 + ut2` / `ut1 - ut2` (and `+=` / `-=`) now raise `TypeError` **when the operands' metadata do not match** (previously they silently discarded labels). With matching metadata they work and preserve labels.

The error messages carry the full migration guidance.

## Testing

`pytests/binding_unitensor_elementwise_test.py` (20 tests): matching-metadata `+`/`-`/`+=`/`-=` work and preserve labels; mismatched labels/bonds raise `TypeError`; `*`/`/`/`*=`/`/=` raise; scalar paths retained. The non-example pytest suite (68 passed, 1 skipped) and the DMRG dense / TDVP1 / ED Ising examples (4 passed, exercising the surviving C++ solver arithmetic end-to-end) confirm no regression.

## Stacking

Base PR #991 is now merged; this branch is rebased onto `master` and merges cleanly.

### DIFF
--- a/include/UniTensor.hpp
+++ b/include/UniTensor.hpp
@@ -349,6 +349,11 @@ namespace cytnx {
     virtual std::vector<Symmetry> syms() const;
 
     // arithmetic
+    // internal/advanced: the UniTensor-vs-UniTensor overloads below (Add_/Mul_/Sub_/Div_
+    // taking a boost::intrusive_ptr<UniTensor_base>) are not TN operations; python surface
+    // removed per #934. See operator+(const UniTensor&, const UniTensor&) in
+    // include/linalg.hpp for the full rationale. The Scalar overloads (scalar scaling)
+    // remain fully public, including from python.
     virtual void Add_(const boost::intrusive_ptr<UniTensor_base> &rhs);
     virtual void Add_(const Scalar &rhs);
 
@@ -4778,6 +4783,9 @@ namespace cytnx {
         @pre
         The two UniTensor need to have same structure.
         @note Compared to Add(const UniTensor&)const, this is an inplace function.
+        @note internal/advanced: not a TN operation; python surface removed per #934; see
+        operator+(const UniTensor&, const UniTensor&) in include/linalg.hpp for the
+        full rationale.
         @see Add_(const Scalar&), Add(const UniTensor&)const, Add(const Scalar&)const ,
         operator+=(const UniTensor&), operator+=(const Scalar&), \ref operator+
         */
@@ -4799,6 +4807,9 @@ namespace cytnx {
         @pre
         The two UniTensor need to have same structure.
         @note Compared to Mul(const UniTensor&)const, this is an inplace function.
+        @note internal/advanced: not a TN operation; python surface removed per #934; see
+        operator+(const UniTensor&, const UniTensor&) in include/linalg.hpp for the
+        full rationale.
         @see Mul_(const Scalar&), Mul(const UniTensor&)const, Mul(const Scalar&)const ,
         operator*=(const UniTensor&), operator*=(const Scalar&), \ref operator*
         */
@@ -4820,6 +4831,9 @@ namespace cytnx {
         @pre
         The two UniTensor need to have same structure.
         @note Compared to Sub(const UniTensor&)const, this is an inplace function.
+        @note internal/advanced: not a TN operation; python surface removed per #934; see
+        operator+(const UniTensor&, const UniTensor&) in include/linalg.hpp for the
+        full rationale.
         @see Sub_(const Scalar&), Sub(const UniTensor&)const, Sub(const Scalar&)const ,
         operator-=(const UniTensor&), operator-=(const Scalar&), \ref operator-
         */
@@ -4841,6 +4855,9 @@ namespace cytnx {
         @pre
         The two UniTensor need to have same structure.
         @note Compared to Div(const UniTensor&)const, this is an inplace function.
+        @note internal/advanced: not a TN operation; python surface removed per #934; see
+        operator+(const UniTensor&, const UniTensor&) in include/linalg.hpp for the
+        full rationale.
         @see Div_(const Scalar&), Div(const UniTensor&)const, Div(const Scalar&)const ,
         operator/=(const UniTensor&), operator/=(const Scalar&), \ref operator/
         */
@@ -4922,6 +4939,9 @@ namespace cytnx {
         @pre
         The two UniTensor need to have same structure.
         @note Compared to Add_(const UniTensor&), this function will create a new UniTensor.
+        @note internal/advanced: not a TN operation; python surface removed per #934; see
+        operator+(const UniTensor&, const UniTensor&) in include/linalg.hpp for the
+        full rationale.
         @see Add_(const UniTensor&), Add_(const Scalar&), Add(const Scalar&)const ,
         operator+=(const UniTensor&), operator+=(const Scalar&), \ref operator+
         */
@@ -4952,6 +4972,9 @@ namespace cytnx {
         @pre
         The two UniTensor need to have same structure.
         @note Compared to Mul_(const UniTensor&), this function will create a new UniTensor.
+        @note internal/advanced: not a TN operation; python surface removed per #934; see
+        operator+(const UniTensor&, const UniTensor&) in include/linalg.hpp for the
+        full rationale.
         @see Mul_(const UniTensor&), Mul_(const Scalar&), Mul(const Scalar&)const ,
         operator*=(const UniTensor&), operator*=(const Scalar&), \ref operator*
         */
@@ -4982,6 +5005,9 @@ namespace cytnx {
         @pre
         The two UniTensor need to have same structure.
         @note Compared to Div_(const UniTensor&), this function will create a new UniTensor.
+        @note internal/advanced: not a TN operation; python surface removed per #934; see
+        operator+(const UniTensor&, const UniTensor&) in include/linalg.hpp for the
+        full rationale.
         @see Div_(const UniTensor&), Div_(const Scalar&), Div(const Scalar&)const ,
         operator/=(const UniTensor&), operator/=(const Scalar&), \ref operator/
         */
@@ -5012,6 +5038,9 @@ namespace cytnx {
         @pre
         The two UniTensor need to have same structure.
         @note Compared to Sub_(const UniTensor&), this function will create a new UniTensor.
+        @note internal/advanced: not a TN operation; python surface removed per #934; see
+        operator+(const UniTensor&, const UniTensor&) in include/linalg.hpp for the
+        full rationale.
         @see Sub_(const UniTensor&), Sub_(const Scalar&), Sub(const Scalar&)const ,
         operator-=(const UniTensor&), operator-=(const Scalar&), \ref operator-
         */
--- a/include/linalg.hpp
+++ b/include/linalg.hpp
@@ -27,6 +27,20 @@ namespace cytnx {
    * @return [UniTensor] The result of the addition.
    * @pre \p Lt and \p Rt must have the same shape.
    * @see linalg::Add(const cytnx::UniTensor &Lt, const cytnx::UniTensor &Rt)
+   * @note internal/advanced: not a TN operation; python surface removed per #934
+   * (decision record 2026-07-06). This note is the canonical rationale for every
+   * UniTensor-vs-UniTensor elementwise arithmetic entry point: the free operators in
+   * this header ('+', '-', '*', '/') and the UniTensor::Add/Sub/Mul/Div member
+   * functions (in-place and out-of-place). The python UniTensor dunders raise
+   * TypeError for the UniTensor-vs-UniTensor case (see pybind/unitensor_py.cpp,
+   * raise_unitensor_elementwise_removed) while the C++ layer remains. operator+ and
+   * operator- are load-bearing: src/linalg/Lanczos_Gnd_Ut.cpp and
+   * src/linalg/Lanczos_Exp.cpp use them as genuine Krylov-subspace vector-space
+   * arithmetic (axpy) inside the solvers backing linalg::Lanczos(method="Gnd") and
+   * linalg::Lanczos_Exp(). operator* (elementwise/Hadamard product) and operator/
+   * (elementwise division) are basis-dependent, have no tensor-network meaning, and
+   * no in-tree C++ caller as of the audit; they are kept un-deprecated only because
+   * the removal task's scope was the python surface.
    */
   cytnx::UniTensor operator+(const cytnx::UniTensor &Lt, const cytnx::UniTensor &Rt);
 
@@ -63,6 +77,9 @@ namespace cytnx {
    * @return [UniTensor] The result of the subtraction.
    * @pre \p Lt and \p Rt must have the same shape.
    * @see linalg::Sub(const cytnx::UniTensor &Lt, const cytnx::UniTensor &Rt)
+   * @note internal/advanced: not a TN operation; python surface removed per #934;
+   * see operator+(const cytnx::UniTensor &Lt, const cytnx::UniTensor &Rt) for the
+   * full rationale.
    */
   cytnx::UniTensor operator-(const cytnx::UniTensor &Lt, const cytnx::UniTensor &Rt);
 
@@ -99,6 +116,9 @@ namespace cytnx {
    * @return [UniTensor] The result of the multiplication.
    * @pre \p Lt and \p Rt must have the same shape.
    * @see linalg::Mul(const cytnx::UniTensor &Lt, const cytnx::UniTensor &Rt)
+   * @note internal/advanced: not a TN operation; python surface removed per #934;
+   * see operator+(const cytnx::UniTensor &Lt, const cytnx::UniTensor &Rt) for the
+   * full rationale.
    */
   cytnx::UniTensor operator*(const cytnx::UniTensor &Lt, const cytnx::UniTensor &Rt);
 
@@ -135,6 +155,9 @@ namespace cytnx {
    * @return [UniTensor] The result of the division.
    * @pre \p Lt and \p Rt must have the same shape.
    * @see linalg::Div(const cytnx::UniTensor &Lt, const cytnx::UniTensor &Rt)
+   * @note internal/advanced: not a TN operation; python surface removed per #934;
+   * see operator+(const cytnx::UniTensor &Lt, const cytnx::UniTensor &Rt) for the
+   * full rationale.
    */
   cytnx::UniTensor operator/(const cytnx::UniTensor &Lt, const cytnx::UniTensor &Rt);
 
--- a/pybind/unitensor_py.cpp
+++ b/pybind/unitensor_py.cpp
@@ -95,24 +95,23 @@ inline bool parse_get_blocks_silent_arg(const py::args &args, const py::kwargs &
   return silent;
 }
 
-// Phase-2 Task 3 (issue #934, decision record 2026-07-06): elementwise
-// UniTensor(+)UniTensor arithmetic ('+', '-', '*', '/' and their in-place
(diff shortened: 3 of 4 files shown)


ianmccul:
(Unitensor + Unitensor) and (UniTensor - UniTensor)  are definitely meaningful and essential operations.  As are (UniTensor * Scalar) and (UniTensor / Scalar).

The non-meaningful operations are (UniTensor +- Scalar) and (UniTensor */ UniTensor)

ianmccul:
From #934:

The following should be removed or disabled:

```c++
UniTensor * UniTensor
UniTensor / UniTensor
UniTensor + Scalar
Scalar + UniTensor
UniTensor - Scalar
Scalar - UniTensor
Scalar / UniTensor
UniTensor::Pow
UniTensor::Inv  // if element-wise
UniTensor % ...
```

For comparison, these operations ***do*** make sense on `UniTensor`, at least conceptually:

- `Add_(const UniTensor_base&)` and `Sub_(const UniTensor_base&)`, if the tensor spaces/bonds match exactly.
- `Mul_(const Scalar&)`, scalar multiplication.
- `Div_(const Scalar&)`, scalar division.
- `contract`, if leg ordering, symmetry, and fermionic/braiding semantics are handled correctly.
- `Trace`, if the traced legs are compatible.
- `Conj`, `Dagger`, `Norm`, `normalize`.
- `permute`, provided it handles fermionic signs/braiding correctly.
- `astype`, `clone`, `to(device)`, and metadata accessors.

I don't have any good suggestion for resolving #753 and #675.  Certainly if both tensors have the same labels and in the same order and the bonds match, then preserving them is a good idea.  If they have the same labels but in a different order, it should probably be an error; the alternative is some implicit permutation/braiding which should be a deliberate operation.  I have no suggestion of what to do if some labels match but others do not.

The typical operation will be something like `y += a * matvec(x)`, so its the matvec contraction that needs to get the labels correct for that operation.  Linear algebra operations "don't care", just want to get operations such as `y += a * matvec(x)` and inner products etc to work, do not need to know what `x` and `y` actually are or anything about labels.

pcchen:
## Code Review

Implements the maintainer ruling to remove elementwise `UniTensor (+) UniTensor` arithmetic from the Python surface (dunders now raise a guidance `TypeError` via a `[[noreturn]]` helper), while deliberately keeping the C++ operators the Krylov solvers depend on. Clean, focused, and unusually well-documented.

### What's done well
- **Correct mechanism.** The UniTensor-RHS overload is registered first and matches `ut1+ut2` exactly, so its body runs and throws `py::type_error` -> propagates as a Python `TypeError` (pybind does not fall back to `__radd__`/NotImplemented when a *selected* overload throws). Scalar/numpy overloads are untouched, so `ut + 2.0` etc. still work. ✅
- **`[[noreturn]]` + declared return types** (`-> UniTensor` / `-> UniTensor &`) compile cleanly — verified by green wheel builds on macOS x2 + Ubuntu x2. `std::format` is fine (`<format>` included; already used elsewhere in the file). ✅
- **C++ layer correctly preserved** — the audit-backed reasoning (Lanczos_Gnd_Ut / Lanczos_Exp use `+`/`-` as axpy) is sound, and the doxygen cross-references are thorough and honest (including admitting `*`/`/` have no in-tree caller). ✅
- **Excellent tests + guidance messages** — 16 tests (8 removal with ordered-token pins on `934.*Contract/Kron/get_block`, 8 scalar-survival), and the error strings give real migration paths. ✅

### Findings

**1. Incomplete removal — `//` and `//=` still perform elementwise division (should-fix; it defeats the stated goal).**
`__floordiv__(UniTensor, UniTensor)` (unitensor_py.cpp:1612) and `__ifloordiv__(UniTensor, UniTensor)` (:1708) were overlooked — they still call `linalg::Div(self, rhs)` / `self.Div_(rhs)`. So after this PR:
```python
ut1 / ut2    # TypeError (removed) ✅
ut1 // ut2   # silently returns linalg::Div(ut1, ut2) — the removed op, via a different dunder ❌
ut1 //= ut2  # silently self.Div_(rhs) ❌
```
This leaves a live bypass for exactly the label-discarding/basis-dependent operation #934 targets, and makes the surface inconsistent (`/` gone, `//` kept). The description enumerates "8 dunders (`+ - * / += -= *= /=`)" — it should be **10**, adding `//` and `//=` to `raise_unitensor_elementwise_removed` (reuse `kUniTensorDivRemovedGuidance`) plus two tests. Bonus: this also removes a pre-existing oddity where UniTensor `//` was aliased to *true* division rather than floor division.

**2. CI gap — the test suite hasn't demonstrably run on this commit.** The visible checks show wheels/formatting/docs green but **no `BuildAndTest`/pytest job** (the branch is behind master; the run looks partial/stale). Compilation is verified, but the 16 new tests' only evidence is the local "71 passed." Recommend a fresh full CI run before merge (as with #989/#991).

**3. Breaking change — needs a changelog/release note.** Well-flagged in the title (`!`), body, and error messages, but a user-facing removal like this should land a CHANGELOG entry + a minor/major version signal so downstream `ut1 + ut2` breakage is discoverable.

**4. Minor (acknowledged): `*`/`/` C++ operators kept un-deprecated** despite no in-tree caller — honestly documented as out-of-scope, flagged for follow-up. Fine.

### Verdict
Strong, well-documented refactor with correct mechanics and great tests — but the removal is **incomplete**: `//`/`//=` slip through and still perform the very operation being removed. I'd fix that (finding #1) and re-run full CI before merging. **Requesting changes on the one real gap.**

pcchen:
@yingjerkao Please complete the removal.

ianmccul:
I am lost for words.  A tensor network library that doesn't allow addition of tensors?

pcchen:
## Technical summary: which UniTensor arithmetic is well-defined?

Following @ianmccul's points, here is an operation-by-operation summary to help settle the direction, since the disagreement is about the *axis* of the removal rather than a detail.

A `UniTensor` is an element of a labeled tensor-product vector space with (block-sparse) symmetry structure. That fixes which operations are mathematically sound:

| Operation | Well-defined? | Why |
|---|---|---|
| `T ± T` (same bonds/labels) | **Yes — essential** | Tensors form a vector space; adds corresponding blocks; the sum of two symmetric tensors is symmetric. This is `y = a·x + b·z` — the backbone of Krylov/DMRG/TDVP. |
| `scalar · T`, `T · scalar`, `T / scalar` | **Yes** | Scaling every block preserves the zero/symmetry pattern. |
| `T ± scalar`, `scalar ± T` | **No** | Adds a constant to *every* element, including entries symmetry forces to zero → destroys block-sparsity/symmetry. Not a canonical TN op. |
| `scalar / T` | **No** | Elementwise reciprocal; basis-dependent, inf/nan. |
| `T * T`, `T / T` (Hadamard) | **No** | Elementwise product/quotient; basis-dependent, no TN meaning, breaks block structure. |

This partition matches **#934's own list** (remove `T*T`, `T/T`, `T±scalar`, `scalar±T`, `scalar/T`, `Pow/Inv/%`; keep `Add_/Sub_(UniTensor)`, `Mul_/Div_(Scalar)`).

### What this PR does vs. that partition

| | Correct action | This PR |
|---|---|---|
| `T ± T` | **keep** | ❌ removes (raises TypeError) |
| `T * T`, `T / T` | remove | ✅ removes |
| `T // T`, `T //= T` | remove | ❌ still performs it (see my earlier review) |
| `T ± scalar`, `scalar ± T` | **remove** | ❌ keeps |
| `scalar / T` | **remove** | ❌ keeps |
| `scalar·T`, `T·scalar`, `T/scalar` | keep | ✅ keeps |

**The PR is inverted on the additive axis:** it removes the meaningful `T ± T` and retains the ill-defined `T ± scalar` / `scalar / T`.

### The decisive internal contradiction

This PR **keeps the C++ `operator+`/`operator-`** on UniTensor, explicitly justified in its own doxygen note because the Lanczos/TDVP solvers rely on them as *"genuine vector-space arithmetic (axpy)."* If `T + T` is a legitimate vector-space operation in C++, it is equally legitimate from Python — removing it only from the Python surface is inconsistent.

### On #753/#675 (the label footgun)

The real problem those issues raise is that `T1 + T2` with **mismatched labels** silently misaligns/discards labels. The remedy is to **guard** `±` (require matching labels; error on same-labels-different-order rather than implicitly permuting), **not remove** `±` — that preserves the essential operation while closing the footgun.

### Bottom line

#934's text supports keeping `T ± T`. Suggest resolving direction before further work here: at minimum **keep `T ± T`** (guarded on labels) and move `T ± scalar` / `scalar ± T` / `scalar / T` into the removed set. The `T*T`, `T/T` (and the missed `//`, `//=`) removals stand either way.

I'll revise my review verdict accordingly: this needs a direction decision between the maintainers, not just the `//` completion.

pcchen:
Here is what I thought now.

T1 + T2 is allowed only when **bond structure** and **bond labels** and **bond ordering** are exactly the same. For example, if T1 and T2 are of the same shape, but with different label, T1 + T2 will raise an error. If user is sure that T1 + T2 make sense, then user has to explcitly relabel/permute T1 and T2 so that all bond metadata maches.

yingjerkao:
Yes, we should keep tensor addition and subtraction when the bonds/quantum numbers match.

pcchen:
### `//` and `//=` (UniTensor ÷ UniTensor) bypass the `/` removal

While `/` and `/=` on two UniTensors now correctly raise `TypeError`, the floor-division operators are still wired to `Div`:

```cpp
.def("__floordiv__",  [](UniTensor &self, const UniTensor &rhs) { return linalg::Div(self, rhs); })
.def("__ifloordiv__", [](UniTensor &self, const UniTensor &rhs) { return self.Div_(rhs); })
```

So after this PR:

```python
ut1 / ut2     # TypeError — Hadamard quotient removed ✅
ut1 // ut2    # silently returns linalg::Div(ut1, ut2) — the *same* removed op ❌
ut1 //= ut2   # silently self.Div_(rhs) ❌
```

Two problems, and they compound:

1. **It's a bypass of the removal.** `//`/`//=` perform exactly the element-wise Hadamard quotient that `/`/`/=` were removed for — the operation is still fully reachable, just under a different operator.
2. **It's also mislabeled.** These are Python's *floor*-division operators, but they call `Div` (true element-wise division) — no flooring. So on a UniTensor `//` just aliases `/`, which is surprising in its own right.

Suggested fix (consistent with the `*`/`/` handling): route the UniTensor-RHS `__floordiv__`/`__ifloordiv__` through `raise_unitensor_elementwise_removed(" // ", …)` / `(" //= ", …)` using the div-guidance string, and add two tests (`ut1 // ut2` and `ut1 //= ut2` raise), mirroring the existing `__truediv__`/`__itruediv__` cases.

pcchen:
### Ruling needed: scalar ± UniTensor and scalar / UniTensor

The tensor–tensor axis is now settled (keep guarded `T ± T`, remove `T * T` / `T / T`). One set of operations remains where the amended PR and #934's original list disagree, and it needs an explicit maintainer ruling:

| Operation | #934 list (@ianmccul) | Current PR (`340c8e8b`) |
|---|---|---|
| `UniTensor + Scalar` | remove | **kept** |
| `Scalar + UniTensor` | remove | **kept** |
| `UniTensor - Scalar` | remove | **kept** |
| `Scalar - UniTensor` | remove | **kept** |
| `Scalar / UniTensor` | remove | **kept** |

(Note: `Scalar * UniTensor`, `UniTensor * Scalar`, `UniTensor / Scalar` — scalar scaling — are uncontroversial and kept by everyone.)

**Technical considerations, for the decision:**

- **`± Scalar`** adds a constant to every *stored* block element. For a block/symmetric UniTensor this does **not** fill symmetry-forbidden entries (those aren't stored), so it doesn't corrupt block structure — but it's still not a symmetry-covariant operation, and `T + c` is **not** the same as `T + c·I`. So it "works" mechanically but has questionable tensor-network meaning.
- **`Scalar / UniTensor`** is an element-wise reciprocal of the stored blocks: basis-dependent, and produces `inf`/`nan` wherever a stored element is 0.

**Two coherent options:**

1. **Remove them** (follow #934's list) — treat `± Scalar` and `Scalar /` as ill-defined, alongside the `* / UniTensor` removals; keep only scalar *scaling* (`*` and `/ Scalar`).
2. **Keep them** (amend #934) — accept the "operate on stored blocks" semantics as good enough, and update the #934 record accordingly.

@yingjerkao — the current PR keeps all five; is that a deliberate scope decision or just out of scope for this pass?

ianmccul:
I thought UniTensor + Scalar throws an exception for Block and Fermionic cases?

yingjerkao:
Two semantic choices in the metadata guard worth a reviewer's eye — both easy to change if you'd prefer different behavior:

**1. Label matching is order-sensitive (strict), no auto-permute.**
`unitensor_addsub_metadata_mismatch` requires `a.labels() == b.labels()` in the *same order*, so `a.relabel(["i","j"]) + a.relabel(["j","i"])` raises `TypeError` rather than transposing `b` to align with `a` before adding. I went strict because it's unambiguous and keeps the guard cheap; the alternative (permute the RHS to the LHS's label order when the label *sets* match) is more convenient but adds a hidden transpose + a copy. Easy to relax later if you want numpy-broadcast-style convenience.

**2. `rowrank` and `is_diag` are part of the match.**
Two tensors with identical labels/bonds but different `rowrank` (row/col leg split) or different `is_diag` will raise. Rationale: they describe genuinely different tensor slots (rowrank changes contraction semantics; a diagonal vs dense block differ in storage/meaning). This is stricter than the C++ `Add_`, which for `DenseUniTensor` keys off bonds/shape and would happily add across a rowrank difference. If you'd rather have `+`/`-` ignore `rowrank` (treat it as pure bookkeeping for untagged Dense tensors), I can drop it from the check.

For reference, the full match criteria are: same `uten_type`, rank, labels (in order), `rowrank`, `is_diag`, and bonds (via `Bond::operator==`). On a match the op runs and the shared labels/name are restored (since `linalg::Add`/`Sub` reset them); on a mismatch the `TypeError` names the specific difference.

yingjerkao:
We should allow UniTensor/(Rank-0 UniTensor/Tensor)

pcchen:
## Resolution of the open review items — this PR is now approvable as-is

Maintainer direction (@pcchen) resolving the two items left open in review:

**1. The scalar-operand ruling** (my "Ruling needed" table above — `UniTensor ± Scalar`, `Scalar ± UniTensor`, `Scalar / UniTensor`): resolved by the **rank-0 path** rather than in this PR.
- **#1026** (@ianmccul) implements dense rank-0 tensors, resolving the #1016 design question.
- **After #1026 merges, a follow-up issue will be filed to implement `UniTensor / (rank-0 UniTensor/Tensor)`** — the principled, tensor-typed scalar division (dtype/symmetry-aware, the `normalize_`-style broadcast) that supersedes the ad-hoc Python-scalar-division semantics debated here. The Python-scalar operands this PR *keeps* remain as-is until that lands; no change required in this PR.

**2. The `//`/`//=` bypass finding**: superseded by direction — integer dtypes are slated for removal from UniTensor, so floor division has no future operand set and `//` will be **unbound entirely** (as #1015 already implements per #941) rather than given raise-with-guidance handling here. No change required in this PR.

With both items resolved by direction, what this PR delivers stands on its own and matches the agreed rulings: **guarded `T ± T`** (matching uten_type/rank/labels/rowrank/is_diag/bonds, label-and-name-preserving on match), **`T * T` / `T / T` removed** with guidance errors, scalar arithmetic untouched. Review verdict updates to **approve**.

pcchen:
## Correction to point 1 above (maintainer ruling, @pcchen)

**`UniTensor ± Scalar` and `Scalar ± UniTensor` will NOT be supported** — the ruling adopts #934's original list for the additive scalar operands (adding a constant to every stored element is not a tensor-network operation), not the keep-as-is reading my previous comment implied.

Updated resolution map for the scalar operands:

| Operation | Ruling |
|---|---|
| `T ± scalar`, `scalar ± T` | **Removed** (this ruling; no replacement needed — the operation is ill-defined on a TN object) |
| `T / scalar` (and `T * scalar`, `scalar * T`) | Kept — scalar *scaling* is well-defined; the principled tensor-typed form (`T / rank-0`) arrives via the #1026 follow-up issue |
| `scalar / T` | Remains in the remove set per #934; the #1026 follow-up issue will settle whether any rank-0 reciprocal form exists |

The `± scalar` removal is follow-up implementation work (it can ride the same follow-up as the rank-0 division, or land as its own small PR mirroring this PR's guidance-error pattern) — per maintainer direction it is **not a blocker for merging this PR**, which already delivers the agreed tensor-tensor semantics. Verdict unchanged: approve.

pcchen (approved):
Approving. The amended direction is correct and verified: `T ± T` is kept and **guarded on matching metadata** (uten_type, rank, labels in order, rowrank, is_diag, per-leg bonds), performing with the shared labels and name restored on match — which directly fixes the #753/#675 label-discarding complaint — and raising a specific, actionable `TypeError` on mismatch; `T * T` / `T / T` (Hadamard) are removed with guidance errors; scalar arithmetic and the C++ solver-facing `operator+/-` are untouched. Tests pin all of it (matching-metadata perform ×4, mismatch raises, removal raises, scalar survival ×8) and CI is fully green (16 pass).

The remaining semantic questions are resolved by documented maintainer rulings rather than in this PR: `UniTensor ± Scalar` / `Scalar ± UniTensor` are ruled **not supported** (removal is follow-up work mirroring this PR's pattern); tensor-typed scalar division arrives via the #1026 rank-0 follow-up; and `//`/`//=` will be unbound entirely per the integer-removal direction (#1015 implements this) — none block this merge.

Issue #841, opened 2026-05-27
Remove public mutable accessor UniTensor::signflip_(); enforce _signflip invariants inside BlockFermionicUniTensor


IvanaGyro:
## Problem

`UniTensor::signflip_()` ([`include/UniTensor.hpp:3204`](https://github.com/Cytnx-dev/Cytnx/blob/master/include/UniTensor.hpp#L3204)) exposes a non-const reference into the private `_signflip` member of `BlockFermionicUniTensor` through the refcounted `_impl` pointer:

```cpp
std::vector<bool> &signflip_() { return this->_impl->signflip_(); }
```

This is a public-API encapsulation hole.

### Why it's dangerous (beyond style)

1. **Invariant break.** `_signflip.size()` must equal `_blocks.size()` and stay in lockstep with `_inner_to_outer_idx` and the qnums on the contracted bond. The public `signflip_()` lets any caller mutate `_signflip` independently. `ut.signflip_().push_back(true)` would silently desynchronize the three vectors and the next fermionic operation would produce wrong signs or UB.
2. **UB hazard in the base.** `UniTensor_base::signflip_()` at `src/UniTensor_base.cpp:71-78` errors and then `return t;` where `t` is a stack-local `std::vector<bool>` &mdash; returning a reference to it is UB if the error message is ever softened to a non-fatal warning.
3. **Dangling-reference hazard.** The reference is into `_impl`'s heap object. If the holding `UniTensor` is reassigned while the reference is live, it dangles silently (refcount drops to zero on the previous `_impl`).
4. **`_signflip` is currently `public`** in `BlockFermionicUniTensor` (`include/UniTensor.hpp:1838-1845`), so `src/linalg/Svd.cpp:587,607` and `src/linalg/Gesvd.cpp:584,604` reach in and assign to it directly through raw `BlockFermionicUniTensor*`. Second layering hole, independent of the public `signflip_()`.

### Google C++ Style

- "Make data members `private`" (Access Control).
- "Output parameters should usually be values or pointers" / "prefer to return a value rather than a pointer" (Inputs and Outputs).
- General guidance against returning non-const references to internal state.

## Why a public `EraseBlocks` / `EraseSignflipEntries` helper is also wrong

[...]

PR #1001, opened 2026-07-06
refactor(Bond)!: make Bond immutable; encapsulate BlockFermionicUniTensor signflip

### LINKED ISSUE #846: Make Bond immutable: in-place bond mutation breaks the derived block-tensor symmetry invariant
## Motivation

A `Bond` is part of the data that *defines* a symmetric tensor's block structure: for a `BlockUniTensor` / `BlockFermionicUniTensor`, the set of valid blocks (`_blocks`, `_inner_to_outer_idx`, `_degs`) is derived from the bonds' quantum numbers and degeneracies combined through the `Symmetry` combine rule (see #841, which states the block set is "conceptually derived from the Symmetry combination rule applied to bond qnums").

Because `Bond` today exposes in-place mutators **and** copies share the impl (`boost::intrusive_ptr<Bond_impl>`), it is possible to mutate a bond *after* it has been bound into a block tensor:

```cpp
ut.bonds()[0].qnums().push_back(...);  // or .redirect_(), .set_type(...), .combineBond_(...)
```

This silently desynchronizes the bond from the block structure that was derived from it: the qnums/degeneracies/direction no longer match `_blocks` / `_inner_to_outer_idx`, so the next symmetric operation produces **wrong symmetry sectors, wrong results, or UB**. The same edit can also leak through aliasing into any other tensor or holder that shares the bond's impl.

The fix decided in #843 is to make the **`Bond` public interface immutable** while keeping the shared PIMPL (so copies stay O(1)). Immutability removes this class of bug at the source: a bond bound to a tensor can no longer be changed out from under the derived block structure, sharing becomes safe, copy-on-write is unnecessary, and `clone()` collapses to identity. This issue tracks that implementation.

References: decision in #843 (https://github.com/Cytnx-dev/Cytnx/issues/843#issuecomment-4561220581); related encapsulation/invariant discussion in #841; ABI coordination with #842.

### Remove from the public `Bond` interface
- In-place mutators: `set_type` (`include/Bond.hpp:430`), `redirect_` (`:464`), `combineBond_` (`:522`), `clear_type` (`:474`).
- Non-const reference accessors: `qnums()` (`:379`), `syms()` (`:413`), `getDegeneracies()` (`:764`).

### Keep
- All const accessors: `qnums() const` (`:375`), `syms() const` (`:409`), `type()` (`:367`), `dim()` (`:395`), `Nsym()`, `qnums_clone()`, `syms_clone()`, etc.
- The return-new variants that already exist and should become the only way to "change" a bond: `retype` (`:444`), `redirect` (`:454`), `combineBond` (non-`_`, the variant that returns a new `Bond`).
- `clone()` — now an identity copy (`return *this`); the shared impl is safe to alias once immutable.
- `.cybd` Save/Load wire format, byte-for-byte (`_dim`, `_type`, `_degs`, `_qnums`, `_syms` in the same order).

### Migrate these in-place call sites to build a new `Bond` and assign it into the tensor's `_bonds` (instead of mutating a shared impl)
- `redirect_` on a tensor's own bonds: `src/BlockUniTensor.cpp:1216`, `src/BlockFermionicUniTensor.cpp:1814`, `src/DenseUniTensor.cpp:1256`.
- `set_type` on a tensor's own bonds: `src/SparseUniTensor.cpp:1689`.
- `combineBond_` on stored bonds: `src/DenseUniTensor.cpp:877,915`, `src/BlockUniTensor.cpp:1923`, `src/BlockFermionicUniTensor.cpp:2672` (and the `combineBonds` orchestration around them).
- Output-bond `set_type` in the linalg decompositions (these mutate freshly-built output bonds, lower risk, but should move to the immutable pattern for consistency): `src/linalg/Svd.cpp`, `Gesvd.cpp`, `Qr.cpp`, `Qdr.cpp`, `Rsvd.cpp`, `Svd_truncate.cpp`, `Gesvd_truncate.cpp`, `Rsvd_truncate.cpp`.

Note: making `Bond` immutable does **not** forbid a `UniTensor` from changing *which* bonds it holds (e.g. `combineBonds`, `permute`, decomposition outputs). Those operations remain legal — they just construct new `Bond` objects and replace entries in `_bonds`, and (for block tensors) rebuild the derived block structure consistently, rather than mutating an existing `Bond`'s impl in place.

## Suggested work

- [ ] Delete the in-place mutators and non-const reference accessors listed above from `Bond` / `Bond_impl`'s public surface.
- [ ] Migrate every call site above to `retype` / `redirect` / `combineBond` + assignment into `_bonds`.
- [ ] Confirm no migrated site relied on the mutation propagating through a shared impl to a separate holder (i.e. the old aliasing behavior was never load-bearing).
- [ ] Make `clone()` an identity copy.
- [ ] Verify `.cybd` Save/Load round-trips byte-for-byte.
- [ ] Audit pybind (`pybind/bond_py.cpp`) for any binding of the removed mutators; drop or redirect them to the return-new variants.
- [ ] ABI-breaking change — coordinate with the version bump in #842.
- [ ] Run the symmetry / Block / BlockFermionic tests; add a regression test that a `Bond` obtained from a tensor cannot desynchronize the tensor's block structure.

yingjerkao:
## Problem

Fixes #846 and #841 per the consensus reached in both threads (full public immutability; "correctness guarded by design, not only by tests"). Bond's in-place mutators let user code silently break the block-tensor symmetry invariant of any UniTensor holding that bond, and the public mutable `signflip_()` accessor could desynchronize `BlockFermionicUniTensor`'s `_signflip`/`_blocks` lockstep.

## What changed

- **Bond public surface is now immutable**: `set_type`, `redirect_`, `combineBond_`, `clear_type`, `group_duplicates_`, `operator*=`, non-const `qnums()`/`syms()`, and the mutable `getDegeneracies()` are removed. Return-new forms (`retype`, `redirect`, `combineBond`, `group_duplicates`) remain and are pinned by new tests.
- **`UniTensor::bonds()` is const-only**; internals migrated.
- **`_signflip` is private** (#841). The issue's enumeration turned out incomplete — the direct-access pattern existed across 8 linalg files, and the per-helper friend design was infeasible (file-static helpers can't be friended) — so a single documented gateway `linalg::_fermionic_signflip_()` is friended once, with the invariant story in its header comment.
- **Real bug found and fixed during migration**: the combine-bonds helpers copied a bond handle that *shared the impl* with the tensor's own bond, then mutated it in place — silently rewriting any user-held bond copy. Now clones first; regression tests pin that a user-held `Bond` survives a `force=true` combine untouched.
- Python: the removed mutators now raise `AttributeError` (their replacements are same-named siblings — `redirect`, `retype`, `combineBond`); the misleading `group_duplicates_` binding (which never behaved as documented) is also removed.

## BREAKING CHANGE

C++ and Python code calling the removed Bond mutators must switch to the return-new forms. Release notes should list: `set_type`→`retype`, `redirect_`→`redirect`, `combineBond_`→`combineBond` (out-of-place), `clear_type` (construct anew), `group_duplicates_`→`group_duplicates`.

## Testing

C++: 1071 passed / 11 skipped / only the 4 pre-existing failures addressed by #977 (8 new tests incl. a truncation test pinning `signflip().size()==Nblocks()` through real Svd/Gesvd truncation, and the aliasing regression pair). pytest: 47 passed. The two known pre-existing failures in `tests/test_apply_pybind.py` are fixed here (d4c07dc2): the `_are_nearly_eq` helper compared `get_blocks_()` pairwise in storage order, which is not canonical, so it zipped mismatched sectors; it now matches blocks by quantum-number sector via `get_qindices()`, mirroring the C++ `AreEqUniTensor`. Unlike the C++ helper it *compares* per-block `signflip()` flags instead of compensating them — compensation would let a no-op `apply_()` still compare equal, making the in-place test vacuous. The file also moved to `pytests/apply_test.py` so CI's `pytest pytests/` actually runs it (it was silently skipped under `tests/`).

## Stacking

**Base is `fix/pybind-inplace-return-self` (PR #986)** — this branch removes pybind mutator bindings that #986 rewrote. Merge #986 first, then retarget to master.


include/UniTensor.hpp
```diff
@@ -1818,6 +1816,17 @@ namespace cytnx {
...
   //======================================================================
+  class BlockFermionicUniTensor;
+
+  namespace linalg {
+    // Internal accessor granting the fermionic sign-flip bookkeeping to the SVD/QR/Eig/Exp
+    // decomposition internals that legitimately construct or truncate BlockFermionicUniTensor
+    // blocks (see #841). Not part of the public API: _signflip must stay in lockstep with
+    // _blocks/_inner_to_outer_idx and the qnums on the contracted bond, so no other code may
+    // touch it.
+    std::vector<cytnx_bool> &_fermionic_signflip_(BlockFermionicUniTensor &self);
```
IvanaGyro: It is discouraged to use variable name prefixed with `_`. BTW, will we set the style rules in AGENT.md or CLAUDE.md for bots?

include/UniTensor.hpp
```diff
@@ -1827,10 +1836,14 @@ namespace cytnx {
       _inner_to_outer_idx;  // stores the qindices for each block
     std::vector<Tensor> _blocks;
     Tensor NullRefTensor;  // this returns when accessed block does not exists!
-    // additional informaiton for fermions:
+
+   private:
+    // additional information for fermions:
     std::vector<cytnx_bool>
       _signflip;  // if true, the sign of the corresponding block needs to be flipped
+    friend std::vector<cytnx_bool> &linalg::_fermionic_signflip_(BlockFermionicUniTensor &self);
```
IvanaGyro: If others may call this function directly, this should be public not private.

IvanaGyro (changes requested):

pcchen:
## Code Review

Removes Bond's public in-place mutators (Bond is now immutable, so user code can't silently break a UniTensor's symmetry invariant), makes `BlockFermionicUniTensor::_signflip` private behind a single friended gateway, and fixes a real aliasing bug surfaced during the migration.

### Correctness — verified (high confidence)
- **Bond immutability is clean.** Public mutators (`set_type`, `redirect_`, `combineBond_`, `clear_type`, `group_duplicates_`, `operator*=`, non-const `qnums()/syms()`, mutable `getDegeneracies()`) are removed from the wrapper; the impl keeps them for internal use; return-new forms (`retype`/`redirect`/`combineBond`) call `out._impl->set_type(...)` on a deep `clone()`, so they're fully independent. ✅
- **The aliasing bug fix is real and correct.** The old `Bond tmp = this->_bonds[i]` copied a wrapper that *shared* the `intrusive_ptr` impl, then mutated it in place — silently rewriting any user-held bond copy. Now `Bond tmp = this->_bonds[i].clone()` isolates it; the `force` branch mutating `tmp._impl` is safe because `tmp` is now a clone. Regression tests pin it. ✅
- **`group_duplicates_()` -> `group_duplicates(mapper)` preserves behavior** — the mapper is still captured into `idx_mappers`, and `_bonds[i]` is replaced with the deduplicated bond. ✅
- **Signflip migration verified end-to-end** (dedicated pass over all 11 linalg files): the gateway is a transparent alias, read/write semantics are identical to the old `_signflip`/`signflip_()`, and **every truncation branch (Gesvd/Svd/Rsvd) erases blocks, qnums, and signflip with the same `to_be_removed` index set in the same order**, keeping `signflip().size() == Nblocks()` consistent. No un-migrated mutators, no staleness/ordering/clone differences. ✅
- **The Python `_are_nearly_eq` fix is a genuine correctness improvement** — it matches blocks by quantum-number sector (`get_qindices`) instead of non-canonical storage order (fixing the two pre-existing failures that zipped mismatched sectors) and *compares* signflips rather than compensating them (so a no-op `apply_()` can't vacuously pass). Moving it to `pytests/` so CI actually runs it (it was silently skipped under `tests/`) is a good catch. ✅
- **Compiles green on all wheel platforms.** ✅

### Findings (low / notes — none blocking correctness)

**1. The signflip gateway is a transparent mutable-ref alias, not an invariant-enforcing interface.**
`_fermionic_signflip_` just `return self._signflip;` (full `std::vector<cytnx_bool>&`). This *does* achieve #841's goal — external/user code can no longer touch `_signflip` (private + one internal friend), so the "public `signflip_()` desync" bug is genuinely eliminated. But the `_signflip`/`_blocks` lockstep is still maintained by hand inside the backend, not guaranteed by the gateway. That's fine and honestly documented; I'd just calibrate the "correctness guarded by design" framing — the *Bond* half is truly design-guarded, the *signflip* half is friend-restricted + discipline. **Follow-up idea:** narrow the gateway to the specific ops the backend needs (append-false, erase-by-index) so lockstep becomes hard to break, not just hard to reach.

**2. CI gap — the test evidence is local-only.** Wheels compile green, but no `BuildAndTest` (C++ gtest) or pytest job ran on this commit — so "1071 C++ / 47 pytest passed," the new `signflip().size()==Nblocks()` truncation test, and the aliasing regression tests are unverified in CI. Given this PR's entire thesis is correctness-by-invariant, a **fresh full-test CI run is important before merge** (same note as #997).

**3. Breaking change — needs a changelog/release note + version signal.** Well-flagged (title `!`, BREAKING section, and the removed->replacement migration table). That table is a good basis for the release notes.

### Verdict
High-quality, careful refactor: Bond immutability is cleanly executed, the signflip encapsulation meets the #841 goal, the surfaced aliasing bug is a genuine fix, and the whole linalg migration is behavior-preserving (independently verified, including the truncation lockstep). Findings are notes, not defects. **Approve-with-nits, pending a fresh full-test CI run** (the one real gate given the correctness-focused nature of the change).

pcchen:
## Review update — folding in @IvanaGyro's two comments on the signflip gateway

Both land on `linalg::_fermionic_signflip_`, and both are fair. Precise take on each:

### 1. Naming: leading/trailing underscores on the gateway (`include/UniTensor.hpp:1827`)
Agree this is worth changing. To be precise on the reserved-identifier rule: `_fermionic_signflip_` is declared inside `namespace cytnx::linalg` and starts with `_` + a **lowercase** letter, so it is *not* UB — a single leading underscore is reserved only at **global** scope, and there's no `__`/`_[A-Z]`. So it compiles legally. But the style objection stands, and it's actually doubly confusing here:
- cytnx uses a **leading `_`** for *member variables* (`_signflip`, `_blocks`, `_bonds`) — so a leading `_` on a free function reads like a member.
- cytnx uses a **trailing `_`** for *in-place mutator methods* (`redirect_`, `combineBond_`, `tag_`) — but this is an accessor gateway, not a mutator, so the trailing `_` is misapplied.

Suggest renaming to something that signals "internal accessor" without borrowing either convention, e.g. put it in a `detail`/`internal` namespace: `cytnx::linalg::internal::fermionic_signflip_ref(BlockFermionicUniTensor&)`.

On the meta question (style rules in `CLAUDE.md`/`AGENT.md`): +1, worth doing. Codifying the three conventions — leading `_` = member, trailing `_` = in-place mutator returning self, no leading `_` on free functions — would keep both humans and bots consistent; this PR is a concrete case where it would have helped.

### 2. `friend` in the `private:` section (`include/UniTensor.hpp:1844`) — "make it public not private"
One C++ clarification first: the access section a `friend` declaration sits in is **irrelevant** — friendship is granted regardless of whether it's written under `public:` or `private:`. And the gateway itself is *already* public: it's a namespace-scope function in `linalg` (line 1827), callable by anyone who includes the header. So nothing here is actually "private" except the `_signflip` **member**.

So the design is: `_signflip` is `private` (no code can touch it) + exactly **one** `friend` gateway is the only door. That's deliberate and it's what satisfies #841 — making `_signflip` (or a plain accessor) *public* would let any code mutate it again and reintroduce the very `_signflip`/`_blocks` desync #841 removes. So I'd push back on "make it public": private-member + single-friend is strictly more encapsulated and is the right choice.

That said, this connects to **finding #1 in my review above**: the gateway returns a raw mutable `std::vector<cytnx_bool>&`, so it's "one door, but the door is wide open" — any linalg code can resize/reorder it out of lockstep with `_blocks`. The stronger version (which also softens IvanaGyro's concern) is to **narrow the gateway to the specific operations** the backend needs — e.g. `append_false(n)` and `erase_at(indices)` that mutate `_signflip` and `_blocks`/qnums together — instead of handing out the bare vector. That keeps the single-friend encapsulation *and* makes the invariant hard to break, rather than merely hard to reach.

### Net
These are clarity/design-polish items, consistent with my earlier finding #1 — they don't change the correctness verdict (**approve-with-nits, pending a fresh full-test CI run**), but the rename is cheap and worth doing, and the "narrow the gateway" refactor would be a good follow-up (or a small addition here).

IvanaGyro:
> `std::vector<cytnx_bool>&`, so it's "one door, but the door is wide open"

The correct or better design is to only modify `_signflip` while preserving consistency and integrity. This means the current usage of the accessor should be changed. Not only change the var name. Make the public member var to private friend method seems narrow a little bit the accessibility but it didn't explicitly avoid developers to do this dangerous action.

> The stronger version (which also softens IvanaGyro's concern) is to **narrow the gateway to the specific operations** the backend needs — e.g. `append_false(n)` and `erase_at(indices)` that mutate `_signflip` and `_blocks`/qnums together — instead of handing out the bare vector. That keeps the single-friend encapsulation _and_ makes the invariant hard to break, rather than merely hard to reach.

I see the bot said what mean here.

For the naming problem, prefix var names with  `_` is Python semantics of "private". C++ can just put things under private section to make the var private, so there is no need to apply this semantics. We are using AI agents to land more code. Changing all vars names of the whole repo becomes easier, we can decide if want to use snake case for function name or follow the Google Style guide which uses pascal case instead. We have landed many breaking changes since last release, so the next release must be v2.0.0. It is a good time to unify the var naming rules.

pcchen:
## Gateway redesign pushed — `_fermionic_signflip_()` is gone

Implemented the narrowing @IvanaGyro asked for (and that the earlier review update sketched): the friended bare-`vector&` accessor is removed entirely, replaced by two invariant-preserving members on `BlockFermionicUniTensor`:

- **`reset_signflip_()`** — `_signflip.assign(_blocks.size(), false)`. Used by the decomposition internals right after (re)building `_blocks`, so the ledger is sized to the blocks *by construction* (12 construction sites in Svd/Gesvd/Rsvd/Qr/Eig/Eigh/ExpH/ExpM + 1 internal site).
- **`erase_signflip_(positions)`** — erases ledger entries with the same index list that erased the blocks, then **throws if `_signflip.size() != _blocks.size()`** afterwards. All 18 truncation sites (Svd_truncate/Gesvd_truncate/Rsvd) erase blocks immediately before flips, so the postcondition actively enforces the #841 lockstep instead of merely hiding the vector — a caller that forgets to erase blocks first now fails loudly.
- Plain **reads** (8 sites) switch to the existing public `signflip()` const accessor — they never needed mutable access.

Consequences:
- No `friend` declarations left; `_signflip` is private with no escape hatch.
- The reserved-looking `_fermionic_signflip_` name is gone with the gateway, resolving the naming thread as a by-product; the new names follow the codebase's trailing-underscore mutator convention.
- On the "public vs private" question: the two mutators are public members — safe to expose because each preserves the lockstep invariant (they can change *values*, which is already possible through `get_blocks_()`, but cannot desynchronize the bookkeeping).

Verified locally (Release, no HPTT): 1166 tests ran, 1155 passed, 11 skipped, 0 failed -- including the #841 regression test `BkFUt_TruncateKeepsSignflipInLockstepWithNblocks`. Formatted with clang-format 14.0.6. Pushed as a59e3fbd.

@IvanaGyro would you re-review? I believe this addresses both inline comments and the follow-up.

include/UniTensor.hpp
pcchen: Resolved in a59e3fbd by removing the gateway altogether: the `_fermionic_signflip_` name no longer exists. The replacements (`reset_signflip_()` / `erase_signflip_()`) follow the codebase's trailing-underscore in-place-mutator convention with no leading underscore. (On the side question: #1008 proposes a CLAUDE.md/AGENTS.md with style rules for bots.)

include/UniTensor.hpp
pcchen: Agreed, and a59e3fbd lands on the public side of this: the friend/private gateway is gone, and the two replacement mutators (`reset_signflip_()`, `erase_signflip_()`) are public members -- safe to expose because each preserves the `_signflip`/`_blocks` lockstep (erase_signflip_ actively throws on size divergence). `_signflip` itself stays private with no remaining escape hatch.

yingjerkao:
Merged `master` (98 commits) into this branch — now `MERGEABLE` (was `CONFLICTING`). Four conflicts resolved (merge commit `c663c2f3`):

- **`pytests/apply_test.py`** — dropped this branch's duplicate `_are_nearly_eq` in favor of master's canonical copy (landed via #985; functionally identical). Kept the immutable-Bond adaptation: `_make_bfut3` now uses `.redirect()` instead of `.redirect_()`, since this PR removes the in-place `redirect_` binding (Python binds only `redirect` here).
- **`Transpose_()` in DenseUniTensor / BlockUniTensor / BlockFermionicUniTensor** — took master's cleaner `std::ranges` reverse-index computation, but swapped its in-place `bond.redirect_()` for this PR's immutable `bond = bond.redirect()`. Numerically identical (`idxnum+1 == rank`).

Verified the only public `Bond` mutators this PR removes (`redirect_`, `combineBonds_`) have no remaining live callers. **openblas-cpu build clean; full `pytests/` 139 passed / 1 skipped**; `Transpose_` on tagged/block tensors checked directly (bonds redirected, order reversed, original intact).

Note for whoever lands this: the apply-test move + helper rewrite are now on master via #985, so this PR no longer carries them — the collision flagged earlier is resolved.

include/UniTensor.hpp
```diff
@@ -1868,10 +1866,15 @@ namespace cytnx {
       _inner_to_outer_idx;  // stores the qindices for each block
     std::vector<Tensor> _blocks;
     Tensor NullRefTensor;  // this returns when accessed block does not exists!
-    // additional informaiton for fermions:
-    std::vector<cytnx_bool>
-      _signflip;  // if true, the sign of the corresponding block needs to be flipped
 
+   private:
```
IvanaGyro: private section should at the end of the class by the style guide.

pytests/binding_inplace_test.py
```diff
@@ -58,13 +58,27 @@ def test_bond_group_duplicates_tuple():
...
+
+def test_bond_removed_mutators_raise_attribute_error():
+    """#846: Bond's in-place mutators (set_type, redirect_, combineBond_,
+    clear_type) are removed from the public interface; only the return-new
+    forms (retype, redirect, combineBond) remain. group_duplicates_ is also
+    gone: it was a misbinding of the out-of-place C++ group_duplicates under
+    an in-place-mutator name (the non-underscore group_duplicates stays)."""
+    b = cytnx.Bond(2, cytnx.BD_KET)
+    for name in ("set_type", "redirect_", "combineBond_", "clear_type", "group_duplicates_"):
+        assert not hasattr(b, name), f"Bond.{name} should have been removed"
```
IvanaGyro, line 81: This does not help for the future code.

src/linalg/Gesvd.cpp
```diff
@@ -163,18 +163,19 @@ namespace cytnx {
           cytnx::UniTensor &Cy_U = outCyT[t];
           Cy_U._impl->_is_tag = true;
           for (int i = 0; i < Cy_U.rowrank(); i++) {
-            Cy_U.bonds()[i].set_type(Tin.bonds()[i].type());
+            Cy_U._impl->_bonds[i] = Cy_U._impl->_bonds[i].retype(Tin.bonds()[i].type());
           }
-          Cy_U.bonds().back().set_type(cytnx::BD_BRA);
+          Cy_U._impl->_bonds.back() = Cy_U._impl->_bonds.back().retype(cytnx::BD_BRA);
```
IvanaGyro, line 168: This is still dangerous operation. Changing the bond direction on a block tensor may break symmetries.

IvanaGyro: Need to address that block tensors are not fully encapsulated on bond directions.

src/linalg/Gesvd_truncate.cpp
```diff
@@ -306,20 +307,20 @@ namespace cytnx {
       S.get_itoi() = new_itoi;
       if (!to_be_removed.empty()) {
         vec_erase_(S.get_blocks_(), to_be_removed);
-        vec_erase_(S.bonds()[0].qnums(), to_be_removed);
+        vec_erase_(S._impl->_bonds[0]._impl->_qnums, to_be_removed);
         if (Tin.uten_type() == UTenType.BlockFermionic) {
-          vec_erase_(S.signflip_(), to_be_removed);
+          static_cast<BlockFermionicUniTensor *>(S._impl.get())->erase_signflip_(to_be_removed);
```
IvanaGyro, line 312: Allowing to remove blocks in public API of block tensors is also suspicious. Symmetry consistency is not checked by internal functions. It is easy to get wrong for developers. The better workaround may be creating a new one and shadow copy the blocks that need to be preserve. Need to check the shape of block internally while copying.

src/BlockFermionicUniTensor.cpp
```diff
@@ -60,6 +60,16 @@ namespace cytnx {
...
+
+  void BlockFermionicUniTensor::erase_signflip_(const std::vector<cytnx_uint64> &positions) {
+    vec_erase_(this->_signflip, positions);
+    cytnx_error_msg(this->_signflip.size() != this->_blocks.size(),
+                    "[ERROR][BlockFermionicUniTensor][erase_signflip_] after erasing, _signflip "
+                    "(len %d) is out of lockstep with _blocks (len %d); erase the blocks with the "
+                    "same index list before erasing the signflips.%s",
+                    (int)this->_signflip.size(), (int)this->_blocks.size(), "\n");
+  }
+
```
IvanaGyro, line 72: The better API is `remove_blocks`, which will also remove signflips. Although, it is still dangerous....

tests/linalg_test/linalg_test.cpp
```diff
@@ -1141,6 +1141,41 @@ TEST_F(linalg_Test, BkFUt_GesvdTruncateUnitaryAndReconstruction) {
...
+  for (auto trunc : {linalg::Svd_truncate(Msf, 1, 0., true, 0),
+                     linalg::Gesvd_truncate(Msf, 1, 0., true, true, 0)}) {
+    ASSERT_EQ(trunc.size(), 3u);  // S, U, vT (return_err=0)
+    const UniTensor &S = trunc[0];
+    const UniTensor &U = trunc[1];
+    const UniTensor &vT = trunc[2];
+
+    EXPECT_EQ(S.signflip().size(), S.Nblocks());
+    EXPECT_EQ(U.signflip().size(), U.Nblocks());
+    EXPECT_EQ(vT.signflip().size(), vT.Nblocks());
```
IvanaGyro, line 1167: If there is `remove_blocks` function in BlockUniTensor, this should be tested on BlockUniTensor instead.

IvanaGyro (approved):
Better than the original. Other encapsulation issue can be fixed in the future.

yingjerkao:
Brought the branch up to date with master (#1039, #1047) and addressed the review comments:

**CI failure fixed (was `UNSTABLE`).** Merging current master surfaced a compile break that the clean textual merge didn't flag: #1026 added `UniTensor_baseTest.rank_detects_label_bond_mismatch`, which desyncs bonds/labels via `bonds().pop_back()` — no longer valid now that this PR makes `bonds()` const. Since that desync is unreachable through the public API by design (the point of this PR), the test now `const_cast`s the internal `_bonds` vector to still exercise `rank()`'s defensive size-mismatch guard. Reproduced CI's `RUN_TESTS` editable-install compile locally; 38 UniTensor_base/BlockFermionic/Bond gtests pass.

**@IvanaGyro "private section should be at the end of the class"** — fixed in 764337b3: moved `BlockFermionicUniTensor`'s `_signflip` `private:` block (it was sandwiched between two `public:` sections) to the end. Pure reorder; the reset_/erase_signflip_ gateway and #841 lockstep are unchanged.

**Deferred encapsulation comments** (Gesvd bond-direction change; `Gesvd_truncate`/`BlockFermionicUniTensor` block removal; a `remove_blocks` API; moving the linalg test to `BlockUniTensor`) — these are the "other encapsulation issue can be fixed in the future" items from your approval. They're genuine follow-up work (block tensors aren't yet fully encapsulated on bond directions / block removal), better scoped to a dedicated PR than bundled here. Happy to open a tracking issue capturing them.

yingjerkao:
The deferred block-tensor encapsulation comments (Gesvd bond retype, `Gesvd_truncate` block removal, `remove_blocks` API, and the test-location note) are now tracked in #1051 — to be tackled in a dedicated PR rather than expanding #1001's scope.

PR #1004, opened 2026-07-06
Use cuda::std::complex for typed GPU complex dispatch

### LINKED ISSUE #994: GPU build broken: cuKron_kernel fails for mixed ComplexFloat×Double (regression from #858)

### LINKED ISSUE #999: GPU build broken: cuKron fails to compile (CUDA 13) after the #982 type_promote change
## Summary

With `USE_CUDA=ON`, the library **fails to compile** — `Kron.cpp` (via `cuKron_internal.cuh`) no longer type-checks:

```
/home/.../src/backend/linalg_internal_gpu/cuKron_internal.cuh(49): error: no operator "=" matches these operands
/home/.../src/backend/linalg_internal_gpu/cuKron_internal.cuh(49): error: no operator "=" matches these operands
gmake[3]: *** [.../cytnx.dir/src/linalg/Kron.cpp.o] Error 1
```

Line 49 is the Kron kernel body:

```cpp
out[blockIdx.x * blockDim.x + threadIdx.x] = Lin[x] * Rin[y];   // out is TO*, Lin is TL*, Rin is TR*
```

This blocks the entire GPU library build (nothing downstream links), so no GPU tests can run.

- **Environment:** CUDA 13.0, `cmake -S . -B build_gpu -DUSE_CUDA=ON -DRUN_TESTS=ON`
- **Where:** present on `master` (tip `d02cb299`, the #982 merge). Reproduces on a fresh CUDA build of master; independent of any in-flight PR.

## Root cause

`Kron` allocates its output buffer and derives the kernel's output element type through **two different** promotion mechanisms that #982 made inconsistent:

```cpp
// src/linalg/Kron.cpp
Tensor out(new_shape, Type.type_promote(Tl.dtype(), Tr.dtype()), Tl.device());   // L66: runtime enum promotion
...
using out_type = Type_class::type_promote_from_gpu_pointer_t<decltype(tl), decltype(tr)>;  // L88: compile-time trait
cytnx::linalg_internal::cuKron_general(out.gpu_ptr_as<out_type>(), tl, tr, ...);
```

The kernel then computes `Lin[x] * Rin[y]` using the device `operator*` overloads in `src/utils/cucomplex_arithmetic.cu` (e.g. `cuDoubleComplex * cuFloatComplex -> cuDoubleComplex`) and assigns it to `out` (`TO*`). #982 changed `Type_class::type_promote` to promote across the real/complex boundary (`ComplexFloat x Double -> ComplexDouble`) but the compile-time `type_promote_from_gpu_pointer_t` / the device product type were not brought in line, so for a mixed complex/real (or mixed-precision) pair the product type no longer matches `TO` and the assignment fails to compile.

This is the **same class of defect** as the `Outer`/`Exp` bugs fixed in the review of #984 (output dtype chosen by one rule, kernel writes a different element type) — but here it surfaces at **compile time** rather than as a silent runtime buffer mismatch. The `#984` review branch (`claude/amazing-lederberg-7cf653`) confirms it is pre-existing: it reproduces with that branch's changes stashed. (A pre-#982 branch such as #980's base compiles `cuKron` fine.)

## Suggested fix

Make `Kron` follow the same pattern `Matmul`/`Outer` now use: compute the promoted dtype once, `astype` **both** operands to it, and dispatch by the promoted dtype so the kernel instantiates as `cuKron_general<TO, TO, TO>` (`TO * TO -> TO`, assignable). Then audit the other GPU kernels that still derive the output type from a `type_promote_from_*pointer_t` trait / dispatch by raw `[TL][TR]` — they likely share the same latent inconsistency with the post-#982 `type_promote`.

## Broader context

This is concrete support for @ianmccul's point on #984 that changing `type_promote` while backends still compute/reinterpret types by the old rule is unsafe, and argues for the opaque-`dtype` direction (#941) before further `type_promote`-dependent changes.

ianmccul:
This is an alternative fix to #995.

- use `cuda::std::complex<float/double>` as the typed GPU complex element type in `Type_list_gpu`
- keep `cuComplex` / `cuDoubleComplex` only at explicit CUDA library ABI boundaries
- teach Cytnx type traits about `cuda::std::complex`
- update templated GPU Add/Sub/Mul/Div/Kron/Sort paths to use normal C++ complex arithmetic in the promoted output type

Fixes #994
Fixes #999 
Supersedes #995

## Motivation

`cuComplex` is a C ABI type, not a C++ arithmetic type. Using it as the element type in typed C++/CUDA templates makes mixed real/complex dispatch fragile and forces local helper functions for operations that should just be ordinary arithmetic.

`cuda::std::complex` is available in the CUDA toolchain already used by Cytnx and is the better type for typed CUDA kernel code. CUDA library calls that require `cuComplex` pointers can still cast explicitly at the call boundary.

## Notes

This PR does not attempt to rewrite all older generated CUDA kernels or cuBLAS/cuSOLVER call sites. Those still use the existing `cuComplex` ABI where appropriate.

In particular, some large internal `.cu` kernel dispatch tables still cast `void*` buffers directly to `cuComplex` / `cuDoubleComplex`. This PR leaves those paths alone and only changes the typed GPU dispatch surface.

## Tests

- `clang-format-14 -i <touched files>`
- `git diff --check`
- `cmake --build build/openblas-cuda --target cytnx -j 8`
- `cmake --build build/openblas-cuda --target test_main -j 8`
- `ctest --test-dir build/openblas-cpu --output-on-failure -j 8`
  - `1086/1086` passed
- `cmake --build build/openblas-cuda --target gpu_test_main -j 8`
- `ctest --test-dir build/openblas-cuda -I 1087,1852 --output-on-failure -j 8`
  - `763/766` passed
  - failures appear unrelated to this PR:
    - `linalg_Test.gpu_BkFUt_Qr`: GPU QR not implemented without cuQuantum
    - `Gesvd_truncate.gpu_flag_combinations`: `cusolverDnDgesvdj` execution failure in
      the real-double cuSOLVER path
    - `Rsvd.gpu_flag_combinations`: same real-double cuSOLVER path

### DIFF
--- a/include/Tensor.hpp
+++ b/include/Tensor.hpp
@@ -514,9 +514,10 @@ namespace cytnx {
 
     // Convert storage to the logical Cytnx element pointer type.
     //
-    // Use ptr_as<T>() for host code and raw memory copies, including cudaMemcpy. For complex
-    // tensors, T is std::complex<...>. Use gpu_ptr_as<T>() only for CUDA APIs that require CUDA
-    // ABI element types such as cuDoubleComplex or cuComplex.
+    // Use ptr_as<T>() for host code and raw memory copies. For complex tensors, T is
+    // std::complex<...>. Use gpu_ptr_as<T>() for CUDA kernel code; for complex tensors, T is
+    // cuda::std::complex<...>. CUDA library APIs that require cuComplex ABI pointers should cast
+    // explicitly at the library-call boundary.
     template <typename T>
     T *ptr_as() const {
       cytnx_error_msg(this->dtype() != Type_class::cy_typeid_v<std::remove_cv_t<T>>,
@@ -535,10 +536,11 @@ namespace cytnx {
     // convert this->_impl->_storage->Mem to a typed variant of pointers, excluding void*
     gpu_pointer_types gpu_ptr() const;
 
-    // Convert storage to the CUDA ABI element pointer type.
+    // Convert storage to the CUDA kernel element pointer type.
     //
-    // This is for CUDA/cuBLAS/cuSOLVER/kernel interfaces. Generic host code and cudaMemcpy byte
-    // copies should use ptr_as<T>() so the logical Cytnx dtype mapping is checked.
+    // This is for CUDA kernel interfaces. Generic host code should use ptr_as<T>() so the logical
+    // Cytnx dtype mapping is checked. CUDA library APIs that require cuComplex ABI pointers should
+    // cast explicitly at the library-call boundary.
     template <typename T>
     T *gpu_ptr_as() const {
       cytnx_error_msg(
--- a/include/Type.hpp
+++ b/include/Type.hpp
@@ -13,6 +13,10 @@
 
 #include "cytnx_error.hpp"  // also brings in cuComplex.h
 
+#ifdef UNI_GPU
+  #include <cuda/std/complex>
+#endif
+
 #define MKL_Complex8 std::complex<float>
 #define MKL_Complex16 std::complex<double>
 
@@ -51,6 +55,11 @@ namespace cytnx {
   typedef std::complex<double> cytnx_complex128;
   typedef bool cytnx_bool;
 
+#ifdef UNI_GPU
+  using cytnx_cuda_complex64 = cuda::std::complex<float>;
+  using cytnx_cuda_complex128 = cuda::std::complex<double>;
+#endif
+
   namespace internal {
     template <class>
     struct is_complex_impl : std::false_type {};
@@ -58,11 +67,21 @@ namespace cytnx {
     template <class T>
     struct is_complex_impl<std::complex<T>> : std::true_type {};
 
+#ifdef UNI_GPU
+    template <class T>
+    struct is_complex_impl<cuda::std::complex<T>> : std::true_type {};
+#endif
+
     template <typename>
     struct is_complex_floating_point_impl : std::false_type {};
 
     template <typename T>
     struct is_complex_floating_point_impl<std::complex<T>> : std::is_floating_point<T> {};
+
+#ifdef UNI_GPU
+    template <typename T>
+    struct is_complex_floating_point_impl<cuda::std::complex<T>> : std::is_floating_point<T> {};
+#endif
 
     template <std::size_t I, typename T, typename Tuple>
     constexpr std::size_t index_in_tuple_helper() {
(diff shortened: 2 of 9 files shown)


ianmccul:
/gemini review
@codex review

ianmccul:
I have since reconfigured with CUQUANTUM and the previous failures now pass:
```text
linalg_Test.gpu_BkFUt_Qr                Passed
Gesvd_truncate.gpu_flag_combinations    Passed
Rsvd.gpu_flag_combinations              Passed
```

yingjerkao (approved):
The modified code LGTM.

ianmccul:
For bookkeeping: this should be treated as fixing #994 and #999, unless someone can reproduce after #1004. It also supersedes #995, which was the narrower Kron-only fix for the same `cuFloatComplex * double -> cuFloatComplex` failure mode.

PR #998, opened 2026-07-06
fix(UniTensor): rebuild block metadata instead of mutating shared blocks in-place

yingjerkao:
## Problem

Fixes the core of #724. UniTensors can share block Tensors (e.g. `relabel()` copies the block handles), but the in-place `permute_`/`reshape_`/`contiguous_`/`combineBonds` implementations mutated per-block metadata directly — corrupting every other holder of the same block (wrong shapes, out-of-bounds reads).

## Fix

Per the approach agreed in the issue thread (manuschneider's proposal, endorsed by ianmccul): 13 sites across `DenseUniTensor`/`BlockUniTensor`/`BlockFermionicUniTensor` now rebuild block metadata via the non-mutating `Tensor::permute()/reshape()/contiguous()` and rebind (`_blocks[i] = _blocks[i].permute(...)`) — storage stays shared, metadata detaches. The one deliberate remainder: `Block*UniTensor::contract`'s transient mutate-restore of blocks (three hot-path branches) is marked with in-code `KNOWN ISSUE (#724)` blocks and breadcrumbs and is being fixed separately (in-flight follow-up session).

## Testing

Six new gtests: shared-block `permute_` (Dense, Block, BlockFermionic — the latter also pins the signflip vector), shared-block `reshape_` (Dense), and two non-shared control tests. Red-proved by reverting fix sites. Full suite: 1073 passed / 11 skipped / only the 4 pre-existing failures addressed by #977.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #855, opened 2026-05-30
Unaddressed review items from merged PR #760 (mixed real/complex BlockUniTensor contract)


IvanaGyro:
(Issue written by Claude on behalf of @IvanaGyro.)

## Context

PR #760 was merged (merge commit `24e8cc6c`, current `master` HEAD). Its review — https://github.com/Cytnx-dev/Cytnx/pull/760#pullrequestreview-4055172015 (by @pcchen) — raised 9 items that were listed under "Recommended Action Before Merge". I re-checked each against current `master`: **all of them are still present.** Filing this so they aren't lost now that the PR is merged. Line numbers below are on `master @ 24e8cc6c`.

## Critical

**1. Missing LHS cast in the MKL contract path (fragile correctness)** — `src/BlockUniTensor.cpp:1040` (`#ifdef UNI_MKL` block), `src/BlockFermionicUniTensor.cpp:1614`.
The cast guard only casts RHS (`if (Rtn->dtype() != common_dtype) …`); it never casts `this->_blocks` when `this->dtype() != common_dtype`. For `real.contract(complex)` the LHS stays `Double` and RHS is already `ComplexDouble`, so `linalg::Matmul(Double_blocks, ComplexDouble_blocks)` is called with mismatched types. It currently works only because `Matmul` promotes internally — undocumented side-effect that breaks if that promotion changes or the `gemm_batch` path is re-enabled. Cast LHS symmetrically (or document the dependency explicitly).

**2. Scalar (all-legs) contraction path ignores `common_dtype`** — `src/BlockUniTensor.cpp:~847`, `src/BlockFermionicUniTensor.cpp:~1402`.
The scalar branch assigns `tmp->_block = linalg::Vectordot(...)` (and `+=`) without initializing from `common_dtype`; the output dtype is whatever `Vectordot` returns, so the `common_dtype` fix has no effect on this path.

## Important

**3. Dead variables allocated on every contraction call** — `src/BlockUniTensor.cpp:991-997`, `src/BlockFermionicUniTensor.cpp:1565-1571`.
`transs, ms, ns, ks, LMems, RMems, CMems, group_size, alphas` are sized to `Rtn->_blocks.size()` every call but unused (the `gemm_batch` code is commented out). Delete them with the dead code.

[...]

PR #1005, opened 2026-07-07
fix(UniTensor): thread non-mutating block views through contract()

### LINKED ISSUE #724: permute_ destroys data structure of UniTensor with shared blocks
If two 'UniTensor's have the same blocks, and I use `permute_` on one of them, then the data structure of the other `UniTensor` is not consistent with its metadata anymore.

Here is an example that looks innocent enough:
```
uT = cytnx.UniTensor.arange(6).reshape_(2,3).set_name("uT").set_rowrank(1)
print(uT) #shows Tensor with shape (2,3)
uT2 = uT.relabel(["a","b"]).set_name("uT2").permute_([1,0]).set_name("uT2")
uT2[0,1] = 9
uT.print_diagram() #has shape (2,3)
print(uT) #shows Tensor with shape (3,2)!!!
uT2.print_diagram() #has shape (3,2)
print(uT2) #shows Tensor with shape (3,2)
```
Output:
```
-------- start of print ---------
Tensor name: uT
Tensor type: Dense
is_diag    : False
contiguous : True

Total elem: 6
type  : Double (Float64)
cytnx device: CPU
Shape : (2,3)
[[0.00000e+00 1.00000e+00 2.00000e+00 ]
 [3.00000e+00 4.00000e+00 5.00000e+00 ]]




-----------------------
tensor Name : uT
tensor Rank : 2
block_form  : False
is_diag     : False
on device   : cytnx device: CPU
          ---------     
         [/](https://file+.vscode-resource.vscode-cdn.net/)         \    
   0 ____| 2     3 |____ 1
         \         [/](https://file+.vscode-resource.vscode-cdn.net/)    
          ---------     
-------- start of print ---------
Tensor name: uT
Tensor type: Dense
is_diag    : False
contiguous : False

Total elem: 6
type  : Double (Float64)
cytnx device: CPU
Shape : (3,2)
[[0.00000e+00 9.00000e+00 ]
 [1.00000e+00 4.00000e+00 ]
 [2.00000e+00 5.00000e+00 ]]




-----------------------
tensor Name : uT2
tensor Rank : 2
block_form  : False
is_diag     : False
on device   : cytnx device: CPU
          ---------     
         [/](https://file+.vscode-resource.vscode-cdn.net/)         \    
   b ____| 3     2 |____ a
         \         [/](https://file+.vscode-resource.vscode-cdn.net/)    
          ---------     
-------- start of print ---------
Tensor name: uT2
Tensor type: Dense
is_diag    : False
contiguous : False

Total elem: 6
type  : Double (Float64)
cytnx device: CPU
Shape : (3,2)
[[0.00000e+00 9.00000e+00 ]
 [1.00000e+00 4.00000e+00 ]
 [2.00000e+00 5.00000e+00 ]]
```

This leads to errors that are difficult to find.

I suggest that `UniTensor.permute_()` should call `Tensor.permute()` internally for each block. The metadata of the `Tensor` is copied in this case, but the data in memory remains untouched, so the overhead is not large.

yingjerkao:
Completes the #724 fix by converting the last remaining shared-block mutation site — the KNOWN-ISSUE remnant documented in-code by the parent PR #998. **Stacked on #998** (base = `fix/blockut-permute-shared-blocks`); the diff here is a single commit.

## Problem

`BlockUniTensor::contract` and `BlockFermionicUniTensor::contract` `permute_()`/`reshape_()`'d the operands' blocks in place before each Matmul/Gemm_Batch call and restored the metadata afterward. Any UniTensor sharing those blocks (e.g. via `relabel()`, which is documented to share block storage):

- observed transiently corrupted blocks during the contraction,
- was left with replaced, permuted (non-contiguous) block storage afterward even though values were restored, and
- was corrupted permanently if an exception fired before the restore ran.

In particular, contracting a UniTensor with a `relabel()` of itself threw a rank-mismatch error from `permute_` on an already-reshaped shared block.

## Fix

In all three branches of both files (UNI_MKL integer-dtype Matmul fallback, UNI_MKL fp/complex Gemm_Batch, non-MKL Matmul fallback), the rank-2 matrix views of the blocks are now built with the non-mutating `Tensor::permute()`/`reshape()` into local variables instead of mutating and restoring the blocks. Right-block views are cached per block index (the same right block can pair with several left blocks), preserving the old code's one-copy-per-participating-block cost. The Gemm_Batch eager cast-clone of the right operand (`clone_meta`/`astype`/manual `delete`) is replaced by a lazy `astype(common_dtype)` when each right block's view is built, which also removes the manually-deleted heap clone from the exception path.

## Tests (TDD red/green)

For both `BlockUniTensor` and `BlockFermionicUniTensor`:

- `ContractAliasedSharedBlocksOperandsIntact` — contracts a UniTensor with a `relabel()` of itself, so the operands alias each other's blocks (including block-with-itself pairs); pre-fix this threw mid-contraction. Asserts the result matches contracting independent clones and that both operands keep their values, shapes, and contiguity.
- `ContractLeavesSharedBlockObserverUntouched` — a third UniTensor sharing an operand's blocks must observe no change; pre-fix it ended up with replaced, non-contiguous block storage.

No dedicated exception-path test: on non-MKL builds no reachable error fires inside the (now removed) mutation window on valid inputs, and post-fix there is no mutation to leak in the first place.

The UNI_MKL branches are not compiled by the local `USE_MKL=OFF` build; they were additionally syntax/type-checked with `-DUNI_MKL -fsyntax-only`.

Full C++ suite on top of #998: 1092 ran / 1077 passed / 11 skipped / 4 known-failing (pre-existing `linalg_Test.BkUt_*Svd_truncate_return_err*` failures, unrelated to this change).

**Merge order:** merge this PR into `fix/blockut-permute-shared-blocks` before #998 merges, or merge #998 first and let GitHub retarget this PR to `master`. Fixes #724 once both land.

### DIFF
--- a/src/BlockFermionicUniTensor.cpp
+++ b/src/BlockFermionicUniTensor.cpp
@@ -1495,26 +1495,23 @@ namespace cytnx {
 
         std::vector<cytnx_uint64> Lgbuffer;
         std::vector<cytnx_uint64> itoiR_idx;
-        std::vector<cytnx_uint64> oldshapeL;
-        std::vector<std::vector<cytnx_uint64>> oldshapeR(Rtn->_blocks.size(),
-                                                         std::vector<cytnx_uint64>());
+        // The operands' blocks may be shared with other UniTensors (#724), so they must not
+        // be permute_()/reshape_()'d in place. Matrix views are built with the non-mutating
+        // permute()/reshape() instead; right-block views are cached here since the same
+        // right block can pair with several left blocks.
+        std::vector<Tensor> Rmat(Rtn->_blocks.size());
+        std::vector<bool> Rmat_ready(Rtn->_blocks.size(), false);
         std::vector<std::vector<cytnx_uint64>> oldshapeC;
         std::vector<bool> reshaped(tmp->_blocks.size(), false);
         for (cytnx_int64 a = 0; a < tmp->_blocks.size(); a++) {
           oldshapeC.push_back(tmp->_blocks[a].shape());
         }
-        std::vector<cytnx_uint64> mapperL, mapperL_reversed, inv_mapperL(this->rank());
-        std::vector<cytnx_uint64> mapperR, inv_mapperR(rhs->rank());
+        std::vector<cytnx_uint64> mapperL, mapperL_reversed;
+        std::vector<cytnx_uint64> mapperR;
         vec_concatenate_(mapperL, non_comm_idx1, comm_idx1);
         std::reverse(comm_idx1.begin(), comm_idx1.end());
         vec_concatenate_(mapperL_reversed, non_comm_idx1, comm_idx1);
         vec_concatenate_(mapperR, comm_idx2, non_comm_idx2);
-        for (int aa = 0; aa < mapperL.size(); aa++) {
-          inv_mapperL[mapperL[aa]] = aa;
-        }
-        for (int aa = 0; aa < mapperR.size(); aa++) {
-          inv_mapperR[mapperR[aa]] = aa;
-        }
 
         // fermion signs
         std::vector<bool> signfliplhs = this->_lhssigns_(mapperL_reversed, comm_idx1.size());
@@ -1546,15 +1543,6 @@ namespace cytnx {
             }
           }
         } else {
-          // KNOWN ISSUE (#724, not fixed by this change): same as the analogous branch in
-          // BlockUniTensor::contract() (src/BlockUniTensor.cpp) — the branches below temporarily
-          // mutate this->_blocks[a] / Rtn->_blocks[b] / tmp_Rtn->_blocks[b] in place via
-          // permute_()/reshape_(), then restore afterward. If those blocks are shared with
-          // another UniTensor, that other UniTensor is transiently (or, on an exception,
-          // permanently) corrupted. Left as a follow-up; see BlockUniTensor.cpp for details.
-          BlockFermionicUniTensor *tmp_Rtn = Rtn;
-          bool tmp_rtn_is_casted = false;
-
           // check if all sub-tensor are same dtype and device
           if (User_debug) {
             bool all_sub_tensor_same_dtype = true;
@@ -1598,14 +1586,14 @@ namespace cytnx {
               itoiR_idx = mp[itoiL_common[a]];
               for (cytnx_uint64 aa = 0; aa < comm_idx1.size(); aa++)
                 comm_dim *= this->_blocks[a].shape()[comm_idx1[aa]];
-              this->_blocks[a].permute_(mapperL);
-              oldshapeL = this->_blocks[a].shape();
-              this->_blocks[a].reshape_({-1, comm_dim});
+              // matrix view of this->_blocks[a]; non-mutating (the block may be shared, #724)
+              const Tensor Lmat = this->_blocks[a].permute(mapperL).reshape({-1, comm_dim});
               for (cytnx_uint64 binx = 0; binx < itoiR_idx.size(); binx++) {
                 cytnx_uint64 b = itoiR_idx[binx];
-                Rtn->_blocks[b].permute_(mapperR);
-                oldshapeR[b] = Rtn->_blocks[b].shape();
-                Rtn->_blocks[b].reshape_({comm_dim, -1});
+                if (!Rmat_ready[b]) {
+                  Rmat[b] = Rtn->_blocks[b].permute(mapperR).reshape({comm_dim, -1});
+                  Rmat_ready[b] = true;
+                }
                 Lgbuffer.resize(non_comm_idx1.size() + non_comm_idx2.size());
                 for (cytnx_uint64 cc = 0; cc < non_comm_idx1.size(); cc++)
                   Lgbuffer[cc] = this->_inner_to_outer_idx[a][non_comm_idx1[cc]];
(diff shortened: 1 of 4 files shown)


pcchen:
## Code Review

Eliminates the last shared-block mutation site: `contract()` on `BlockUniTensor`/`BlockFermionicUniTensor` used to `permute_`/`reshape_` the operands' blocks in place (then restore), transiently — or permanently, on an exception — corrupting any UniTensor sharing those blocks (e.g. a `relabel()`). Now it builds rank-2 matrix views via non-mutating `permute()`/`reshape()` into locals, across all three branches of both files.

### Correctness — verified by reading all 6 branches
- **The view transformation is behavior-preserving.** `permute(mapperL).reshape({-1, comm_dim})` yields the exact tensor value the old `permute_`/`reshape_` produced in place, so `Matmul`/`Gemm_Batch` receive identical inputs; the inverse-permute/reshape restore is correctly dropped since originals are never touched. ✅
- **Right-block caching is sound.** `Rmat[b]` (guarded by `Rmat_ready[b]`) is computed once per right block: `comm_dim` is necessarily equal across all left blocks pairing with a given `b` (else the contraction is invalid), and it's actually *fewer* copies than the old restore-per-pairing. ✅
- **Lazy `astype(common_dtype)`** per right-block view matches the old eager whole-operand cast, and removes the manually-`delete`d heap clone — eliminating the exception-path leak the old `tmp_Rtn` had. ✅
- **Subtle spot done right:** the Gemm target reshape uses `Lmat.shape()[0]` / `Rmat[b].shape()[1]` (the *view* shapes), not the original block shapes — exactly where a naive port would go wrong. ✅
- **Fermionic sign logic is untouched**; `BlockUniTensor` mirrors the fermionic version minus signs. ✅
- **Exception safety is now structural** — nothing is mutated, so there's no window to leak; removing the dedicated exception test is justified.
- **Tests are strong.** Both classes get `ContractAliasedSharedBlocksOperandsIntact` (operands alias via `relabel()`, incl. block-with-itself pairs; asserts correct result vs independent clones, bit-exact operand intactness at `0.0` tol, and `signflip()` equality) and `ContractLeavesSharedBlockObserverUntouched` (a third sharer stays bit-exact + contiguous). These genuinely pin the fix. ✅

### Findings

**1. 🟠 The two MKL branches per file are runtime-untested.** The regression tests and the "1077 passed" local run were `USE_MKL=OFF`, so they only exercise the **non-MKL Matmul** branch. The **MKL integer-Matmul** and **MKL Gemm_Batch** branches — which received the more intricate edits (cached `Rmat[b]`, lazy `astype`, the batch loop, the view-shape reshape target) — were only `-fsyntax-only` checked, and no CI job builds+runs tests with MKL (`coverity-scan` uses the mkl preset but is a static scan). My manual read gives confidence they're correct, but a logic slip there (stale cached view, wrong reshape dim, dtype) would pass every current gate. **Recommend running the new gtests against an MKL build** (locally or a CI job) before merge — the Gemm_Batch path especially.

**2. 🟡 No `BuildAndTest` gtest job on this commit.** Only wheels ran (compilation verified, tests not). The 4 new regression tests are local-only evidence; a fresh full-test CI run would confirm them.

**3. 🟢 Peak-memory note.** `Rmat` retains all participating right-block views for the whole contraction (vs one-at-a-time in the old same-dtype path) — fewer total copies but higher peak. Reasonable trade, and no regression for the dtype-mismatch path (old already cloned all of `Rtn`); worth a one-line mention.

### Verdict
A clean, correct, well-tested completion of the #724 shared-block fix — the transformation is behavior-preserving, the caching and lazy-cast are sound, and the tests pin the aliasing behavior precisely (verified across all six branches by reading). The one real gate is that the **MKL branches have no runtime test coverage** anywhere; I'd want the new tests run against an MKL build before merge. **Approve-with-nits**, pending MKL test coverage.

pcchen:
### Update to my review — the MKL-coverage nit is already tracked (no new issue needed)

On my earlier finding that the MKL contract branches (integer-Matmul fallback + `Gemm_Batch`) are runtime-untested (local suite + CI run `USE_MKL=OFF`): rather than a new issue, this is covered by existing ones —

- **#583** ("Run GitHub action with all allowed configurations") already proposes a CPU CI matrix that includes **MKL** builds. I've added a note there asking that those configs also **run `ctest`** (not just build), so the MKL contract paths get runtime coverage.
- **#855** already tracks MKL-contract-specific test gaps in this exact code (e.g. item 5: "BlockFermionicUniTensor mixed-dtype contract has zero test coverage", plus MKL-path correctness/exception-safety items 1/2/6).

So the nit stands but is **not a blocker** for this PR: I verified all six branches by reading, the non-MKL branch is exercised by the new regression tests and proves the transformation, and the MKL runtime-coverage gap is tracked in #583/#855. Verdict unchanged — approve-with-nit (tracked).

pcchen (approved):
Approving on code correctness. The non-mutating view transformation is behavior-preserving across all six branches (verified by reading): matrix views via `permute()/reshape()`, sound right-block caching, lazy `astype` matching the old eager cast (and killing the exception-path leak), and the Gemm reshape-target correctly taken from the view shapes. Fermionic signs untouched. The two regression tests pin the aliasing fix precisely (operands + observer bit-exact, signflip equality).

Two non-blocking notes before merge:

1. **No `BuildAndTest` on the current checks.** This PR's CI last ran while it was behind-base (pre-#1007), so its C++ gtests are local-only evidence. Now that #1007 (the git-identity fix) is merged and this is retargeted to master, a fresh run should actually execute BuildAndTest — worth triggering (e.g. an empty commit / re-run) before merge so the regression tests are CI-verified, not just local.

2. **MKL branches lack runtime coverage** — tracked (not a new issue): #583 (MKL in the CI matrix; I asked there that it also run `ctest`) and #855 (MKL-contract test gaps). Read-verified here; the non-MKL branch is exercised by the new tests.

Verdict: approve; recommend a fresh CI run (real BuildAndTest) before landing.

PR #1007, opened 2026-07-07
ci: set git committer identity before merging the target branch

yingjerkao:
## Summary

The shared **"Merge with latest target branch (pull_request only)"** step runs

```
git merge --no-edit origin/${{ github.event.pull_request.base.ref }}
```

without a configured git committer identity. Whenever a PR head is **behind its base branch**, the merge is not a fast-forward no-op, so git must create a merge commit — and fails with:

```
fatal: empty ident name (for <runner@...>) not allowed
```

killing the job with exit code 128 before any real work runs (observed 2026-07-06 on PR #984, run [28790515780](https://github.com/Cytnx-dev/Cytnx/actions/runs/28790515780)). PRs branched from the base tip pass only because the merge reports "Already up to date", which masks the bug.

## Fix

Configure the `github-actions[bot]` identity before the merge, exactly as `release_pypi.yml` already does:

```yaml
git config user.name "github-actions[bot]"
git config user.email "github-actions[bot]@users.noreply.github.com"
```

Applied to the four remaining workflows that carry the step:

- `.github/workflows/ci-cmake_tests.yml`
- `.github/workflows/ci-downstream-find-package.yml`
- `.github/workflows/clang-format-check.yml`
- `.github/workflows/conda_build.yml`

The step is otherwise unchanged. (`release_pypi.yml` already had the fix, so it needs no change.)

## Test plan

- [x] `yaml.safe_load` passes on all five workflow files
- [x] Verified every `git merge` in `.github/workflows/` is now preceded by the identity config
- [ ] CI on this PR exercises the merge step itself (this branch is at the master tip, so the merge will be a no-op here; the real proof is the next behind-base PR no longer failing with exit 128)

### DIFF
--- a/.github/workflows/ci-cmake_tests.yml
+++ b/.github/workflows/ci-cmake_tests.yml
@@ -30,9 +30,13 @@ jobs:
       # When re-running a PR job, GitHub Actions uses the merge commit created at
       # the time of the original run, not a fresh merge with the latest target branch.
       # This step ensures we always test against the most recent target branch.
+      # A committer identity is required because a non-fast-forward merge creates
+      # a merge commit.
       - name: Merge with latest target branch (pull_request only)
         if: github.event_name == 'pull_request'
         run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
           git fetch origin ${{ github.event.pull_request.base.ref }}
           git merge --no-edit origin/${{ github.event.pull_request.base.ref }}
           # Refresh submodules in case the merge moved any gitlinks (e.g. cmake_modules/morse_cmake).
--- a/.github/workflows/ci-downstream-find-package.yml
+++ b/.github/workflows/ci-downstream-find-package.yml
@@ -61,9 +61,13 @@ jobs:
 
       # Match ci-cmake_tests.yml: re-runs of a PR job use the original merge
       # commit, so merge the latest target branch to test against current tip.
+      # A committer identity is required because a non-fast-forward merge creates
+      # a merge commit.
       - name: Merge with latest target branch (pull_request only)
         if: github.event_name == 'pull_request'
         run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
           git fetch origin ${{ github.event.pull_request.base.ref }}
           git merge --no-edit origin/${{ github.event.pull_request.base.ref }}
           # Refresh submodules in case the merge moved any gitlinks.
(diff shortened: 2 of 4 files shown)


pcchen:
## Code Review — LGTM

Correct, minimal CI fix. The two `git config` lines are placed first inside the "Merge with latest target branch" `run: |` block, before `git fetch`/`git merge --no-edit` — which is exactly right: `git merge --no-edit` of the target branch creates a non-fast-forward merge commit, and the ephemeral runner has no committer identity by default, so the merge would fail without it. The `github-actions[bot]` identity is the conventional one, indentation matches each block's existing git commands, and the fix is applied uniformly across all 4 workflows.

Full CI is green here — including the `pull_request` merge-target step itself, which validates the fix end-to-end.

Note: this supersedes #1002 (@IvanaGyro), which is a functionally identical fix to the same 4 files (only comment wrapping/indentation differs). Linking it there.

pcchen:
### Context for whoever merges: this bug is currently masking BuildAndTest on behind-base PRs

On the last (unchecked) test-plan item — #1007's own CI can't exercise the fix because this branch is at the master tip, so its merge step is a no-op. That's accurate. But the fix isn't speculative, and the "real proof" already exists:

- Master's `ci-cmake_tests.yml` (the **BuildAndTest** job) still lacks the `git config`, so any PR **behind its base** dies at the "Merge with latest target branch" step with `fatal: empty ident name … not allowed` / exit 128 — before BuildAndTest runs. (Direct evidence: the cited PR #984 run `28790515780` fails exactly this way on Formatting Check.)
- This appears to be the cause of the missing/absent **BuildAndTest** checks on the currently-behind-base PRs in flight (e.g. #997, #1000, #1001, #1005): their **wheel** jobs run (the `release_pypi`/BuildWheel workflow already has the identity fix) but their **BuildAndTest** job does not — matching this failure exactly. On #1001, no BuildAndTest check is present at all.

**So merging #1007 is higher-value than it looks:** it should restore BuildAndTest CI coverage for every behind-base PR, which is what would let the recurring "test evidence is local-only / BuildAndTest didn't run" notes on those PRs finally be cleared by real CI.

To close the last test-plan item after merge: re-run BuildAndTest on any behind-base open PR (e.g. #1001) and confirm it now gets past the merge step.

pcchen (approved):
Approving. Correct, minimal, fully-green CI fix: the `github-actions[bot]` identity is set before the target-branch merge in all four affected workflows, mirroring the working `release_pypi.yml` precedent. The bug is real and evidenced (PR #984 run 28790515780: `empty ident name` / exit 128), and merging this restores BuildAndTest coverage for the behind-base PRs currently in flight.

Issue #650, opened 2025-08-26
Refactor internal functions  using std::variant and visitor pattern


yingjerkao:
In commit https://github.com/Cytnx-dev/Cytnx/commit/710c98c73d24dba02bdbecad54dcfc8dcdf5b6fc , multiple dispatch using `std::variant`
together with the visitor pattern. 

@ianmccul already provided an example implementation in `Kron_internal.hpp` and a simpler example is available here https://en.cppreference.com/w/cpp/utility/variant/visit2

We need to  update other  implementations listed below for both CPU and GPU. 

Some attempt has been tried in #648 by @jysh1214 but we want to align our implementation. 

- [ ] Abs_internal
- [ ] Add_internal
- [ ] Arithmetic_internal
- [ ] Axpy_internal
- [ ] Batch_matmul_internal
- [ ] Conj_inplace_internal
- [ ] Cpr_internal
- [ ] Det_internal
- [ ] Diag_internal
- [ ] Div_internal
- [ ] Eig_internal
- [ ] Eigh_internal
- [ ] Exp_internal
- [ ] Gemm_Batch_internal
- [ ] Gemm_internal
- [ ] Ger_internal
- [ ] Gesvd_internal
- [ ] iAdd_internal
- [ ] iArithmetic_internal
- [ ] iDiv_internal
- [ ] iMul_internal
- [ ] Inv_inplace_internal
- [ ] InvM_inplace_internal
- [ ] iSub_internal
- [x] Kron_internal
- [ ] Lstsq_internal
- [ ] Matmul_dg_internal
- [ ] Matmul_internal
- [ ] Matvec_internal
- [ ] MaxMin_internal
- [ ] memcpyTruncation
- [ ] Mod_internal
- [ ] Mul_internal
- [ ] Norm_internal
- [ ] Outer_internal
- [ ] Pow_internal
- [ ] QR_internal
- [ ] Sdd_internal
- [ ] Sub_internal
- [ ] Sum_internal
- [ ] Trace_internal
- [ ] Tridiag_internal
- [ ] Vectordot_internal

ianmccul:
That is a really annoying function because all it has to do is return a variant with `dtype` as the index and the pointer as the value, but there is no portable way to do that directly! I don't know if the compiler can recognise that all the cases are the same and collapse them to something trivial. It would be interesting to look at the assembly output.

yingjerkao:
Explained by Gemini


[...]

PR #1008, opened 2026-07-07
docs: add CLAUDE.md/AGENTS.md agent guide (#759)

### LINKED ISSUE #759: Introducing Agentic Code Generators into Workflow
We have a lot of unsolved issues and unmerged PRs, and it is hindering our progress. 
I have been experimenting with Claude Code/Codex for generating codes. I think a lot of the refactoring we want to do can be automated through them. Since we are using Codex/Claude to review the PR/commits already, I think this is a good time to have some discussion on how much we want this project to be agent driven.  

I propose we first add CLAUDE.md/AGENT.md to guide the coding agents. Since we have CI and pre-commit hook setup, this should serve as a good safeguard for any future PRs.

We then can fix CI/CD easily for PyPi.

The next stage will have  Codex/Claude to do a full review on the codebase architecture and design. Spot  hot spots that can be optimized away. Code agents are pretty good at profiling and identifying hotspots. 

The third will be have the GPU version improved and optimized. 

 We still need to review changes to make sure the physics is correct. 

After we reach a stable version, we can start adding features and scaling things up. 

I hope this can bring down our open issues and unmerged PR in a few months.

yingjerkao:
## Summary

Seeds the agent-driven workflow proposed in #759 with an operational guide for coding agents (Claude Code, Codex). This is the "Phase 1" safeguard from that issue: a single place that tells an agent how to produce PRs that pass CI, plus the guardrails and domain rules agents otherwise get wrong.

`AGENTS.md` is a **symlink** to `CLAUDE.md` (git mode `120000`) so both toolchains read one source of truth — edit `CLAUDE.md` only. Resolves cleanly on Linux/macOS/WSL, where Cytnx is developed.

## Scope of decisions (per discussion)

- **Shared file** — one canonical `CLAUDE.md`, `AGENTS.md` symlinked to it.
- **Lean operational contract** (~126 lines) — loads fast every session, stays current.
- **Mechanical only** — the phased roadmap stays in #759; this file is *how to work correctly*, not *what to work on*.

## Contents

- **Build** — submodule init + preset configure/build (`CMakePresets.json`), toggle summary.
- **Test** — C++ (`debug-*` preset → `test_main` → `ctest --preset cpu-only`), GPU (`gpu_test_main`, `cpu-and-cuda`), Python (the exact editable-install + `pytest pytests/` CI runs).
- **Format** — clang-format **pinned to v14**; run `pre-commit`, re-add after the hook reformats (a newer clang-format fails CI).
- **CI gates** — the four checks to clear before pushing (clang-format-check, ci-cmake_tests, downstream `find_package`, version-consistency).
- **Commits & PRs** — branch off master, scoped commits, conventional subjects, `Co-Authored-By` trailer, problem→fix→testing body.
- **Guardrails (per #759)** — scoped diffs; **physics/numerical-correctness changes require human review**; flag mixed-dtype changes.
- **Domain gotchas** — dtype enum ordering (lower = higher precision), `Type.type_promote` across the real/complex boundary, GPU in-place kernel gaps (#988), `cuKron` broken under CUDA 13 (#999), the `cytnx_error_msg`/`cytnx::error` macros, and the `std::variant` dispatch direction (#650, #938).
- **Repo map** + pointers to `CONTRIBUTING.md` / `RELEASING.md` / `Readme.md`.

## Notes for reviewers

- All commands were fact-checked against the repo (presets, `test_main`/`gpu_test_main` targets, ctest presets).
- Docs-only change — no CI impact beyond the docs build; the clang-format hook has no C/C++ files to check here.
- Open questions I'd welcome input on: (1) keep `AGENTS.md` as a symlink vs. a one-line pointer file (symlink risk on native-Windows checkouts); (2) whether the `Co-Authored-By` agent-attribution line should be project policy.

Closes #759 is **not** asserted — this is only Phase 1 of that issue; leaving #759 open for the later phases.


pcchen:
## Review

Nice work — this is an unusually well-fact-checked guide. I verified the claims directly against the repo (presets, `test_main`/`gpu_test_main` targets, `.clang-format`, pre-commit v14 pin, submodules, the `pyproject` preset pin + `[dev]` extra, the `Type.hpp` enum ordering, `type_promote`, `CYTNX_FUNC_NAME`, `cytnx_error_msg`, and each CI workflow). Almost everything holds up. One claim is wrong, plus two minor notes.

### 1. The `version-consistency` gate is mischaracterized (describes a CI safety net that doesn't exist)

> `- **version-consistency** — if you change the minimum Python version, update every site listed in CONTRIBUTING.md.`

This conflates two unrelated things:

- The `version-consistency` workflow (display name **"Release metadata consistency"**) runs `tools/check_release_consistency.py` and checks **`version.cmake` / `docs/site_root/versions.json`** — the *release package version* and docs slug. It has nothing to do with the minimum Python version.
- The minimum Python version has **no automated check at all**. `CONTRIBUTING.md` says so explicitly: *"There is no automated check for this — please grep for the old version string."*

So an agent that bumps `requires-python`, sees "version-consistency" listed as the enforcing gate, and assumes CI will catch a missed site, will push drift that CI never validates. Suggest describing this gate as guarding `version.cmake`/`versions.json` release metadata, and keeping the min-Python-version note with the (accurate) "See also → CONTRIBUTING.md" pointer.

### Minor notes (non-blocking)

- **CI-gate identifiers vs. displayed check names.** The guide names the gates `clang-format-check`, `ci-cmake_tests`, `ci-downstream-find-package`, `version-consistency` — these are the workflow *filenames*, but the checks shown on the PR UI read *"Formatting Check"*, *"Tests and Codecov"*, *"Downstream find_package"*, *"Release metadata consistency"*. An agent scanning the PR status list for those exact strings won't find them verbatim; worth aligning or noting they're filenames.
- **`AGENTS.md` symlink on native Windows.** Already flagged as an open question in the PR description — on a checkout with `core.symlinks=false`, `AGENTS.md` materializes as a 9-byte text file containing `CLAUDE.md` rather than resolving. Given Cytnx is developed on Linux/macOS/WSL this is acceptable, but the fallback content is misleading if a Windows agent reads it.

Everything else I spot-checked — presets, `debug-*`/ctest presets, `tests/gpu/`, clang-format v14, dtype enum ordering (ComplexDouble=1 … Int64=5, matching the `dtype <= 2` complex idiom in `Type.hpp`), `type_promote`, `CYTNX_FUNC_NAME`, the `.[dev]` install command, and the pytest invocation — is accurate.

CLAUDE.md
```diff
@@ -0,0 +1,126 @@
...
+
+```bash
+cmake --preset debug-openblas-cuda                # add -DCMAKE_CUDA_ARCHITECTURES=<arch> if autodetect fails
+cmake --build --preset debug-openblas-cuda --target gpu_test_main test_main
+ctest --preset cpu-and-cuda --output-on-failure   # full GPU arith suite is slow (~10 min)
+```
+
+Python — CI installs editable and runs pytest (pyproject pins `--preset=openblas-cpu`):
+
+```bash
```
IvanaGyro, line 52: It is possible to use preset here

CLAUDE.md
```diff
@@ -0,0 +1,126 @@
...
+  ```
+- The clang-format hook rewrites files in place and **aborts the commit** if it
+  changed anything → `git add` the reformatted files and commit again.
+
+## Before you push — the CI gates
+
+Green locally before opening a PR. CI enforces:
+
+- **clang-format-check** (v14) — formatting clean.
+- **ci-cmake_tests** — `ctest` (`test_main`) *and* `pytest pytests/` both pass; Codecov.
```
IvanaGyro, line 103: It is able to set up cuda build environment in cloud agent environment. We can consider to ask build cuda to align with the previous section "use cuda preset if touch cuda code"

CLAUDE.md
```diff
@@ -0,0 +1,126 @@
...
+- **ci-downstream-find-package** — installed `find_package(Cytnx)` still works;
+  don't break exported targets or public headers.
+- **version-consistency** — if you change the minimum Python version, update every
+  site listed in `CONTRIBUTING.md`.
+
+## Commits & PRs
+
+- Branch off `master`; never push straight to `master`.
+- Small, single-purpose commits; conventional subjects: `fix(linalg): …`,
```
IvanaGyro: How about just tell the bots follow conventional commits? It is a commit standard

IvanaGyro: Maybe this should be defined in Contribution.md

CLAUDE.md
```diff
@@ -0,0 +1,126 @@
...
+default), `USE_HPTT`. CPU presets iterate fastest; only use a `*-cuda` preset
+when a change actually touches GPU code.
+
+## Test
+
+C++ tests need `RUN_TESTS=ON` — use a `debug-*` preset:
+
+```bash
+cmake --preset debug-openblas-cpu
+cmake --build --preset debug-openblas-cpu --target test_main
```
IvanaGyro, line 33: While building and running debug cuda preset, some env need to be set to suppress ASAN errro

IvanaGyro (changes requested):

CLAUDE.md
yingjerkao: Could you provide some examples?

CLAUDE.md
yingjerkao: > It is able to set up cuda build environment in cloud agent environment. We can consider to ask build cuda to align with the previous section "use cuda preset if touch cuda code"

Does this mean running agents off Github? Or we can subscribe to the Github GPU workers.

ianmccul:
its worth having a separate GEMINI.md because that is used by the Gemini bot for reviews.  For example it keeps suggesting changes for C++17 compatibility, which is really annoying and pointless.  GEMINI.md can be more focussed on what needs reviewing, rather than developer-orientated CLAUDE.md and AGENTS.md.  See for example https://github.com/Uni20-dev/uni20/blob/main/GEMINI.md.  Its basically a moving target of "stop bringing up these issues in automated reviews".

Gemini free reviews go away in another week, but you could get a gemini free trial for another month, and maybe worth paying for it longer-term.  I have used gemini-cli briefly (now replaced by Antigravity), but only the free tier so I am not sure how good the higher tier models are, but I've been generally impressed with the gemini reviews.

CLAUDE.md
IvanaGyro: `ASAN_OPTIONS='protect_shadow_gap=0:replace_intrin=0:detect_leaks=0'`
Refer to:
https://github.com/google/sanitizers/issues/629#issuecomment-161755902
https://stackoverflow.com/a/68027496

CLAUDE.md
IvanaGyro: > Does this mean running agents off Github?

Yes. In both codex and Claude Code cloud session. cuda compiler and libraries can be installed on a machine without GPU. The limitation is cuda code can be built on those machines but can not be run. We can also build cuda variant with GitHub action without GPU workers.

pcchen:
## Code Review

Useful, accurate agent guide — I spot-verified the content against the repo: every named preset exists in `CMakePresets.json`, the clang-format-v14 pin, dtype-enum ordering (lower index = higher precision), `type_promote` cross-kind behavior, `cytnx_error_msg` → `cytnx::error` → `CytnxError`, the variant-dispatch migration direction, and the CI-gates list all check out. The "domain gotchas" section is exactly the right list — it's the traps this month's PRs actually hit.

### Findings

**1. 🟠 `AGENTS.md` is a broken symlink.** It's committed with git mode `120000` (symlink) but its blob is a 103-byte *markdown string* — a symlink's blob must be exactly the target path. On POSIX checkout this materializes as a **dangling symlink** whose "target" is the literal text `# Agent Guide⏎⏎Please see…`; any non-Claude agent reading `AGENTS.md` gets ENOENT. (GitHub's web UI renders the target text, which is why the PR view looks fine.) Fix: either a true symlink (blob content exactly `CLAUDE.md`) or a regular `100644` file with the pointer text — and align the "AGENTS.md is a symlink to it" sentence accordingly.

**2. 🟡 Consider folding in the review-attribution policy.** There's an (uncommitted) maintainer `CLAUDE.md` convention in use — agent-authored posts carry "Posted by Claude Code on behalf of @user" attribution. That's an operational-contract item for agents and belongs in this canonical file; it would also avoid a local-file collision when this lands.

**3. 🟡 Add the naming conventions @IvanaGyro asked to codify** (in the #1001 thread): leading `_` = member variable, trailing `_` = in-place mutator returning self, snake_case for new code (#836). This PR is the natural home — a short "Naming" section would close that loop.

**4. 🟢 Forward-looking gotcha:** "a length-1 scalar RHS must stay CPU-resident (#988)" describes post-#980/#992 semantics — neither merged yet. Fine as guidance; consider "(once #980/#992 land)".

Content: LGTM. The symlink defect (finding 1) is the one must-fix.

IvanaGyro:
> **Add the naming conventions @IvanaGyro asked to codify** (in the #1001 thread): leading `_` = member variable, trailing `_` = in-place mutator returning self, snake_case for new code (#836). This PR is the natural home — a short "Naming" section would close that loop.

I may have bad wording so the bot cannot understand me.

I suggest follow Google style for naming for all new code except that file has consistent naming other than the style guide. Most of C++ files have messy naming now. It is not benefit to try to follow the messy naming if the file has already messy. Google style picked in this project (not by me). Additionally, leading `_` should be prohibited for new code strictly.

ianmccul:
I suggest also adding to the AI guidance, is to pass simple C++ types by ***value***.  The AI tends to follow the surrounding code so it passes simple values by `const&` when it should pass by value.  For Cytnx scalars, passing by reference will *never* be faster.  Even `complex<double>` is better passed by value, and will actually be passed via CPU registers in some cases.  Passing by reference requires pushing the value onto the stack, passing a pointer to that value into the function, and then the function has to load the pointer and dereference it to retrieve the value.

The other thing is `if (x) cytnx_error_msg(true, "...... %s", "\n");`  But I regard that as _Cytnx house style_ now.  Cytnx wouldn't be complete without it.  But suggest `cytnx_error_msg` adds `"\033[1A"` to the string, to get the extra line back :rofl:

ianmccul:
I don't like the Google style of `T &x`.  It sort-of makes sense in C code for pointers, `T *x`, where `(*x)` is is an object of type `T`, but in C++ we prefer to think that `x` is a "pointer to `T`".  And for references, `T &x` looks very strange to me. There is no sense in which `&x` functions like something of type `T`.

yingjerkao:
I think the easiest way to enforce Google C++ style is using Skills, such as [Google C++ Style Guide Skill](https://mcpmarket.com/tools/skills/google-c-style-guide),[Google Python Style Guide Skill](https://mcpmarket.com/tools/skills/python-style-guide).

PR #1009, opened 2026-07-07
test(pybind): drop shadow-name guard, cover @= identity (follow-up to #986)

yingjerkao:
Follow-up to the merged #986, addressing the two review comments on `pytests/binding_inplace_test.py`.

## Changes

- **Remove `test_shadow_bindings_are_gone`** (per @IvanaGyro, [review comment](https://github.com/Cytnx-dev/Cytnx/pull/986#discussion_r3533995928): *"This test is useless for the future code. This should be clean up."*). It only asserts that the deleted `c__*`/`c_*` shadow names are absent — a one-time migration guard with no value for future code, and the last place in the tree that even referenced those dead names.

- **Add `test_imatmul_preserves_identity`** (per @pcchen's approval follow-up). `@=` is the one genuine behavioral change from #986: on master the wrapper was misspelled `def __imatmul` (missing the trailing `__`), so `a @= b` fell back to `a = a @ b` and rebound `a` to a **new** object. With `__imatmul__` bound correctly it now mutates in place and preserves identity. This path was previously untested.

## Verification

Built the Python module from the post-#986 tree and ran the file:

- New test **passes** on the post-#986 module (`__imatmul__` present, `a @= b` mutates in place, `a is alias`, and the alias observes the result of `a @ b`).
- The same test **fails** on the pre-#986 module (v1.0.0: `a @= b` rebinds `a`, `a is alias` is False) — confirming it genuinely pins the fixed behavior rather than passing vacuously.
- Full file: **7 passed**.

Net: −29/+15 (one dead test out, one meaningful test in).

### DIFF
--- a/pytests/binding_inplace_test.py
+++ b/pytests/binding_inplace_test.py
@@ -67,32 +67,18 @@ def test_bond_redirect__chains():
     assert b.type() == cytnx.BD_BRA
 
 
-def test_shadow_bindings_are_gone():
-    """The c__*/c_* shadow names must no longer be reachable on any instance;
-    only the real dunder/method names remain."""
-    t = cytnx.zeros([2])
-    for name in (
-        "c__iadd__", "c__isub__", "c__imul__", "c__itruediv__", "c__ifloordiv__",
-        "c__ipow__", "c__imatmul__", "cConj_", "cExp_", "cInvM_", "cInv_", "cAbs_", "cPow_",
-    ):
-        assert not hasattr(t, name), f"Tensor.{name} shadow binding still present"
-
-    ut = cytnx.UniTensor(cytnx.ones([2, 2]))
-    for name in (
-        "cConj_", "cTrace_", "cTranspose_", "cnormalize_", "cDagger_", "ctag",
-        "c__ipow__", "cPow_", "cInv_", "ctruncate_", "c_set_name", "c_set_label",
-        "c_set_labels", "c_relabel_", "c_relabels_", "c_set_rowrank_", "cfrom",
-    ):
-        assert not hasattr(ut, name), f"UniTensor.{name} shadow binding still present"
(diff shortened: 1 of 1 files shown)


pcchen:
## Code Review

Clean, correct test-only follow-up to #986 that closes both review comments on `binding_inplace_test.py`. **LGTM.**

- **`test_imatmul_preserves_identity` is correct and non-vacuous.** It snapshots `expected = Dot(a, b)` *before* the `@=` (independent storage), then asserts both identity (`a is alias`, `id(a) == aid`) and correctness (`alias - expected` has zero norm). The `== 0.0` exact check is safe here since the operands are small exact-integer values. Per the PR notes, the test also *fails* on the pre-#986 module (where `@=` rebound `a` to a new object) — so it genuinely pins the fixed behavior rather than passing vacuously.
- **Removing `test_shadow_bindings_are_gone` is reasonable** — it only asserted the deleted `c__*`/`c_*` names are absent (a one-time migration guard with no value for future code), and it was the last reference to those dead names.
- **No collateral:** `Type` is still used by `test_storage_pylist_roundtrip`, so the removal doesn't orphan the import.
- **CI is fully green** — `BuildAndTest` (Linux + macOS) + pytest + all wheels + codecov all pass on the current commit; base is up-to-date with master.

No findings — a strict improvement to the test suite.

pcchen (approved):
Approving. Correct, non-vacuous `@=` identity test (snapshots `Dot(a,b)` before the in-place op, asserts identity + zero-norm correctness, and per the PR notes fails on the pre-#986 module), reasonable removal of the one-time shadow-name migration guard, no orphaned imports, and full test CI (BuildAndTest Linux+macOS, pytest, wheels, codecov) green on the current commit. Closes both #986 review comments.

PR #1006, opened 2026-07-07
docs: record 2026-07 API decisions and archive the phase execution plans

yingjerkao:
Adds `docs/dev/api-decisions.md` — a contributor-facing record of the API rulings made on 2026-07-06 (resolving #934/#753/#675, #676, #335/#336/#381 + #421/#422, #293) and the issue-thread consensus positions implemented this cycle (#846/#841, #724, #836), each linked to the implementing PRs (#997, #998, #1000, #1001, #1005).

Also archives the three phase execution plans under `docs/dev/plans/` as historical records, clearly marked as such (they contain machine-specific paths from the agent sessions that executed them and are not build instructions).

Docs-only; no code changes. Independent of all open chains — merge anytime. If any linked PR's scope changes materially before merging, this file should be updated in that PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1010, opened 2026-07-07
refactor(Symmetry): make Symmetry a plain value type over std::variant

### LINKED ISSUE #842: Refactor Symmetry to a plain value type (drop PIMPL + virtual dispatch)
## Summary

`Symmetry` is, semantically, a small tagged value: a `SymmetryType` enum plus an integer `n`. It carries no per-instance storage beyond that — every behaviour difference between U1 / Zn / fermionParity / fermionNumber is determined entirely by `stype_id` (and `n` for `Zn`).

Today it is implemented as a PIMPL with virtual dispatch, heap allocation, and intrusive ref counting:

- `include/Symmetry.hpp:64-97` — `Symmetry_base : public intrusive_ptr_base<Symmetry_base>` holds just two `int` fields (`stype_id`, `n`) plus a vtable of `combine_rule_`, `reverse_rule_`, `check_qnum`, `get_fermion_parity`, etc.
- `include/Symmetry.hpp:101-200` — four trivial subclasses (`U1Symmetry`, `ZnSymmetry`, `FermionParitySymmetry`, `FermionNumberSymmetry`) each add zero data members and override the virtuals.
- `include/Symmetry.hpp:205-538` — the public `Symmetry` class wraps a `boost::intrusive_ptr<Symmetry_base> _impl` and forwards every call through it.

So for what is effectively `struct Symmetry { SymmetryType type; int n; };` we pay:

- a heap allocation on every `Symmetry()` / `Symmetry::U1()` / `Symmetry::Zn(n)` / `clone()`,
- atomic ref-count increments on every copy (intrusive_ptr),
- a virtual call on every `combine_rule_` / `reverse_rule_` / `check_qnum`,
- 8 bytes of payload behind a pointer that is itself 8 bytes plus the control block.

These objects are stored inside every `Bond`, which is itself stored on every `UniTensor`, so symmetric-tensor code allocates and refcounts them constantly.

## Proposal

Refactor `Symmetry` into a plain value type. Sketch:

```cpp
class Symmetry {
 public:
  // Generators stay the same as today
  static Symmetry U1();
  static Symmetry Zn(int n);
  static Symmetry FermionParity();
  static Symmetry FermionNumber();

  SymmetryType stype() const { return stype_id_; }
  int n() const               { return n_; }
  std::string stype_str() const;

  bool check_qnum(cytnx_int64 q) const;
  bool check_qnums(const std::vector<cytnx_int64>& qs) const;

  cytnx_int64 combine_rule(cytnx_int64 a, cytnx_int64 b, bool reverse = false) const;
  void        combine_rule_(cytnx_int64& out, cytnx_int64 a, cytnx_int64 b, bool reverse = false) const;
  std::vector<cytnx_int64> combine_rule(const std::vector<cytnx_int64>& a,
                                        const std::vector<cytnx_int64>& b) const;
  void        combine_rule_(std::vector<cytnx_int64>& out,
                            const std::vector<cytnx_int64>& a,
                            const std::vector<cytnx_int64>& b) const;

  cytnx_int64 reverse_rule(cytnx_int64 q) const;
  void        reverse_rule_(cytnx_int64& out, cytnx_int64 q) const;

  fermionParity get_fermion_parity(cytnx_int64 q) const;
  bool          is_fermionic() const;

  Symmetry clone() const { return *this; }  // value semantics

  bool operator==(const Symmetry& rhs) const = default;
  bool operator!=(const Symmetry& rhs) const = default;

 private:
  SymmetryType stype_id_ = SymmetryType::Void;
  int          n_        = 0;
};
```

Each member function becomes a `switch (stype_id_)` (the same dispatch the per-subclass virtuals were doing manually). The implementation stays in `src/Symmetry.cpp` so the header doesn't grow.

## Why this is safe to do

- `Symmetry_base` has no per-instance data beyond `stype_id` and `n`. The subclasses add nothing. There is no inheritance state to preserve.
- The number of symmetry kinds is small and closed (the `SymmetryType` enum lives in the same header). Adding a new symmetry kind is rare and remains a one-place change (add an enum value + a case in each `switch`), no different from today (add a subclass + register it in `Symmetry::Init`).
- The public API (`Symmetry::U1()`, `Symmetry::Zn(n)`, `combine_rule`, `reverse_rule`, `check_qnum`, `clone()`, `==`, `!=`, `Save` / `Load`, `print_info`, `stype()`, `n()`, `stype_str()`, `is_fermionic`, `get_fermion_parity`) is all preserved by signature, so callers don't change.
- `clone()` becomes a no-op copy, which is what callers actually wanted — they always want a value, never a shared object that mutating one would mutate the other.

## Caveats

- `Symmetry::_impl` and `Symmetry_base` are documented as `@cond` (internal), but they're still part of the compiled ABI of the shared library. This is an ABI-breaking change for anyone linking against `libcytnx.so`. Coordinate with a minor/major version bump.
- The `_Save` / `_Load` binary format must keep emitting / consuming `(stype_id, n)` in the same layout so existing `.cysym` files keep working.
- The Python bindings (pybind11) reference `Symmetry`'s public surface; if any of them poke at `_impl` directly they need to be updated. Quick audit of the bindings is part of the work.

## Suggested work

- [ ] Move `combine_rule` / `reverse_rule` / `check_qnum` / `get_fermion_parity` implementations from the four subclasses into a single `switch`-based implementation in `src/Symmetry.cpp`.
- [ ] Replace `boost::intrusive_ptr<Symmetry_base> _impl` with the two value fields.
- [ ] Delete `Symmetry_base`, `U1Symmetry`, `ZnSymmetry`, `FermionParitySymmetry`, `FermionNumberSymmetry`.
- [ ] Verify `_Save` / `_Load` round-trips byte-for-byte with the prior format.
- [ ] Run the existing symmetry tests; add a microbench (Bond construction / contraction of a small symmetric tensor) to confirm the alloc/refcount overhead is gone.
- [ ] Audit Python bindings and any code that touches `Symmetry::_impl`.

yingjerkao:
## Problem

Fixes #842. `Symmetry` was a `boost::intrusive_ptr<Symmetry_base>` PIMPL over four virtual subclasses — reference semantics, virtual dispatch, and `clone()` boilerplate for what is a small immutable value.

## What changed

`Symmetry` now holds a by-value `std::variant<U1Symmetry, ZnSymmetry, FermionParitySymmetry, FermionNumberSymmetry>`; every method dispatches via `std::visit`. The generic visit lambdas are the completeness contract: adding a fifth kind fails compilation until all rule methods exist.

**Preserved exactly:** the full public API (verified signature-by-signature — factories, `combine_rule(_)` including out-param forms, `stype()/stype_str()`, the legacy `int& n() const` accessor now backed by a `mutable` member, comparison = stype+n, all error messages) and the Save/Load byte format — pinned by binary fixtures generated at the base commit and byte-compared in tests, both directions, including a mixed-symmetry Bond. *(One deliberate exception, per review: the non-mutating rule methods — `check_qnum`, `check_qnums`, the vector `combine_rule`, both `combine_rule_` overloads, `reverse_rule_` — are now `const`-qualified. This only relaxes callability; no signature is otherwise changed.)*

**Deliberate notes:** `is(Symmetry, Symmetry)` now means address identity (copies are no longer `is` their source); likewise, writing through the legacy `int &n() const` reference no longer propagates to copies (each Symmetry owns its value) — the honest semantics for a value type; the sole in-tree observer (a doc example) is updated. Of the two pre-existing bugs found during the work, `FermionParitySymmetry::check_qnums` (always false for non-empty input) has since been fixed on master by #1012 and this branch ports that fix into the variant implementation; the `NEED_GEN_COMMON_DATA=1` exit crash remains preserved and tracked separately.

## Testing

20 new gtests (byte-compat fixtures, API pinning, quirk pin, bad-magic Load rejection, signature `static_assert`). Full suite: 1106 passed / 11 skipped / only the 4 pre-existing failures addressed by #977. pytest 57 passed; the pybind surface diff is empty.


pcchen:
## Code Review

Converts `Symmetry` from a `boost::intrusive_ptr<Symmetry_base>` PIMPL (four virtual subclasses + `clone()` boilerplate) to a by-value `std::variant<U1/Zn/FermionParity/FermionNumber>` with `std::visit` dispatch. A clean, well-motivated modernization of a small immutable value.

### Correctness — verified
- **Variant + `std::visit` design is clean.** Every method dispatches via a generic `std::visit([](const auto& kind){...}, _impl)`; the generic lambdas are a real completeness contract — adding a 5th alternative fails to compile until it implements every rule method. ✅
- **Legacy `n` values preserved** with correct ctors: `U1Symmetry() : n(1)`, `ZnSymmetry(n)` (validates `n>1`), `fPar = -2`, `fNum = -1` — matching the old impl (documented in-code). The `Symmetry(stype,n)` ctor discards the `n` arg for U1, but that's fine since `U1Symmetry()` sets `n(1)`. ✅
- **`==`/`!=` = `stype()+n()`** — matches old semantics and correctly distinguishes `Zn(3)` vs `Zn(5)` by `n` (their `stype()` is both `Z`). ✅
- **Save/Load byte format** writes/reads `777 magic (uint) + stype_id (int) + n (int)` — crucially it writes `stype()` (the `SymmetryType` value), **not** the variant index, so old files stay loadable and new saves are byte-identical. `_Load` validates the 777 magic and routes through `Init`, which errors on an invalid `stype` — so a corrupt file errors rather than silently mis-loading. ✅
- **`clone() const { return *this; }`** — correct for a value type. ✅
- **CI:** `BuildAndTest` (C++ gtests) is actually running on this commit, and the test methodology is strong — base-generated `.cysym`/`.cybd` byte fixtures compared both directions (incl. a mixed-symmetry Bond), API `static_assert`s, quirk pins, and bad-magic rejection.

### Findings (low / notes — none blocking)

**1. A second reference->value semantic change, only partly documented.** The PR flags that `is()` now means address identity (copies aren't `is` their source). The same shift applies to `n()`: since `int& n() const` returns a reference into a now-per-instance `mutable` member, `sym.n() = k` no longer propagates to copies (the old shared-impl did). Arguably safer, but it's a second observable behavior change from the reference->value move — worth adding to the "deliberate notes" alongside `is()`.

**2. `int& n() const` backed by `mutable int n` is a preserved legacy wart** — a const accessor handing out a writable reference (const-mutation + thread-safety hazard). Fine for the exact-API-preservation goal and clearly documented; a good follow-up would be to make `n()` return `int` by value, since the mutable-ref-from-const pattern has no legitimate caller.

**3. Stale comment `// default is U1Symmetry`.** Both old and new actually *throw* on default construction (`Symmetry()` -> `Init(-1)` -> `[ERROR] invalid symmetry type`). Preserved behavior (verified against master), and no in-tree code default-constructs `Symmetry`, so it's harmless — but the comment is misleading and was carried over verbatim; worth correcting while here.

**4. Preserved pre-existing bugs** (`FermionParitySymmetry::check_qnums` always false; `NEED_GEN_COMMON_DATA=1` exit crash) — honestly disclosed, pinned bit-for-bit by tests, and tracked separately. Correct call not to fix them in a "preserve exactly" refactor.

### Verdict
High-quality, faithful refactor: the variant model is cleaner and cheaper than the virtual PIMPL, the public API and Save/Load byte format are preserved (structurally verified here; pinned by base-generated fixtures in the suite), and the semantic changes are deliberate and mostly documented. Findings are notes/nits, not defects. **Approve-with-nits** once `BuildAndTest` goes green (it's running now).

pcchen:
### Correction + verification follow-up to my review above

**Retracting finding #3 — it was my error.** I claimed `Symmetry()` throws and the `// default is U1Symmetry` comment is stale. That's wrong: `enum SymmetryType : int { Void=-99, U=-1, Z=0, fPar=-2, fNum=-3 }`, so `SymmetryType::U == -1` — which is exactly the constructor's default `stype = -1`. `Symmetry()` -> `Init(-1)` -> `U1Symmetry()`, i.e. it correctly default-constructs to U1, and the comment is accurate. I misread `-1` as an invalid stype. Apologies — please disregard finding #3.

**Byte-compat and API preservation now positively verified** (I did a deeper pass on the fixtures):
- **Save/Load is genuinely byte-preserving, pinned against real old-format bytes.** The `.cysym` fixtures decode exactly to `[777][stype][n]` (e.g. `sym_fPar` = `777,-2,-2`; `sym_fNum` = `777,-3,-1`), the four Symmetry blocks are embedded byte-identically inside `bond_mixed_syms.cybd`, and the generator's cited base commit `d6dcd160` is confirmed the last intrusive-ptr commit and an ancestor of master. Tests load the old files **and** re-save freshly-built objects and `EXPECT_EQ` the bytes — a true old-vs-new pin, plus bad-magic rejection. The variant index is never serialized; only `stype()` is.
- **Public API is signature-for-signature identical** (factories, `combine_rule`/`combine_rule_` incl. out-param forms, `reverse_rule`(_), `stype`/`stype_str`/`n`, `check_qnum(s)`, `Save`/`Load` overloads, `_Save`/`_Load`, `clone() const`, `get_fermion_parity`, `is_fermionic`, `print_info`, `==`/`!=`); `pybind/symmetry_py.cpp` is unchanged; no public method dropped. `Symmetry_base`/`_impl` are `@cond`-internal with no external references.
- **Finding #1 impact is negligible in practice:** no `.n() = ` mutation exists anywhere in `src`/`include`/`pybind`, and pybind exposes `n()` as an int copy to Python, so the aliased-mutation-through-`n()` path is never exercised. It remains worth a one-line note next to the `is()` note, but it's cosmetic.

Net: the two nits that stand are #1 (document the `n()` value-semantics note) and #2 (`int& n() const` wart, follow-up). Verdict unchanged and if anything firmer — **approve-with-nits** once `BuildAndTest` is green.

pcchen:
### On @gemini-code-assist's 6 `const`-correctness suggestions

They're all one consistent theme (mark `check_qnum`, `check_qnums`, the `combine_rule` vector overload, both `combine_rule_` overloads, and `reverse_rule_` as `const`), and they're **sound and safe to apply**:

- **They compile** — the per-kind struct methods (`check_qnum`/`combine_rule_`/`reverse_rule_`) are already `const` (Symmetry.hpp:83–89 etc.), so visiting a `const kind_variant` and calling them is valid. These `Symmetry`-level methods genuinely don't mutate `*this` (the visit lambdas take `const auto& kind`), so `const` is the correct qualifier.
- **They don't break the test pins** — the only `static_assert` (Symmetry_test.cpp:20) pins `n()`'s `int&` return, not the const-ness of these six.
- **They don't break pybind** — `check_qnum`/`check_qnums` bind via `&Symmetry::check_qnum` (pybind handles const-member pointers); `combine_rule`/`reverse_rule` bind via lambdas calling on a non-const `self`. `Symmetry` is no longer polymorphic, so there are no overrides to worry about.
- **Source-compatible** — adding `const` only *relaxes* callability (const objects can now call them); it breaks no existing caller.

**One caveat:** these six were non-const in the *old* code too (a legacy accident — the old virtual methods were non-const), so applying them is a small, deliberate deviation from this PR's "signature-by-signature identical" goal. Since it's a strict, non-breaking improvement, I'd take it — either here or as a quick follow-up. If applied:
1. Do all six together (they're internally consistent).
2. Confirm `get_fermion_parity` is already `const` (it's not in Gemini's list) so the read-only surface is uniformly const.
3. Update the "signature-by-signature identical" note to "…plus const-correctness on the non-mutating rule methods."

Not a correctness bug either way, so it doesn't change my approve-with-nits verdict — just a worthwhile cleanup.

pcchen:
## Status update: #1012 has merged — three rebase items now apply here

**#1012** (`FermionParitySymmetry::check_qnums` rejected all qnums) merged to master as `8f69f059`. Since this PR deliberately preserved that bug bit-for-bit as a pinned quirk, its rebase now needs three things — one of which is a correctness trap:

1. **Port the fix into the rewritten variant implementation.** This PR rewrites `src/Symmetry.cpp` wholesale, so a take-mine conflict resolution would silently **re-introduce the bug**. The variant kind-struct's `check_qnums` should delegate per-element to `check_qnum` (the structural fix #1012 landed), not carry the old always-false comparison forward.
2. **Flip the preserved-quirk pin.** The test that pins `check_qnums` returning false for non-empty input now contradicts master and will fail post-rebase — it should assert the fixed behavior (`{0,1}` → true; `{2}`/`{-1}`/`{0,1,2}` → false).
3. **Merge the two `tests/Symmetry_test.cpp` files** (both this PR and #1012 created it — both-added conflict) and dedupe the `tests/CMakeLists.txt` entry (which also gained the include-order canary from #981 on master).

Standing from earlier review, unchanged: approve-with-nits verdict, Gemini's six `const`-correctness suggestions (endorsed — apply together, check `get_fermion_parity` too), the `n()` value-semantics doc note, and the byte-compat fixtures remain valid (the Save/Load format is untouched by #1012).

pcchen:
## Consolidated checklist to land this PR

Everything below is gathered from the review, the const-correctness thread, and the #1012 merge — nothing else is outstanding.

### A. Rebase onto current master (currently conflicting) — three specific resolutions
- [x] **Port #1012's fix into the rewritten variant `check_qnums`** (delegate per-element to `check_qnum`). ⚠️ *Correctness trap:* a take-mine resolution of the wholesale `Symmetry.cpp` rewrite silently re-introduces the bug #1012 fixed.
- [x] **Flip the preserved-quirk pin** — the test asserting `check_qnums` returns false for non-empty input now contradicts master; assert the fixed behavior instead (`{0,1}` → true; `{2}` / `{-1}` / `{0,1,2}` → false).
- [x] **Merge the two `tests/Symmetry_test.cpp`** (this PR's suite + #1012's two tests; both-added conflict) and **dedupe `tests/CMakeLists.txt`** (which also gained #981's include-order canary on master).

### B. Const-correctness (the six suggestions, endorsed)
- [x] Add `const` to `check_qnum`, `check_qnums`, `combine_rule` (vector overload), both `combine_rule_` overloads, and `reverse_rule_` — all six together.
- [x] Verify `get_fermion_parity` is also `const` so the read-only surface is uniformly const.
- [x] Update the "signature-by-signature identical" note to "…plus const-correctness on the non-mutating rule methods."

### C. Documentation
- [x] One-line note for the *second* reference→value semantic change: `sym.n() = k` no longer propagates to copies — alongside the existing `is()` address-identity note.
- [x] Changelog/release-note entry for the `is()` + `n()` semantic changes.

### D. After rebase
- [ ] Fresh CI green — `BuildAndTest` running the merged Symmetry tests and the byte-compat fixture suite (fixtures remain valid; #1012 didn't touch the Save/Load format).

**Follow-up (not this PR):** retire the `int& n() const` mutable-ref wart (return `int` by value) once exact-API-preservation is no longer the constraint.

Everything else was already verified in review (byte-compat pinned against real old-format fixtures, full API surface preserved, `is()` migration). Once A–D are checked: approve + merge. Independent of the #1011 Scalar stack (no shared files), so landing order between them is free.

pcchen:
**Checklist item A implemented** — rebased onto master (head `bfa0e84b`, now `MERGEABLE`); the branch is `master + 1 commit`, 13 files:

- **Fix ported:** the variant `FermionParitySymmetry::check_qnums` now delegates per-element to `check_qnum` (with a comment citing #1012), replacing the quirk-preserving `qnums[i] < this->n` comparison that a take-mine resolution would have carried forward.
- **Quirk pin replaced:** `FermionParityCheckQnumsLegacyQuirk` is removed; master's two #1012 tests (`FermionParityCheckQnum`, `FermionParityCheckQnums`) are carried over verbatim for coverage continuity, plus a small `FermionParityCheckQnumsEmptyIsVacuouslyTrue` preserving the quirk test's one unique assertion. Verified no other test in the suite pinned the old behavior (`CheckQnumRanges` only exercises the singular form for fPar).
- **CMakeLists deduped:** the auto-merge had produced *two* `Symmetry_test.cpp` entries — collapsed to one; #981's canary entry preserved.

Sanity: no conflict markers, brace balances clean, byte-compat fixtures untouched. Fresh CI is running on `bfa0e84b`. **Items B (const-correctness), C (docs/changelog), and D (CI green) remain with the author.**

pcchen:
**Checklist item B implemented** (commit `f60ff7de`):

- The six non-mutating rule methods on the `Symmetry` wrapper are now `const`: `check_qnum`, `check_qnums`, the vector `combine_rule`, both `combine_rule_` out-param overloads, and `reverse_rule_`. All six dispatch via `std::visit` with `const auto &kind` onto kind-struct methods that were already `const`, so visiting a const `_impl` is well-formed.
- Verified `get_fermion_parity` and `is_fermionic` were already `const` — the read-only rule surface is now uniformly callable on `const Symmetry` objects.
- Verified the two live pybind bindings (`&Symmetry::check_qnum` / `check_qnums`) accept the const member pointers unchanged; no member-pointer caller elsewhere.
- PR description's "Preserved exactly" note updated with the deliberate const-qualification exception.

**Remaining: C (changelog entry for the `is()`/`n()` semantic changes) and D (fresh CI green — now running on `f60ff7de`).**

pcchen:
**Checklist item C implemented** (commit `15b2d99e`):

- **In-code note at `n()`**: documents the second reference→value change — the returned reference now points into *this* Symmetry's own variant, so `Symmetry b = a; a.n() = k;` no longer changes `b` (it did under the shared intrusive-ptr impl) — placed alongside the existing legacy-API comment.
- **Release-note entry**: the repo has no CHANGELOG file, so per the #1006 convention the entry went into `docs/dev/api-decisions.md` ("Consensus positions" list): Symmetry as a plain value type, byte-format preservation, and both observable semantic changes (`is()` address identity + `n()` copy independence), referencing this PR.
- **PR description updated**: the `n()` note added to "Deliberate notes", and the now-stale "two pre-existing bugs preserved bit-for-bit" claim corrected (the `check_qnums` quirk was fixed on master by #1012 and this branch ports that fix; the `NEED_GEN_COMMON_DATA=1` exit crash remains the one preserved item).

**Checklist status: A ✅ B ✅ C ✅ — only D remains (fresh CI green, running on `15b2d99e`).** Once green, this should be ready to approve and merge.

pcchen (approved):
Approving. The value-type refactor was verified in review (index-aligned variant pinned by static_asserts; Save/Load byte format pinned against real old-format fixtures generated at the base commit, byte-compared both directions; public API preserved signature-for-signature — with the one deliberate, documented exception of const-qualifying the non-mutating rule methods). The full review checklist is complete:

- **A** — rebased onto master with the #1012 fix ported into the variant `check_qnums` (avoiding silent bug re-introduction), the preserved-quirk pin flipped to the fixed expectations, the two `Symmetry_test.cpp` suites merged, and the CMakeLists entry deduped.
- **B** — the six const-correctness suggestions applied (+ `get_fermion_parity`/`is_fermionic` verified already const); pybind member-pointer bindings unaffected.
- **C** — the `n()` copy-independence change documented at the accessor and recorded with the `is()` change in `docs/dev/api-decisions.md`; PR description updated (including correcting the now-stale preserved-quirk claim).
- **D** — CI fully green on `15b2d99e` (16 pass; the single ubuntu failure was a conda env-setup infra flake, passed on rerun).

Note for the record: the checklist commits (A–C) were implemented by Claude Code on this branch per the review threads; the core refactor is @yingjerkao's.

PR #1014, opened 2026-07-07
fix(pybind): keep-set ordering for Scalar constructors, add __bool__

yingjerkao:
Stacked on #1011 (`refactor/scalar-variant`) — the fixed behavior is pre-existing on master `d6dcd160`, but the regression tests that pinned the old behavior live in that branch's `pytests/scalar_variant_test.py`, so this lands on top of it and flips them in the same commit.

## Problem

`scalar_py.cpp` registered the plain C++ integral constructors ahead of the `py::numpy_scalar<T>` `__init__` overloads. pybind11's plain arithmetic casters accept any `__index__`-capable object even in the no-convert pass — the exact trap documented under "KEEP-SET ORDERING" in `pybind/pyint_dispatch.hpp` (PR #991). Observed on the previous overload set:

| input | before | after |
|---|---|---|
| `np.uint64/int32/uint32/int16/uint16` (in range) | `Int64` | own dtype |
| `np.uint64(2**64-1)` | `Uint64` (int64 caster overflows, falls through) | `Uint64` (unchanged) |
| `Scalar(5)` / `Scalar(2**63)` | `Uint64` / `Uint64` | `Int64` / `Uint64` (magnitude dispatch) |
| `Scalar(2**64)` | **silently `ComplexDouble`** (complex ctor convert pass) | clean error |
| `Scalar(True)` | `Uint64` | `Bool` |
| `bool(Scalar(np.int32(0)))` | `True` (no `__bool__`, default truthiness) | `False` |
| Bool-storage `storage[i] = Scalar(np.bool_(True))` | `RuntimeError: Unable to cast` | works |

## Fix

- Rebind the constructor overload set in the canonical keep-set order: numpy scalars first, then Python `bool` (must precede `py::int_`, which accepts `bool` as an `int` subclass), then a single `py::int_` overload dispatching on magnitude via `dispatch_pyint`, then `double`/`complex128` (absorbing their `np.float64`/`np.complex128` subclasses). Per the discipline, the per-C++-dtype integral constructors and the redundant `numpy_scalar<double>`/`numpy_scalar<complex<double>>` overloads are dropped.
- Add `Scalar.__bool__` with numpy-consistent semantics (nonzero → `True`; either component for complex). This also fills the `nb_bool` slot pybind11's bool caster requires, unblocking Bool-storage round-trips through Scalar.

## Tests

TDD: tests flipped/added first and confirmed failing (21 red) on the unfixed build, then green after the reorder. `pytests/scalar_variant_test.py`: the former `*_narrower_ints_preexisting_quirk` test now asserts correct dtypes inside `test_numpy_scalar_constructor_paths` (all 11 dtypes); new coverage for uint64-above-int64-max value preservation, plain-int magnitude dispatch + out-of-range error, bool/float/complex constructor dtypes, `__bool__` truthiness across all dtypes, and the Bool row in the Storage round-trip.

Full pytest suite against the dev build: **117 passed, 0 failed**.

### DIFF
--- a/include/Type.hpp
+++ b/include/Type.hpp
@@ -380,6 +380,18 @@ namespace cytnx {
       return ComplexDouble;
     }
 
+    // The dtype linalg::Norm() produces for an input of the given dtype: the
+    // real counterpart for floating/complex dtypes, Double for integer/bool
+    // inputs (which Norm computes in double precision). This is the single
+    // home of that policy -- src/linalg/Norm.cpp and callers pre-sizing
+    // norm accumulators (e.g. UniTensor::normalize_) both use it, so they
+    // cannot drift apart.
+    static constexpr unsigned int norm_result_dtype(unsigned int type_id) {
+      check_type(type_id);
+      if (is_float(type_id)) return to_real(type_id);
+      return Double;
+    }
+
     // Find a common type for typeL and typeR
     static constexpr unsigned int type_promote(unsigned int typeL, unsigned int typeR) {
       if (typeL == Void || typeR == Void) return Void;
--- a/include/backend/Scalar.hpp
+++ b/include/backend/Scalar.hpp
@@ -5,2516 +5,69 @@
   #include "Type.hpp"
   #include "cytnx_error.hpp"
   #include "intrusive_ptr_base.hpp"
-  #include <vector>
-  #include <initializer_list>
-  #include <string>
-  #include <iostream>
-  #include <cmath>
+  #include <ostream>
   #include <type_traits>
-  #include <limits>
+  #include <variant>
 namespace cytnx {
 
   ///@cond
   class Storage_base;
-  class Tensor_base;
 
   // The following two declarations are necessary for ADL.
   void intrusive_ptr_add_ref(Storage_base *);
   void intrusive_ptr_release(Storage_base *);
-
-  // real implementation
-  class Scalar_base {
-   private:
-   public:
-    int _dtype;
-
-    // Scalar_base(const Scalar_base &rhs);
-    // Scalar_base& operator=(const Scalar_base &rhs);
-    Scalar_base() : _dtype(Type.Void){};
-
-    virtual cytnx_float to_cytnx_float() const {
-      cytnx_error_msg(true, "[ERROR] Void Type Scalar cannot cast to anytype!!%s", "\n");
-      return 0;
-    };
-    virtual cytnx_double to_cytnx_double() const {
-      cytnx_error_msg(true, "[ERROR] Void Type Scalar cannot cast to anytype!!%s", "\n");
-      return 0;
-    };
-    virtual cytnx_complex64 to_cytnx_complex64() const {
-      cytnx_error_msg(true, "[ERROR] Void Type Scalar cannot cast to anytype!!%s", "\n");
-      return cytnx_complex64(0, 0);
-    };
-    virtual cytnx_complex128 to_cytnx_complex128() const {
-      cytnx_error_msg(true, "[ERROR] Void Type Scalar cannot cast to anytype!!%s", "\n");
-      return cytnx_complex128(0, 0);
-    };
-    virtual cytnx_int64 to_cytnx_int64() const {
-      cytnx_error_msg(true, "[ERROR] Void Type Scalar cannot cast to anytype!!%s", "\n");
-      return 0;
-    };
-    virtual cytnx_uint64 to_cytnx_uint64() const {
-      cytnx_error_msg(true, "[ERROR] Void Type Scalar cannot cast to anytype!!%s", "\n");
-      return 0;
-    };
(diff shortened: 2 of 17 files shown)


pcchen:
## Code Review

Applies the #991 keep-set discipline to `Scalar`'s Python constructors and adds `__bool__` — closing out exactly the two binding quirks #1011 pinned (numpy-int ctor collapse, missing `nb_bool`). **LGTM** (stacked on #1011).

### Correctness — verified
- **Final constructor order is exactly the canonical keep-set** (confirmed on the branch tip): `numpy_scalar<float>`, `<complex64>` → all 7 numpy int/bool scalars → Python `bool` → `py::int_` magnitude dispatch → `double` → `complex128`. Each ordering constraint checks out:
  - `np.bool_` is *not* a Python-`bool` subclass, so it lands on `numpy_scalar<bool>`; Python `True` lands on `cytnx_bool`, which **must** precede `py::int_` (bool subclasses int) — and does. ✅
  - `py::int_` + `dispatch_pyint` (reusing #991's helper — good consolidation): fits-int64 → `Int64`, else `Uint64`, else a **clean error**. Since a *selected* overload that throws doesn't fall through, the silent `Scalar(2**64)` → ComplexDouble path is dead. ✅
  - Dropping `numpy_scalar<double>`/`<complex<double>>` is correct — `np.float64`/`np.complex128` subclass Python `float`/`complex` and are absorbed by the trailing `cytnx_double`/`cytnx_complex128` ctors. Dropping the 9 plain integral ctors is safe: Python ints all route via `py::int_`, and C++ callers are unaffected (pybind-only file). ✅
- **`__bool__`** is numpy-consistent (complex → either component nonzero; branches before the real cast, which would throw on complex). Filling the `nb_bool` slot also unblocks Bool-storage `__setitem__` round-trips — pinned by the new Bool row in the Storage test. ✅
- **Tests:** TDD (21 red on the unfixed build), the former `*_preexisting_quirk` test flipped into positive dtype assertions across all 11 numpy scalar types, uint64-above-int64-max value preservation, magnitude dispatch + out-of-range error, `__bool__` truthiness across dtypes, and the `astype()` workaround correctly retired from the signed/unsigned test. CI green against the branch base. ✅

### Notes (non-blocking)
1. **Stacking:** base is `refactor/scalar-variant` (#1011); the `CLEAN` merge status is vs. that branch, not master. Merge #1011 first, let this retarget, and give it one fresh master-based CI run before landing.
2. **User-visible dtype changes** (`Scalar(5)`: Uint64→Int64; `Scalar(True)`: Uint64→Bool; `Scalar(2**64)`: silent ComplexDouble→error; `bool(Scalar(np.int32(0)))`: True→False) — all bugfix-flavored and documented in the PR's table, but they belong in the same changelog entry as #1011's breaking changes.
3. **Tiny unpinned edge:** `bool(Scalar())` (Void/monostate) will presumably now throw from the double cast rather than return a truthiness value — worth one pin or a comment; not a blocker.

Clean, disciplined fix. **LGTM.**

pcchen:
## Status update — base has moved; what the next rebase needs to handle

The review verdict above (**LGTM**) stands, but three things changed underneath this PR since it was written:

1. **#1011 was rebased onto current master** (its branch was force-pushed; head now `ea426983`+). This branch still contains the pre-rebase #1011 commits, so it needs the cherry-pick treatment: rebuild as `refactor/scalar-variant`-head + this PR's own commit.
2. **#1011's follow-up commit flipped the in-place semantics to promote** and rewrote the in-place tests in `pytests/scalar_variant_test.py` — the same file this PR edits. Expect conflicts there; the reconciliation is: keep #1011's promote-form tests, apply this PR's constructor-dtype corrections on top (e.g. the signed/unsigned test's direct `np.uint64(...)` construction stays valid — it's the raise-expectation lines that changed shape).
3. **#1011's checklist item A will replace the 11 C++ value ctors with one `explicit` constrained ctor** (per @ianmccul's ruling in that thread). This PR's pybind keep-set should be mechanically unaffected (bindings construct explicitly; `dispatch_pyint`'s `return Scalar(v)` is fine with `explicit`), but the keep-set deserves a **recheck after A lands** — in particular that the trailing `cytnx_double`/`cytnx_complex128` `py::init<>` overloads still resolve against the constrained template.

**Also connected:** #1011's CI is currently red on exactly the ctor-collapse quirk this PR fixes (`test_inplace_same_dtype_keeps_dtype` constructs via `np.int32` → Int64). #1011 will carry an `astype()` workaround until this lands — when this PR rebases, it should **remove that workaround** as part of flipping the quirk coverage (one more reason to land this promptly after #1011).

Standing from the review, unchanged: the Void `bool(Scalar())` edge worth one pin; the user-visible dtype corrections belong in the release-notes effort now tracked in **#1018**; merge order remains #1011 → this, with one fresh master-based CI run before landing.

yingjerkao:
superseded by comment in #1016 to drop Scalar python bindings.

Issue #1013, opened 2026-07-07
Typed Storage dispatch: GPU conversion


yingjerkao:
## Context

#941 ("Its time: Typed Storage Dispatch") specifies converting Cytnx's
Storage/arithmetic dispatch from `Storage_base` + `void*` + runtime `dtype()`
casts to typed `intrusive_ptr<StorageImplementation<T>>` dispatch via
`std::visit`.

Per the maintainer's Phase-3 decision record (ruling 3, "GPU scope:
CPU-first"), the typed-storage foundation and the first converted operation
family (Add/Sub/Mul/Div/Cpr/Mod) implemented on branch
`refactor/typed-storage-foundation` are **CPU-only**. GPU arithmetic
continues to dispatch through the legacy `linalg_internal::lii.cuAri_ii`
function-pointer table, reached from the `#ifdef UNI_GPU` branches of:

- `src/linalg/iAdd.cpp`
- `src/linalg/iSub.cpp`
- `src/linalg/iMul.cpp`
- `src/linalg/iDiv.cpp`
- `src/linalg/Cpr.cpp`
- `src/linalg/Mod.cpp`

and implemented in `src/backend/linalg_internal_gpu/cuArithmetic_internal.cu`
/ `.hpp`, registered in `src/backend/linalg_internal_interface.cpp`.

This mirrors the precedent set by #988 (GPU scalar in-place op issues) --
GPU-specific correctness/design work needs a CUDA-equipped environment to
verify, which was not available in the session that did the CPU conversion.

## Scope of this tracking issue

Convert the GPU arithmetic paths to the same typed-storage design used for
CPU in `refactor/typed-storage-foundation`:

- Templated `intrusive_ptr<StorageImplementation<TO>>` kernel signatures
  (mirroring the CPU `AddInternalImpl<TO,TL,TR>` / `DivInternalImpl<TO,TL,TR>`
  / etc. shape) instead of `intrusive_ptr<Storage_base>` + `reinterpret_cast`.
- Replace `linalg_internal::lii.cuAri_ii[dtype][dtype]` table dispatch with
  `std::visit` over `Storage::as_storage_variant()`, consistent with the CPU
  dispatch wrapper in `src/linalg/iArithmetic_visit.hpp` and the six op files
  above.
- Per #941's discussion of `cuDiv_dispatch`/`cuArithmetic_internal`: drop the
[...]

Issue #673, opened 2025-09-15
Segmentation fault when initializing Storage in C++


haydengemeinhardt:
I ran into a "heisenbug" when using the C++ API. When I initialize a Storage object, I get a segmentation fault, but if I e.g. cout it before ending the program, then it works.

Minimal code to reproduce segmentation fault:
```
#include "cytnx.hpp"
#include <iostream>

int main() {
    auto A = cytnx::Storage(10);
    return 0;
}
```

While this works:
```
#include "cytnx.hpp"
#include <iostream>

int main() {
    auto A = cytnx::Storage(10);
    std::cout << A;
    return 0;
}
```

To give some more context, I ran into this while converting my code from python to c++. I had a vector I wanted to represent as a UniTensor. I was using Storage::from_vector() and then Tensor::from_storage(), then converted that tensor into a UniTensor(). It segfaults after from_vector and from_storage if I only have up to that implemented, but once I implemented the conversion of the tensor into a UniTensor, the segfault went away and I was able to continue coding as normal. I can put together an example of that as well if its helpful.

haydengemeinhardt:
> How did you get cytnx pakage? Which version are you using?

I got the package via conda install following the user guide, so I'm on version 1.0.0 on Ubuntu 24.04

ianmccul:
Analysis from Codex:

# Diagnosis of #673: `Storage_init_interface __SII` is not a reliable initializer

Issue #673 reports a C++ heisenbug:

```cpp
#include "cytnx.hpp"

int main() {
    auto A = cytnx::Storage(10);
    return 0;
}
```

can segfault, while this appears to work:

```cpp
#include "cytnx.hpp"
#include <iostream>

int main() {
    auto A = cytnx::Storage(10);
    std::cout << A;
    return 0;
}
```

The likely cause is the initialization pattern used by `Storage_init_interface`.

In `include/backend/Storage.hpp`, `Storage::Init` is inline and calls:

```cpp
this->_impl = __SII.USIInit[dtype]();
```

The table is declared as a static member:

```cpp
class Storage_init_interface : public Type_class {
 public:
  inline static pStorage_init USIInit[N_Type];
  inline static bool inited = false;

  Storage_init_interface() {
    if (!inited) {
      USIInit[this->Double] = SIInit_d;
      USIInit[this->Float] = SIInit_f;
      USIInit[this->ComplexDouble] = SIInit_cd;
[...]

PR #1015, opened 2026-07-08
refactor(Storage)!: typed storage dispatch foundation + capacity_/numpy ownership (#941)

yingjerkao:
## What this delivers (re-scoped)

Rebased onto current `master` (which now carries the Scalar→`std::variant` refactor via #1011) and **narrowed to the foundation**, in two commits:

1. **Typed storage foundation (#941)** — `StorageVariant` generated from `Type_list` (single source of truth, compile-time exhaustiveness), `storage_cast<T>`, `as_storage_variant()`, `storage_as_type_or_replace<T>`, and typed `StorageImplementation<T>::data() -> T*`. Static-asserts pin the variant to the dtype list. Exercised directly by new `Storage_test` cases (no dependency on the linalg conversion).
2. **`capacity_` removed + numpy zero-copy capsule ownership** (#941 rulings 2, 4) — `append`/`resize` are exact realloc, O(n), numpy-like; `Storage.numpy()`/`Tensor.numpy(share_mem=True)` genuinely share via an `intrusive_ptr`-holding capsule (the leaky `release()` is gone).

## What changed vs the original PR

The original #1015 also carried two commits that are now **dropped**:

- **`refactor(Scalar)!: PIMPL → std::variant`** — fully **superseded by #1011** (the same refactor, in its final reviewed form, already on `master`).
- **`fix(linalg)!: convert CPU Add/Sub/Mul/Div/Cpr/Mod to typed dispatch` + true-division** — **deferred to a follow-up PR**. It collides head-on with master's #1026 (rank-zero refactor of all six arithmetic files + the `init_broadcast_binary_output` helper), #1043 (zero-extent), and #984 (type_promote). Correctly landing it means re-implementing the `storage_cast` dispatch **and** the breaking true-division (int/int → Double, #941 §True division) on top of master's now-refactored arithmetic — a numerical change that warrants its own focused, reviewable PR against current master. The original work is preserved at commit `ff30d0090` for that follow-up.

## Conflict-resolution note

During the rebase, master's `StorageTest.InitByPtrRejectsZeroLength` (a safety test added after this PR branched) required `_Init_byptr(ptr, 0)` to always throw; the capacity_-removal commit had weakened that guard to `#ifdef UNI_DEBUG` (a release-build `free()` hazard). Kept the always-on guard.

## Testing

Full CPU `test_main` **1784 tests, 1773 passed / 11 skipped / 0 failed** on the `openblas-cpu` (non-ASan) preset; storage/numpy-ownership/append pytests pass; clang-format v14 clean.

## Follow-ups (tracked separately)

CPU linalg typed-dispatch conversion + true-division (deferred here); GPU conversion (#1013).

### DIFF
--- a/include/backend/Storage.hpp
+++ b/include/backend/Storage.hpp
@@ -38,9 +38,12 @@ namespace cytnx {
     // void Init(const std::initializer_list<unsigned int> &init_shape);
     std::string dtype_str() const;
     std::string device_str() const;
-    virtual const unsigned long long capacity() const {
-      cytnx_error_msg(true, "Not implemented.%s", "");
-    }
+    // capacity()/capacity_ were removed (#941 ruling 2): Storage always
+    // allocates exactly size() elements; append()/resize() reallocate
+    // exactly (documented O(n) per call, numpy-like) instead of the previous
+    // STORAGE_DEFT_SZ-rounded over-allocation. This also guarantees a
+    // buffer's allocated extent always equals its size, which the zero-copy
+    // numpy view (ruling 4) relies on.
     virtual const unsigned long long size() const {
       cytnx_error_msg(true, "Not implemented.%s", "");
     }
@@ -169,19 +172,12 @@ namespace cytnx {
                             const std::vector<std::vector<cytnx_uint64>> &locators,
                             const cytnx_uint64 &Nunit, const bool &is_scalar);
 
-    /**
-     * @brief Drop the ownership of the underlying contiguous memory.
-     *
-     * The caller MUST take the ownership before calling this method. Any operation following this
-     * method causes undefined behavior.
-     *
-     * @return The pointer referencing the underlying storage.
-     * @deprecated This method may be removed without any notification.
-     */
-    [[deprecated("This method is deprecated and may be removed in the future.")]] virtual void *
-      release() noexcept {
-      return nullptr;
-    }
+    // release() was removed (#941 migration step 12): it returned an erased
+    // void*, dropped the allocation from StorageImplementation<T>, and left a
+    // typed storage object in a deliberately invalid state -- incompatible
+    // with trustworthy typed storage invariants. Its only callers (the
+    // .numpy() bindings) now use a py::capsule holding an
+    // intrusive_ptr<Storage_base> for ownership/keepalive instead.
 
     // these is the one that do the work, and customize with Storage_base
     // virtual void Init(const std::vector<unsigned int> &init_shape);
@@ -267,7 +263,7 @@ namespace cytnx {
     static constexpr unsigned int dtype_value = Type_class::cy_typeid_v<DType>;
 
     StorageImplementation()
-        : capacity_(0), size_(0), start_(nullptr), dtype_(Type.cy_typeid(DType())), device_(-1){};
+        : size_(0), start_(nullptr), dtype_(Type.cy_typeid(DType())), device_(-1){};
     void Init(const unsigned long long &len_in, const int &device = -1,
               const bool &init_zero = true);
     void _Init_byptr(void *rawptr, const unsigned long long &len_in, const int &device = -1,
@@ -291,7 +287,6 @@ namespace cytnx {
     boost::intrusive_ptr<Storage_base> real();
     boost::intrusive_ptr<Storage_base> imag();
 
-    const unsigned long long capacity() const override { return capacity_; }
     const unsigned long long size() const override { return size_; }
     void *data() const override { return start_; }
     int dtype() const override { return dtype_; }
@@ -311,23 +306,6 @@ namespace cytnx {
     // as_storage_variant()'s visitor, which is the primary typed-dispatch
     // entry point anyway.
     value_type *data() { return static_cast<value_type *>(start_); }
-
-    /**
-     * @brief Drop the ownership of the underlying contiguous memory.
-     *
-     * The caller MUST take the ownership before calling this method. Any operation following this
-     * method causes undefined behavior.
-     *
-     * @return The pointer referencing the underlying storage.
-     * @deprecated This method may be removed without any notification.
-     */
-    void *release() noexcept override {
-      void *original_start = start_;
-      start_ = nullptr;
-      size_ = 0;
-      capacity_ = 0;
-      return original_start;
-    };
 
     // generators:
     void fill(const cytnx_complex128 &val);
@@ -385,7 +363,6 @@ namespace cytnx {
 
     void *start_;
     unsigned long long size_;
-    unsigned long long capacity_;
     unsigned int dtype_;
     int device_;
   };
@@ -886,26 +863,6 @@ namespace cytnx {
 
     */
     unsigned long long size() const { return this->_impl->size(); }
-
-    /**
-    @brief the capacity in the Storage.
-    @details the capacity is the actual allocated memory in the Storage. The behavior of
-      capacity is similar to std::vector::capacity() in c++.
-    @return [cytnx_uint64]
-
-    */
-    unsigned long long capacity() const { return this->_impl->capacity(); }
-
-    /**
-     * @brief Drop the ownership of the underlying contiguous memory.
-     *
-     * The caller MUST take the ownership before calling this method. Any operation following this
-     * method causes undefined behavior.
-     *
-     * @return The pointer referencing the underlying storage.
-     * @deprecated This method may be removed without any notification.
-     */
-    void *release() noexcept { return this->_impl->release(); }
 
     /**
     @brief print the info of the Storage, including the device, dtype and size.
--- a/include/backend/Storage.hpp
+++ b/include/backend/Storage.hpp
@@ -67,6 +67,19 @@ namespace cytnx {
 
     // `Storage_base` can be instanitiated directly. It's deconstructor calls `data()`, so we cannot
     // throw a runtime error here.
+    //
+    // Legacy erased data accessor (#941). Retained ONLY for consumers not yet
+    // converted to typed dispatch via as_storage_variant()/storage_cast<T>:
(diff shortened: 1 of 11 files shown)


ianmccul:
Wow!

include/backend/Storage.hpp
```diff
@@ -344,7 +353,6 @@ namespace cytnx {
 
     void *start_;
```
ianmccul, line 364: why not `value_type *start_` ?

include/backend/Storage.hpp
ianmccul, line 321: Better implemented via a variant visitor

ianmccul, line 336: Better implemented as a variant visitor

ianmccul, line 350: Better implemented as a variant visitor

ianmccul, line 337: Better if this returns `value_type`.  Higher level operation on `Storage_base` layer can turn it into `Scalar` as needed.

```c++
Scalar Storage::get_item(idx) {
  return visit_storage_value(..., [](auto v) { return Scalar(v); });
}
```

Why not name it `operator[]` ?

ianmccul, line 154: I think its possible to remove this function?  It seems to be dead code?  Even before this PR, `grep -rn _Init_byptr_safe` the only hits are these lines.

ianmccul, line 526: This has to go.  If for no other reason, the `extern __SII` is an order-of-initialization bug #673 and resolving that is a key goal of #941 

Replace with something like
```c++
boost::intrusive_ptr<Storage_base> make_storage(unsigned int dtype) {
  constexpr std::size_t n_non_void = std::variant_size_v<Type_list> - 1;
  return internal::make_storage_impl(dtype, std::make_index_sequence<n_non_void>{});
}
```
which could be implemented using the same `std::index_sequence` pattern already used for `as_storage_variant()`:
```c++
template <std::size_t... Is>
boost::intrusive_ptr<Storage_base>
make_storage_impl(unsigned int dtype, std::index_sequence<Is...>) {
  boost::intrusive_ptr<Storage_base> out;
  bool matched = (... || [&] {
    using T = std::variant_alternative_t<Is + 1, Type_list>; // skip void
    if (dtype == Type_class::cy_typeid_v<T>) {
      out = new StorageImplementation<T>();
      return true;
    }
    return false;
  }());
  cytnx_error_msg(!matched, "[Storage] invalid dtype %u%s", dtype, "\n");
  return out;
}
```

ianmccul, line 1033: also needs to go, same pattern as for `Storage_init_interface`

src/backend/Storage.cpp
ianmccul, line 112: Change to visitor pattern

include/backend/Storage.hpp
ianmccul: Split into two helpers:
```c++
boost::intrusive_ptr<Storage_base>
make_storage(unsigned int dtype);

boost::intrusive_ptr<Storage_base>
make_storage(unsigned int dtype, cytnx_uint64 size, int device = Device.cpu,
             bool init_zero = true);
```
Then for typed C++ sources:

```c++
template <CytnxType T>
boost::intrusive_ptr<StorageImplementation<T>>
make_storage(cytnx_uint64 size, int device = Device.cpu, bool init_zero = true) {
  auto out = boost::intrusive_ptr<StorageImplementation<T>>(new StorageImplementation<T>());
  out->Init(size, device, init_zero);
  return out;
}
```
For from_vector, the cleanest version does not need dtype dispatch at all because T is already known:
```c++
template <CytnxType T>
void _from_vector(const std::vector<T>& vin, int device = Device.cpu) {
  auto impl = make_storage<T>(vin.size(), device, false);

  if constexpr (std::is_same_v<T, cytnx_bool>) {
    impl->_cpy_bool(impl->data(), vin);
  } else {
    std::memcpy(impl->data(), vin.data(), sizeof(T) * vin.size());
  }

  this->_impl = impl;
}
```

ianmccul: even better to remove `->Init()` and pass `size`, `device` and `init_zero` into the constructor

ianmccul:
Once this has been through another iteration to resolve the above issues, I'll work on a PR for #1003

pcchen (changes requested):
## Code Review

**Verdict: Request changes.** The architecture is right, the deletion count (−16.4k) is glorious, and this unblocks #1003/#941 — but there are two verified HIGH runtime defects, plus a needed rebase. Both defects are line-verified against the branch head `d2893495`.

### HIGH-1: every GPU integer÷integer `Div` now throws at runtime
- `src/linalg/Div.cpp` allocates the output as `make_floating_point_dtype(type_promote(...))` → **Double** for int/int — correct, matches this PR's CPU true-division semantics.
- But the new `cuDiv_dispatch_typed<TL,TR>` (`cuDiv_dispatch.cu`) computes `TO = Type_class::type_promote_gpu_t<TL,TR>` — the **plain** promotion (`Type.hpp:471`; Int64 for int/int, no floating-point flooring) — and then hard-errors on mismatch:
  `cytnx_error_msg(out->dtype() != cy_typeid_gpu_v<TO>, "[cuDiv_dispatch] output dtype mismatch...")`.
- Net effect: `(integer tensor on GPU) / (integer tensor or scalar)` throws "output dtype mismatch" on every call.
- **One-line fix:** `using TO = Type_class::make_floating_point_t<Type_class::type_promote_gpu_t<TL, TR>>;` — mirroring the CPU visitor in `Div.cpp`.
- ⚠️ **CI cannot catch this**: there is no GPU CI (#538). It will only surface on a user's machine.

### HIGH-2: numpy zero-copy view → use-after-free on `resize`/`append`
- The capsule design (ruling 4) correctly keeps the `Storage_base` *object* alive as long as the numpy array.
- But `Storage::resize` (`Storage.hpp:793`) calls `_impl->resize()`, which does `free(this->start_); this->start_ = htmp;` **on the same impl** (`StorageImplementation.cpp`). The numpy array captured the old data pointer at view creation → dangling after any size change. Ruling 2 (capacity_ removed; *every* size change reallocates) makes the window maximal — even a single `append` triggers it.
- `pytests/numpy_ownership_test.py`'s docstring claims safety on the premise that "every size-changing Storage operation **replaces the impl** with a fresh buffer" — that premise is false: the impl is mutated in place; only the buffer inside it is replaced.
- Repro:
  ```python
  s = cytnx.Storage(5, cytnx.Type.Double)
  arr = s.numpy()   # zero-copy view
  s.resize(10)      # frees the old buffer in place
  arr[0]            # use-after-free; arr[0] = 1 → heap corruption
  ```
- Fix options, in order of preference: **(a)** make size-changing ops actually swap in a fresh impl, which makes the docstring's safety argument true and matches the evident ruling-4 intent; (b) mark the storage borrowed after `numpy()` and error on size changes; (c) copy instead of view. Whichever is chosen, add a regression test that resizes under a live view.

### Interaction with @ianmccul's comments
His 11 inline comments (visitor-pattern conversions, `make_storage()` factory replacing the `__SII` extern globals — which is the #673 order-of-initialization bug that #941 exists to fix — `value_type *start_`, dead `_Init_byptr_safe`, `get_item` → `value_type`/`operator[]`, constructor-instead-of-`Init()`) are all orthogonal to the two defects above and all worth doing; nothing here conflicts with them. Since #1003 is queued behind this PR, resolving both lists in one iteration keeps that path short.

### Housekeeping
- The PR is **CONFLICTING** — master has moved since 07-07 (#997, #1010, #1022, #1024, #1031). Rebase needed before the next iteration.
- Commit `ba822de1` (Scalar → std::variant) overlaps #1011's territory — a sentence on how this sequences with #1011 would prevent double-work.

ianmccul:
My recommendation is rebase, and have claude address the two Code Review comments, and also the comments I made - they are all straightforward and might as well get it done in one, rather than churn through another PR.

I wouldn't worry too much about HIGH-1: it will get rewritten with #1003 anyway; an emergency fix would be to revert to the old behavior, but if doing that please make a comment on #1003 that the existing behavior is incorrect.

pcchen:
## Plan for the next iteration + coordination with #1011

Agreed with @ianmccul's recommendation above — one combined iteration rather than churning through another round. Claude Code will execute it on this branch. Concretely, the iteration will cover:

1. **Rebase onto current master** (#997, #1010, #1022, #1024, #1031 have landed since 07-07).
2. **HIGH-1 (GPU integer Div):** fixed properly rather than reverted — it's one line in `cuDiv_dispatch_typed`: `using TO = Type_class::make_floating_point_t<Type_class::type_promote_gpu_t<TL, TR>>;`, mirroring the CPU visitor in `Div.cpp`. That's less work than the revert-plus-#1003-note and keeps GPU semantics consistent with this PR's CPU true-division. (#1003 rewrites it later either way.)
3. **HIGH-2 (numpy view UAF):** make size-changing Storage operations swap in a fresh impl instead of mutating the buffer in place — which makes `numpy_ownership_test.py`'s safety argument actually true — plus a regression test that resizes under a live view.
4. **@ianmccul's 11 inline comments:** visitor-pattern conversions, `make_storage()` factory replacing the `__SII` extern globals (#673), `value_type *start_`, dead `_Init_byptr_safe` removal, `get_item` → `value_type`/`operator[]`, constructor-instead-of-`Init()`.

**Sequencing with #1011:** this PR's first commit (`ba822de1`) is a snapshot of #1011 — same refactor, same title — with the Storage work stacked on top, and the snapshot is already one iteration stale (#1011 was revised 07-08 with the visit-variant `astype` dispatch and in-place-promotion fixes). To avoid restacking on a moving target, the order is: **#1011 merges first, then the combined iteration here** — the rebase then drops the embedded Scalar commit entirely and this PR shrinks to its true Storage-only scope. Until then, please route Scalar review comments to #1011 and Storage/dispatch/linalg comments here; the `Scalar.hpp`/`Scalar.cpp` hunks in this diff are the stale copy.

PR #1022, opened 2026-07-09
refactor(UniTensor): add versioned file header

ianmccul:
Add an explicit file-format version to UniTensor serialization.

Cytnx UniTensor files previously used only the magic value `555`, followed immediately by the serialized payload. That leaves no clean way to evolve the format for future changes such as new quantum-number/sector types, updated block metadata, or revised storage layouts.

This PR introduces a versioned UniTensor header:

```text
magic = 556
version = 1
payload = existing UniTensor payload
```

The payload for version 1 is unchanged from the existing format. The loader accepts both the new versioned header and the legacy `555` header, so existing UniTensor files remain readable.

Using a new magic value avoids making old Cytnx versions silently misread a versioned file as if the version field were the UniTensor type. Old versions will reject the new file immediately instead of corrupting the load state.

The load-side UniTensor implementation selection was also factored into a small helper, keeping the Dense/Block/BlockFermionic dispatch in one place and preserving the existing error paths for Void, Sparse, and unknown UniTensor type ids.

### DIFF
--- a/tests/DenseUniTensor_test.cpp
+++ b/tests/DenseUniTensor_test.cpp
@@ -1,6 +1,7 @@
 #include "DenseUniTensor_test.h"
 
 #include <cstdio>
+#include <fstream>
 #include <filesystem>
 
 #include "test_tools.h"
@@ -4974,6 +4975,17 @@ TEST_F(DenseUniTensorTest, Save) {
   auto ut = UniTensor(bonds, labels, row_rank);
   random::uniform_(ut, 0.0, 5.0, seed);
   ut.Save(temp_file_path);
+  {
+    std::ifstream header(temp_file_path, std::ios::binary);
+    ASSERT_TRUE(header.is_open());
+    unsigned int magic = 0;
+    unsigned int version = 0;
+    header.read(reinterpret_cast<char *>(&magic), sizeof(magic));
+    header.read(reinterpret_cast<char *>(&version), sizeof(version));
+    EXPECT_EQ(magic, 556);
+    EXPECT_EQ(version, 1);
+  }
+
   UniTensor ut_loaded = UniTensor::Load(temp_file_path);
   EXPECT_TRUE(AreEqUniTensor(ut_loaded, ut));
   // for char*
--- a/src/UniTensor.cpp
+++ b/src/UniTensor.cpp
@@ -11,6 +11,38 @@
 #else
 
 namespace cytnx {
+  namespace {
+    constexpr unsigned int kLegacyUniTensorMagic = 555;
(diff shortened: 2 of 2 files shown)


tests/DenseUniTensor_test.cpp
```diff
@@ -4974,6 +4975,15 @@ TEST_F(DenseUniTensorTest, Save) {
...
   random::uniform_(ut, 0.0, 5.0, seed);
   ut.Save(temp_file_path);
+  std::ifstream header(temp_file_path, std::ios::binary);
+  ASSERT_TRUE(header.is_open());
+  unsigned int magic = 0;
+  unsigned int version = 0;
+  header.read(reinterpret_cast<char *>(&magic), sizeof(magic));
+  header.read(reinterpret_cast<char *>(&version), sizeof(version));
+  EXPECT_EQ(magic, 556);
+  EXPECT_EQ(version, 1);
```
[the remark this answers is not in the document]
ianmccul: Windows is such a messed up OS.

ianmccul:
One of the reasons for this is having proper rank-0 tensors changes the ranks in file I/O, so needs a version increment.

IvanaGyro:
Do we really need to keep the backward compatibility? We have done many breaking changes since the last release. The new release will be v2.0.0.

ianmccul:
Needed for future work anyway, but I don't see why cytnx should not be able to read tensors from cytnx 1, there have not been so many changes that affect file format as far as I can tell.  And being able to load old user data is quite important!

tests/DenseUniTensor_test.cpp
```diff
@@ -4974,6 +4975,17 @@ TEST_F(DenseUniTensorTest, Save) {
...
   random::uniform_(ut, 0.0, 5.0, seed);
   ut.Save(temp_file_path);
+  {
+    std::ifstream header(temp_file_path, std::ios::binary);
+    ASSERT_TRUE(header.is_open());
+    unsigned int magic = 0;
+    unsigned int version = 0;
+    header.read(reinterpret_cast<char *>(&magic), sizeof(magic));
+    header.read(reinterpret_cast<char *>(&version), sizeof(version));
+    EXPECT_EQ(magic, 556);
```
IvanaGyro, line 4985: use named constant here?

IvanaGyro (approved):
LGTM

tests/DenseUniTensor_test.cpp
ianmccul: Better to use the actual value here, because this is what will be contained in the actual file.  If we used a constant here and the constant changed then the code *looks* consistent, but users can no longer read their files beause the magic number changed.

Issue #969, opened 2026-07-02
Investigate CPU/CUDA wheel packaging strategy for PyPI, modeled on PyTorch and JAX


IvanaGyro:
@IvanaGyro raised this investigation to inform a future CUDA-enabled release of `cytnx` on PyPI and conda. This report was researched and written by Claude Code (Anthropic's AI coding agent) at her request.

Methodology note: every claim below was checked against a primary source — `https://pypi.org/pypi/<project>/json`, the actual conda-forge feedstock recipes (including their per-variant `.ci_support/*.yaml` build configs, not just `meta.yaml`), the jaxlib/PyTorch/Cytnx C++ and build source, or, where the claim was about what's physically inside a wheel, by downloading that wheel and inspecting its binaries — not from docs or blog summaries. See "Reproducing the wheel-inspection claims" near the end for the exact commands.

## Context

Cytnx currently ships a single CPU/OpenBLAS-oriented wheel on PyPI (`pyproject.toml` has no CUDA-related build variant, extra, or index configuration) and a single CPU-only package on conda (`conda_build/meta.yaml`, detailed below). The CMake side already has both `openblas-cuda` and `mkl-cuda` presets (`CMakePresets.json`; `mkl-cuda` inherits `["mkl", "openblas-cuda"]`) — i.e. the BLAS×CUDA build matrix already exists at the CMake level, it just isn't wired into a PyPI or conda release yet. This issue surveys how PyTorch and JAX solve the "some users have a GPU, most don't, and package managers can't always declare hardware requirements" problem on both channels, as input to that design decision.

---

### 1.1 PyTorch on PyPI: GPU by default, no opt-out extra

- **`pip install torch` on Linux is GPU-capable by default, unconditionally.** In `torch` 2.12.1's `requires_dist`, every CUDA/`nvidia-*` dependency is gated only by `platform_system == "Linux"` — never by an `extra ==` marker, and never by actual hardware detection (pip/PyPI have no such concept). So *every* Linux install of `torch`, GPU or not, downloads the full CUDA stack. The only declared extras on `torch` are unrelated (`optree`, `opt-einsum`, `pyyaml`). Windows/macOS wheels have no CUDA deps, since those markers only match `platform_system == "Linux"`.
[...]

PR #1023, opened 2026-07-09
ci(release): publish cytnx-cuda GPU wheels to PyPI (nightly + stable)

IvanaGyro:
## Problem

#538 asks for GPU testing/CI, and #969 investigated a PyPI/conda CUDA packaging strategy modeled on PyTorch/JAX. cytnx has no CUDA-enabled release today. Along the way, `Device_class`'s constructor was found to crash the whole process on any GPU-less machine running a CUDA-enabled build (`checkCudaErrors` treats `cudaGetDeviceCount`'s legitimate "no GPU here" result as a fatal error) — which would break every import of a `cytnx-cuda` package outside a real GPU host, so that fix landed here first.

## Fix

- **`src/Device.cpp`**: treat `cudaErrorNoDevice`/`cudaErrorInsufficientDriver` from `cudaGetDeviceCount` as `Ngpus == 0` instead of routing through `checkCudaErrors`'s fatal exit.
- **`tools/prepare_cuda_release.py`**: rewrites `pyproject.toml` in place for a `cytnx-cuda` release build — renames the project, adds the NVIDIA/cuTENSOR/cuQuantum runtime packages as `~=`-pinned `[project.dependencies]` (PyTorch-style dynamic linking at runtime, never vendored into the wheel), switches the CMake preset to `openblas-cuda`, and excludes those same libraries from `auditwheel repair`'s vendoring.
- **`tools/cibuildwheel_before_all_cuda.sh`** (chained onto `before-all` by the script above): provisions nvcc and the CUDA/cuTENSOR/cuQuantum dev headers inside the manylinux container from the same pip packages the wheel depends on at runtime, since the container has no system CUDA install.
- **`cytnx/_cuda_preload.py`**: preloads those shared libraries with `ctypes.CDLL(..., RTLD_GLOBAL)` before the compiled extension import, so its `NEEDED` sonames resolve without `LD_LIBRARY_PATH`. Gated on a `cuda` marker `CMakeLists.txt` now writes to `vinfo.tmp` for `USE_CUDA` builds, so it's a no-op on the plain CPU package.
- **`tools/prune_nightly_wheels.py`**: parameterized by `--package` so the shared `cytnx-nightly-wheels` anaconda.org channel can prune `cytnx` and `cytnx-cuda` nightlies independently.
- **`.github/workflows/release_pypi_cuda.yml`**: new pipeline mirroring `release_pypi.yml`'s channel-resolution/publish structure — `master` push → nightly upload, `v*` tag → stable PyPI release via trusted publishing, plus `workflow_dispatch`. Not run on `pull_request` (an nvcc build is too slow/expensive to gate every PR on, and there's no GitHub-hosted GPU runner budget for it). Single Linux x86_64 `manylinux_2_28` leg on a plain (GPU-less) `ubuntu-24.04` runner; `CIBW_TEST_COMMAND` asserts `cytnx.Device.Ngpus == 0`.
- **`.github/workflows/ci-gpu_tests.yml`** removed entirely — its function is superseded by `release_pypi_cuda.yml`'s own build/link/import validation; real GPU-hardware test execution is out of this PR's scope.

## Known gaps — deliberately left open

- No GPU hardware is available anywhere in this PR's validation, so `release_pypi_cuda.yml` can only prove the extension builds, links, and imports cleanly on a GPU-less runner — not that GPU code paths are numerically correct on real hardware.
- Publishing to PyPI's `cytnx-cuda` project needs a maintainer to configure OIDC trusted publishing for this workflow file, the same one-time setup `release_pypi.yml`'s `cytnx` project already has.
- #1063 (an nvcc internal-compiler-error on master's own Storage refactor, found during this PR's rebase but unrelated to it) is tracked separately and not addressed here.

## Testing

- `src/Device.cpp`'s fix: full local CUDA compile+link via the `openblas-cuda` preset (`libcytnx.a` built cleanly with zero errors), both before and after rebasing onto master.
- `tools/prepare_cuda_release.py`: round-tripped against scratch copies of `pyproject.toml` in both pre- and post-#1057 state, then reparsed with `tomllib` to confirm every rewritten field (`name`, `dependencies`, `cmake.args`, `repair-wheel-command`, `before-all`, `environment`) is correct; this caught a real `tomlkit` bug where mutating an already-parsed inline table in place silently drops the separating commas, producing invalid TOML — fixed by rebuilding the table from scratch instead.
- `cytnx/_cuda_preload.py`: exercised `_discover_lib_paths()`/`preload()` directly on this (CPU-only, no GPU) sandbox, confirming the no-op path behaves correctly when the NVIDIA/cuTENSOR/cuQuantum packages aren't installed.
- Every NVIDIA/cuTENSOR/cuQuantum PyPI package name, version, and wheel-internal `include`/`lib`/`bin` layout referenced by `tools/prepare_cuda_release.py` and `tools/cibuildwheel_before_all_cuda.sh` was verified against the real PyPI JSON API and by downloading and inspecting the actual wheel contents, not assumed.
- `.github/workflows/release_pypi_cuda.yml` / `tools/cibuildwheel_before_all_cuda.sh`: reviewed by hand against `release_pypi.yml`'s proven structure; the actual manylinux+CUDA container build has not run anywhere in this environment (no Docker registry access, no GPU) — relying on this PR's own CI for first real validation.

Relates to #538 (adds the GPU wheel-release side of that ask; real GPU-hardware ctest execution is still open elsewhere) and implements the design from #969. See #1063 for the unrelated, separately-tracked nvcc issue found along the way.


CMakePresets.json
```diff
@@ -195,6 +195,17 @@
       "description": "Run tests for builds with CUDA support.",
       "configurePreset": "debug-openblas-cuda",
       "inherits": "cpu-only"
+    },
+    {
+      "name": "gpu-only",
+      "description": "Run only the GPU-labeled tests (tests/gpu/gpu_test_main) for builds with CUDA support.",
+      "configurePreset": "debug-openblas-cuda",
+      "inherits": "cpu-and-cuda",
```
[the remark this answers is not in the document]
IvanaGyro: Good catch, fixed in 1b8932d.

---

IvanaGyro:
@yingjerkao This PR successfully builds with CUDA. The large action runner label is needed for running tests on the GPU machine.

.github/workflows/ci-cuda_tests.yml
```diff
@@ -0,0 +1,235 @@
...
+jobs:
+  build-and-test:
+    # GPU_RUNNER_LABEL should be set as a repository (or org) Actions
+    # variable to the label of the GitHub-hosted GPU runner group (see
+    # #538) — the label was not yet known when this workflow was written.
+    # Until it's set, fall back to ubuntu-latest so the configure/build
+    # steps (which don't need a physical GPU — nvcc compiles fine without
+    # one) still run on every PR; only GPU-and-driver-info and the actual
+    # test execution steps need real hardware.
+    runs-on: ${{ vars.GPU_RUNNER_LABEL || 'ubuntu-latest' }}
```
IvanaGyro: run on the runner as the cpu pypi release workflow. And rename workflow.

tools/prepare_cuda_release.py
```diff
@@ -0,0 +1,111 @@
...
+# NVIDIA/cuTENSOR dependencies, need listing).
+CUDA_RUNTIME_DEPENDENCIES = [
+    "nvidia-cuda-runtime ==13.3.29 ; sys_platform == 'linux'",
+    "nvidia-cublas ==13.6.0.2 ; sys_platform == 'linux'",
+    "nvidia-cusparse ==12.8.2.51 ; sys_platform == 'linux'",
+    "nvidia-curand ==10.4.3.29 ; sys_platform == 'linux'",
+    "nvidia-cusolver ==12.2.6.9 ; sys_platform == 'linux'",
+    "cutensor-cu13 ==2.7.0 ; sys_platform == 'linux'",
+    "cutensornet-cu13 ==2.13.0 ; sys_platform == 'linux'",
+    "custatevec-cu13 ==1.14.0 ; sys_platform == 'linux'",
```
IvanaGyro: I think it's fine to be ~= not ==

IvanaGyro: Switched all eight entries in `CUDA_RUNTIME_DEPENDENCIES` from `==` to `~=` in 57581e8, e.g. `nvidia-cublas ~=13.6.0.2` now allows any `13.6.0.x` patch release instead of pinning to `13.6.0.2` exactly.

---

.github/workflows/ci-cuda_tests.yml
IvanaGyro: Superseded: this file was removed entirely in f0cec7d rather than renamed/re-aligned — its function is covered by `release_pypi_cuda.yml` (in progress on this PR) instead.

---

.github/workflows/release_pypi_cuda.yml
```diff
@@ -0,0 +1,233 @@
...
+        type: choice
+        options:
+          - auto
+          - stable
+          - nightly
+        default: auto
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.run_id }}
+  cancel-in-progress: false
```
IvanaGyro: Should also run on pull_request. Please refer concurrency settings in https://github.com/Cytnx-dev/Cytnx/blob/e37e3c897579c7ac47a72425a77bf1d5ce9395c6/.github/workflows/release_pypi.yml.

.github/workflows/release_pypi_cuda.yml
```diff
@@ -0,0 +1,233 @@
...
+        # adding manylinux_aarch64 later is a one-line change, not a
+        # restructure.
+        os: [ubuntu-24.04]
+
+    steps:
+    - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4.3.1
+      with:
+        fetch-depth: 0
+        submodules: recursive
+
```
IvanaGyro, line 134: Should add a step : Merge with latest target branch (pull_request only) here

CMakeLists.txt
```diff
@@ -174,6 +174,13 @@ cmake_dependent_option(HPTT_ENABLE_ARM "HPTT variant" OFF "USE_HPTT;NOT HPTT_ENA
 cmake_dependent_option(HPTT_ENABLE_AVX "HPTT variant" OFF "USE_HPTT;NOT HPTT_ENABLE_ARM; NOT HPTT_ENABLE_IBM" OFF)
 cmake_dependent_option(HPTT_ENABLE_IBM "HPTT variant" OFF "USE_HPTT;NOT HPTT_ENABLE_ARM; NOT HPTT_ENABLE_AVX" OFF)
 cmake_dependent_option(USE_CUDA "Enable CUDA support." OFF "NOT BACKEND_TORCH" OFF)
+# cytnx/_cuda_preload.py needs to know whether this is a CUDA build before
+# importing the compiled extension (whose own import is what would otherwise
+# reveal that), so record it in vinfo.tmp -- already installed alongside the
+# extension and already read by cytnx/__init__.py, just not this early.
+if(USE_CUDA)
+  FILE(APPEND "${CMAKE_BINARY_DIR}/vinfo.tmp" "cuda\n")
+endif()
```
IvanaGyro: Why not use RPATH setting from workflow file?

.github/workflows/release_pypi_cuda.yml
IvanaGyro: Done in 5f7308c. Added the `pull_request` trigger and copied `release_pypi.yml`'s concurrency block verbatim: `group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}` with `cancel-in-progress: ${{ github.event_name == 'pull_request' }}`. To keep PRs from publishing, `DetermineChannel` now emits a `publish` output (pull_request → nightly + publish=false; push/tag/dispatch → publish=true), and both `ReleasePyPI` and `PublishNightlyAnaconda` additionally gate on `publish == 'true'` — so a PR builds the CUDA wheel as a health check and stops there.

---

.github/workflows/release_pypi_cuda.yml
IvanaGyro: Added in 5f7308c — the "Merge with latest target branch (pull_request only)" step is now the first step after checkout in `BuildWheel`, copied verbatim from `release_pypi.yml` (config the github-actions bot identity, fetch + `git merge --no-edit` the PR base ref, then `git submodule update --init --recursive` in case the merge moved a gitlink).

---

CMakeLists.txt
IvanaGyro: Good call — switched to RPATH and removed the whole ctypes-preload mechanism in the latest push. `tools/prepare_cuda_release.py` now sets `CMAKE_INSTALL_RPATH=$ORIGIN/../nvidia/cu13/lib;$ORIGIN/../cutensor/lib;$ORIGIN/../cuquantum/lib` (the Linux `pycytnx` target already folds `${CMAKE_INSTALL_RPATH}` into its `INSTALL_RPATH`), and `cytnx/_cuda_preload.py`, its pytest, the `vinfo.tmp` "cuda" marker here in CMakeLists.txt, and the `__init__.py` gating are all deleted.

I verified this works end-to-end rather than assuming:

1. **auditwheel preserves a build-set `$ORIGIN`-relative RUNPATH through `repair --exclude`** (auditwheel 6.7.0): built a dummy extension with RPATH `$ORIGIN/../foo/lib` NEEDing an excluded `libfoo.so.1`, repaired it → the repaired `.so` kept `$ORIGIN/../foo/lib` as RUNPATH. When a *second* lib is vendored (not excluded), auditwheel **appends** its `.libs` dir: final RUNPATH `$ORIGIN/../foo/lib:$ORIGIN/../mypkg.libs`. So the excluded CUDA libs (via my RPATH) and the vendored openblas/arpack (via auditwheel's dir) both resolve.
2. **The extension only needs to cover its own direct NEEDED entries** — the real NVIDIA wheels each carry their own `$ORIGIN`-relative RUNPATHs across the three dirs, so the transitive graph self-resolves. Confirmed from the actual wheels: `libcutensornet.so.2` → `$ORIGIN:$ORIGIN/../../cutensor/lib:$ORIGIN/../../nvidia/cu13/lib`; `libcutensor.so.2`/`libcustatevec.so.1` → `$ORIGIN:$ORIGIN/../../nvidia/cu13/lib`; `libcublas`/`libcusolver` → `$ORIGIN`.

Separately, the first PR build surfaced a real bug in the toolchain provisioning (`find_package(CUDAToolkit)` → `missing: CUDA_CUDART`): the pip CUDA wheels ship only versioned sonames, so `find_library()` couldn't see them. Fixed in 9d845eb by creating the unversioned `libX.so` dev symlinks in each toolchain lib dir.

---

.github/workflows/release_pypi_cuda.yml
```diff
@@ -0,0 +1,269 @@
...
+    - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4.3.1
+      with:
+        sparse-checkout: tools
+
+    - name: Prune old nightly releases
+      # Prunes cytnx-cuda specifically; the CPU cytnx nightlies are
+      # pruned by release_pypi.yml's own PublishNightlyAnaconda job.
+      env:
+        ANACONDA_API_TOKEN: ${{ secrets.ANACONDA_ORG_UPLOAD_TOKEN }}
+      run: python3 tools/prune_nightly_wheels.py --keep 4 --package cytnx-cuda
```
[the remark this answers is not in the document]
IvanaGyro: Valid — fixed in 26abc41. `fetch_versions()` now catches a 404 `HTTPError` and returns an empty list (package not on the channel yet → nothing to prune), so the first `cytnx-cuda` nightly's prune step no longer fails before the upload creates the package; any non-404 HTTP error still propagates. Added regression tests (`404 → []`, non-404 → raises, normal version-list parse), and confirmed the pre-fix code raised on the 404 while the fix returns `[]`.

---

yingjerkao (approved):

PR #1024, opened 2026-07-09
fix(UniTensor): read file names without requiring null terminator

### LINKED ISSUE #852: ASan heap-buffer-overflow in UniTensor::_Load (non-terminated std::string) — 11 tests fail on debug-openblas-cuda
(Issue written by Claude on behalf of @IvanaGyro.)

## Summary

Building with the **`debug-openblas-cuda`** preset (`USE_DEBUG=ON` → `-fsanitize=address`) and running `ctest --preset cpu-and-cuda` produces **11 failures out of 1678**. They are not flaky timeouts — AddressSanitizer reports genuine **`heap-buffer-overflow`** errors. Most share a single root cause: an off-by-one over-read in `UniTensor::_Load`. (The release builds — `openblas-cuda` and `mkl-cuda` — are 100% green, because the 1-byte over-read is benign without ASan, but it is undefined behavior.)

## Failing tests

| ctest # | test | time |
|---:|---|---:|
| 94 | `Rsvd_truncate.dense_exp_svals_test` | 3.35 s |
| 95 | `Rsvd_truncate.dense_exp_svals_no_power_iteration_test` | 2.91 s |
| 98 | `Gesvd_truncate.dense_exp_svals_test` | 2.53 s |
| 99 | `Gesvd_truncate.U1_exp_svals_test` | 2.60 s |
| 100 | `Gesvd_truncate.U1_exp_svals_minblockdim_test` | 2.35 s |
| 103 | `Svd_truncate.dense_exp_svals_test` | 2.56 s |
| 104 | `Svd_truncate.U1_exp_svals_minblockdim_test` | 2.58 s |
| 107 | `Rsvd.dense_exp_svals_test` | 2.25 s |
| 108 | `Rsvd.dense_exp_svals_no_power_iteration_test` | 2.09 s |
| 232 | `SearchTreeTest.BasicSearchOrder2` | 2.35 s |
| 859 | `BlockFermionicUniTensorTest.SaveLoad` | 3.54 s |

## Root cause (the `Load`-based failures)

`cytnx::UniTensor::_Load` (`src/UniTensor.cpp`) reads the name into a buffer sized to exactly `len_name` bytes, then constructs a `std::string` with the **C-string constructor**:

```cpp
char *cname = (char *)malloc(sizeof(char) * len_name);   // no room for a '\0'
f.read(cname, sizeof(char) * len_name);
this->_impl->_name = std::string(cname);                 // strlen(cname) reads past the buffer
free(cname);
```

`std::string(const char*)` runs `strlen`, which reads until a NUL — but `cname` is not NUL-terminated, so it reads (at least) one byte past the allocation. Every `Load()` of a UniTensor that has a non-empty `_name` hits this, which is why the SVD/Rsvd/Gesvd-truncate tests (they `Load` reference `.cytnx` files) and `BlockFermionicUniTensorTest.SaveLoad` all trip it.

### Clean ASan trace (`Svd_truncate.dense_exp_svals_test`, run individually, paths sanitized)

```
==ERROR: AddressSanitizer: heap-buffer-overflow ... READ of size 84 at ... thread T0
    #0 __interceptor_strlen
    #1 std::char_traits<char>::length(char const*)
    #2 std::__cxx11::basic_string<...>::basic_string(char const*, allocator const&)
    #3 cytnx::UniTensor::_Load(std::fstream&)            <repo>/src/UniTensor.cpp:130
    #4 cytnx::UniTensor::Load(std::string const&)        <repo>/src/UniTensor.cpp:184
    #5 SvdTruncateTest::CheckResult(...)                 <repo>/tests/linalg_test/Svd_truncate_test.cpp:206
    #6 SvdTruncateTest::..._dense_exp_svals_test::TestBody()  Svd_truncate_test.cpp:97
...
0x... is located 0 bytes after 83-byte region   <-- read of 84 from an 83-byte malloc
allocated by thread T0 here:
    #1 cytnx::UniTensor::_Load(std::fstream&)            <repo>/src/UniTensor.cpp:128
SUMMARY: AddressSanitizer: heap-buffer-overflow in __interceptor_strlen
```

### Suggested fix

Use the length-aware constructor (don't rely on a terminator):

```cpp
this->_impl->_name = std::string(cname, len_name);
```

or allocate `len_name + 1` and set `cname[len_name] = '\0'`. The label-reading block just below in `_Load` (and any other `std::string(buf)`-from-fixed-buffer sites in the load path) should be audited for the same pattern.

## Note on `SearchTreeTest.BasicSearchOrder2`

This one is also a `heap-buffer-overflow` but is not on the `Load` path, so it likely has a **separate** root cause; it should be investigated independently (I can capture its individual ASan trace if wanted).

## Reproduce

```bash
conda activate cytnx
export CUTENSOR_ROOT=... CUQUANTUM_ROOT=...
cmake --preset debug-openblas-cuda
ASAN_OPTIONS=protect_shadow_gap=0:replace_intrin=0:detect_leaks=0 \
  cmake --build --preset debug-openblas-cuda -j
ASAN_OPTIONS=protect_shadow_gap=0:replace_intrin=0:detect_leaks=0 \
  ctest --preset cpu-and-cuda --output-on-failure
# single, clean trace:
ASAN_OPTIONS=protect_shadow_gap=0:replace_intrin=0:detect_leaks=0 \
  ./build/debug-openblas-cuda/tests/test_main --gtest_filter='Svd_truncate.dense_exp_svals_test'
```

(`ASAN_OPTIONS` is required at build and test time because `gtest_discover_tests` runs the ASan binary during the build and the NVIDIA driver otherwise makes `cuInit()` return `cudaErrorMemoryAllocation` — see `.vscode/settings.json`.)

Environment: GTX 1660 SUPER / RTX 2080 Ti, CUDA 12, gcc/gxx 13 (conda `cytnx` env), GoogleTest 1.17.

ianmccul:
## Problem
`UniTensor::Load` reads a length-prefixed tensor name from the file, allocates exactly `len_name` bytes, and then constructs `std::string(cname)`. That constructor expects a null-terminated C string, so it can read past the allocated buffer when the serialized name is not null-terminated.

## Fix
Resize the destination `std::string` to the serialized length and read directly into `string::data()`. This preserves the length-prefixed file format and avoids the out-of-bounds read.

## Testing
- `clang-format-14 -i src/UniTensor.cpp`
- `git diff --check`
- `ASAN_OPTIONS=protect_shadow_gap=0:replace_intrin=0:detect_leaks=0 GCOV_PREFIX=/tmp/cytnx-gcov ./build_codex/debug-openblas-cpu/tests/test_main --gtest_filter=Rsvd.dense_exp_svals_test --gtest_brief=1`

Resolves #852 (but not the SearchTreeTest.BasicSearchOrder2 part)

### DIFF
--- a/src/UniTensor.cpp
+++ b/src/UniTensor.cpp
@@ -119,10 +119,8 @@ namespace cytnx {
     cytnx_uint32 len_name;
     f.read((char *)&len_name, sizeof(cytnx_uint32));
     if (len_name != 0) {
-      char *cname = (char *)malloc(sizeof(char) * len_name);
-      f.read(cname, sizeof(char) * len_name);
-      this->_impl->_name = std::string(cname);
-      free(cname);
+      this->_impl->_name.resize(len_name);
+      f.read(this->_impl->_name.data(), sizeof(char) * len_name);
     }
 
     cytnx_uint64 rank;


ianmccul:
> ## Code Review
> 
> This pull request optimizes the deserialization of the UniTensor name in `src/UniTensor.cpp`. Instead of allocating a temporary character buffer using `malloc` and copying it to a `std::string`, the code now resizes the `std::string` directly and reads the data into it. This improves performance and simplifies memory management. There are no review comments, and I have no additional feedback to provide.

Well, arguably the more important fix is avoiding the buffer overflow.... Do not load untrusted Cytnx tensors from the internet, it starts bitcoin miner and steals your wallet... although I think it would be very hard with this overflow, the attacker doesn't control what gets read into the string.

IvanaGyro:
related: #852

IvanaGyro (approved):

PR #1028, opened 2026-07-09
fix(ci): gate C++ coverage on lines only, drop branch coverage

### LINKED ISSUE #932: CI coverage: gate C++ on line coverage only (drop branch coverage); keep line + branch for Python
## 🎯 Proposal

Stop gating pull requests on **C++ branch coverage**. For compiled C++ (the `cpp` flag), have Codecov evaluate **line coverage only**. For Python (the `python` flag), keep **both line and branch** coverage as today.

## 🔍 Motivation

In #927 the `codecov/patch` status fails at ~5.47% even though every patched C++ line is executed by the new tests. The failure is entirely **branch "partials"**, not uncovered lines.

A "branch" in gcov/gcovr is a CFG arc, so C++ reports include compiler-generated arcs that are not author-written logic:

- 💥 exception-unwind edges ("if this call throws, run destructors") at nearly every potentially-throwing call,
- 🧹 destructor / static-init arcs,
- 📦 inlined STL arcs (e.g. the `std::string` small-string-optimization branch inside an `initializer_list` argument — never taken unless a label string exceeds 15 chars).

Measured on #927's actual `.gcda`, of 234 partial-branch lines in `src/tn_algo/DMRG.cpp`, **206 are throw/unreachable** and ~20 more are inlined-STL arg construction. Only the `if (verbose)` logging branches were genuine, testable control flow. None of the exception/STL arcs can be covered without injecting a throw at each call site — so no amount of additional testing moves the number.

📈 Concretely, dropping the exception/unreachable branches from scoring takes the #927 patch coverage from **4.1% → 95.9%** across the three touched files, with the lines themselves already fully executed.

## ✅ Why line-only for C++ is the right policy (not a workaround)

A survey of comparable scientific/ML packages shows a consistent pattern: **C++ line coverage = yes, C++ branch coverage = no; Python = both.**

| Package | C++ line cov | C++ branch cov | PR gate on C++ branch |
|---|---|---|---|
| PyTorch | line (internal `lcov` tool) | ❌ | ❌ (Codecov disabled 2022) |
| TensorFlow | fragile (`bazel coverage`) | ❌ | ❌ |
| JAX / XLA | not in PR | ❌ | ❌ |
| NumPy | not collected (TODO gcov) | ❌ | ❌ (`informational: true`) |
| SciPy | line via `lcov` | ❌ **deliberately off** | ❌ |
| CuPy | not possible (JIT kernels) | ❌ | ❌ (`project: off`, `patch: 0%`) |
| ContourPy (pybind11, closest analog) | line via `lcov` (`branch_coverage=0`) | ❌ | ❌ |

Projects that report C++ coverage at all (SciPy, ContourPy) collect **line** coverage and explicitly disable branch detection. Branch coverage of compiled C++ is widely treated as a local diagnostic, not a merge gate, because forcing it up requires artificial tests (e.g. simulating allocator failure to cover an STL destructor arc) that add maintenance cost without improving correctness.

## 🐍 Python coverage (for future reference — keep as-is)

This proposal **does not touch the Python side**. Recording the current Python setup here so the contrast with C++ is explicit and so future changes don't accidentally weaken it:

- 🛠️ **How it's collected:** the CI step *"Run pytest with Python coverage"* runs
  ```
  pytest pytests/ --doctest-modules --cov=cytnx --cov-branch \
    --cov-report=xml:coverage-python.xml --cov-report=term
  ```
  `--cov-branch` enables **branch** coverage on top of line coverage, via `coverage.py`/`pytest-cov`. The report is uploaded under the `python` flag (`codecov.yml` → `flags.python.paths: [cytnx/]`).
- 📊 **Why branch coverage is fine for Python (but not C++):** `coverage.py` instruments at the **Python bytecode / AST level**, so a "branch" is an actual author-written control-flow decision (an `if`/`for`/`while`/`except` taken vs. not). There are no compiler-injected exception-unwind or inlined-template arcs — the noise that makes C++ branch coverage unactionable simply doesn't exist. So Python branch percentages are meaningful and worth gating.
- 🔒 **What stays:** both `coverage.status.project.python` and the `patch` status continue to score line **and** branch for the `python` flag. No change requested here.
- ⚠️ **Future-proofing note:** if Option A below is chosen, confirm the `parsers.gcov.branch_detection` toggle only affects the gcov-format (`cpp`) upload and does **not** strip branch data from the pytest-cov Cobertura XML. If there's any risk it does, prefer **Option B** so the two flags stay cleanly separated.

## 🧩 Suggested implementation (pick one; maintainers' call)

**Option A — disable C++ branch detection at the Codecov parser (smallest change, `codecov.yml` only):**
```yaml
parsers:
  gcov:
    branch_detection:
      conditional: false
      loop: false
      method: false
      macro: false
```
This targets the gcov-format report (the `cpp` flag) and leaves the Python report's branch data intact. (Worth confirming it doesn't also affect the pytest-cov Cobertura upload; if it does, prefer Option B.)

**Option B — strip branches at the source for the C++ upload only (cleanest separation by flag):** drop branch output from the `cpp` gcovr report (e.g. don't emit branch data, or post-process the XML to remove `<branch>` entries), while Python keeps `--cov-branch`. This guarantees C++ = line-only and Python = line+branch with no parser ambiguity.

**🚫 Not recommended — `GCOVR_EXCL_BR_START/STOP` source annotations:** would require littering production C++ around every `try`/`catch` and STL declaration; pollutes the code and doesn't scale.

In all options, keep the existing C++ **line** coverage status so "which C++ lines were never executed" remains a visible, enforceable signal.

## 📌 Scope

- **C++ (`cpp` flag):** line coverage gate only; branch coverage no longer gates.
- **Python (`python` flag):** unchanged — both line and branch (`--cov-branch` stays).
- No change to the test suites themselves.

---
*Filed by Claude (Claude Code) at @IvanaGyro's request, summarizing a two-part research survey of C++ testing/coverage practices across PyTorch, TensorFlow, JAX/XLA, NumPy, SciPy, CuPy, and ContourPy. Some peer-project config specifics in the survey were reported second-hand and are worth a quick confirmation before adopting.*

IvanaGyro:
## Problem

Codecov's `codecov/patch` status gates C++ (`cpp` flag) on both line and branch coverage. gcov/gcovr branch coverage for compiled C++ counts compiler-generated CFG arcs — exception-unwind edges, destructor/static-init arcs, inlined STL branches — alongside author-written control flow. Those arcs can't be exercised by ordinary tests, so a PR's branch percentage can stay low even when every patched line is fully executed (see #932 for the measured breakdown on #927).

## Fix

Option B from #932: strip branch records at the source. The gcovr invocation in `.github/workflows/ci-cmake_tests.yml` now passes `--exclude-branches-by-pattern '.*'`, so the Cobertura XML uploaded under the `cpp` flag carries no branch data (`branches-valid="0"`, every line `branch="false"`) and Codecov scores C++ on line coverage only.

Option A from #932 (`parsers.gcov.branch_detection` in `codecov.yml`) was evaluated first on this PR and shown to be ineffective: that setting only configures Codecov's parser for raw gcov-format text reports, while this CI uploads gcovr-generated Cobertura XML, which goes through the Cobertura parser — the setting never takes effect (the initial Codecov report on this PR, with only the Option A change pushed, showed branch/partial counts unchanged at 13823/7391).

- **C++ (`cpp` flag):** line coverage only.
- **Python (`python` flag):** unchanged — pytest-cov still runs with `--cov-branch` and gates on both line and branch coverage, since coverage.py branches are real author-written control flow.

## Testing

- Verified locally with gcovr 8.6 on a small C++ program with a throwing branch: without the flag the Cobertura XML reports `branches-valid="4"` with partials; with `--exclude-branches-by-pattern '.*'` it reports `branches-valid="0"` and every line `branch="false"`, while line coverage is identical.
- `yaml.safe_load` passes on the edited workflow file.
- End-to-end effect (branch/partial counts dropping to 0 for the `cpp` flag) will be visible in this PR's own Codecov report once CI runs.

Closes #932.

### DIFF
--- a/.github/workflows/ci-cmake_tests.yml
+++ b/.github/workflows/ci-cmake_tests.yml
@@ -156,9 +156,16 @@ jobs:
         #   - C++ paths exercised through the Python bindings under pytest
         # pybind/ is intentionally kept; codecov.yml gives it its own
         # component in the PR comment.
+        #
+        # --exclude-branches-by-pattern '.*' strips ALL branch records from
+        # the XML, so the `cpp` flag gates on line coverage only (#932):
+        # gcov branch data counts compiler-generated CFG arcs (exception
+        # unwind, destructors, inlined STL) that tests cannot exercise.
+        # Python coverage is unaffected and keeps --cov-branch.
         run: |
           gcovr -r ${{ github.workspace }} . \
             --xml-pretty -o coverage-cpp.xml \
+            --exclude-branches-by-pattern '.*' \
             --exclude '.*_deps.*' \
             --exclude '.*thirdparty.*' \
             --exclude '.*tests/.*' \


ianmccul:
Yes

ianmccul (approved):

Issue #853, opened 2026-05-29
ASan heap-buffer-overflow in OptimalTreeSolver::solve (adjacencyMatrix index out of bounds), search_tree.cpp:151


IvanaGyro:
(Issue written by Claude on behalf of @IvanaGyro.)

## Summary

`SearchTreeTest.BasicSearchOrder2` fails under the `debug-openblas-cuda` (AddressSanitizer) build with a **`heap-buffer-overflow` READ** inside `OptimalTreeSolver::solve`. This is a **separate** root cause from the `UniTensor::_Load` overflow tracked in #852 — splitting it out as requested there.

Because `OptimalTreeSolver::solve` backs `SearchTree::search_order()` (the contraction-order planner used by `Network`), this is a production-path out-of-bounds access, not a test-only artifact. It is benign in release (no ASan), but it is undefined behavior.

## ASan trace (run individually, paths sanitized)

```
==ERROR: AddressSanitizer: heap-buffer-overflow ... READ of size 8 at ... thread T0
    #0 std::_Base_bitset<2ul>::_M_getword(unsigned long) const            bitset:128
    #1 std::bitset<128ul>::_Unchecked_test(unsigned long) const           bitset:1143
    #2 std::bitset<128ul>::test(unsigned long) const                      bitset:1396
    #3 cytnx::OptimalTreeSolver::solve(std::vector<PseudoUniTensor> const&, bool)   <repo>/src/search_tree.cpp:151
    #4 cytnx::SearchTree::search_order()                                  <repo>/src/search_tree.cpp:218
    #5 SearchTreeTest_BasicSearchOrder2_Test::TestBody()                  <repo>/tests/search_tree_test.cpp:98
...
0x... is located 0 bytes after 64-byte region    <-- one element past the vector
allocated by thread T0 here:
    #0 operator new(unsigned long)
    ...
    #8 cytnx::OptimalTreeSolver::solve(...)                               <repo>/src/search_tree.cpp:112
SUMMARY: AddressSanitizer: heap-buffer-overflow bitset:128 in std::_Base_bitset<2ul>::_M_getword
```

The 64-byte allocation is the adjacency matrix (`n == 4` here → 4 × `std::bitset<128>` of 16 bytes), and the faulting read is at element index `n` (one past the end).

## Root cause

`src/search_tree.cpp`, `OptimalTreeSolver::solve`:

```cpp
const size_t n = nodes.size();
[...]

PR #1031, opened 2026-07-09
fix(UniTensor): make clone_meta() exception-safe to stop leaking on validation throws

IvanaGyro:
## Problem

While verifying #1027 (the fix for #853), I ran the full `ctest` suite under `debug-openblas-cpu` and `debug-mkl-cpu` and found 27 failing tests, all `LeakSanitizer: detected memory leaks` (a different ASan detector than #852's `heap-buffer-overflow`/`strlen` root cause) — reported in [a comment on #852](https://github.com/Cytnx-dev/Cytnx/issues/852#issuecomment-4928277007). @ianmccul then diagnosed the mechanism in [a follow-up comment](https://github.com/Cytnx-dev/Cytnx/issues/852#issuecomment-4928400402): `DenseUniTensor::clone_meta()` returns a raw pointer instead of an owned smart pointer.

Concretely: `DenseUniTensor::clone_meta()`, `BlockUniTensor::clone_meta(bool, bool)`, and `BlockFermionicUniTensor::clone_meta(bool, bool)` allocate a new node with a bare `new` and return a raw pointer. Every caller (`relabel`, `relabels`, `set_rowrank`, `permute`, `permute_nosignflip`, `astype`, `contiguous`, `to_dense`, `get`) then runs further mutation on that raw pointer — typically through `set_label`/`set_labels`/`set_rowrank_`/`Tensor::astype`, all of which validate their arguments via `cytnx_error_msg` and can throw — before finally wrapping the pointer in a `boost::intrusive_ptr`. Since the object has no owner until that final wrap, a throw partway through leaks the allocation.

## Fix

`Tensor_impl::_clone_meta_only()` (`include/backend/Tensor_impl.hpp`) already gets this right: it returns a `boost::intrusive_ptr` from the point of allocation, so the object is owned throughout the caller's body regardless of what throws. This PR changes the three `UniTensor` `clone_meta()` overloads to do the same, and updates every call site's local variable from a raw `T *` to `boost::intrusive_ptr`. No caller's logic changes — ownership now starts at allocation instead of at the end of the function body, so a mid-function throw unwinds through the `intrusive_ptr`'s destructor instead of leaking.

Two related bugs fixed in the same pass since they were touched anyway:
- `BlockUniTensor::contiguous()` and `BlockFermionicUniTensor::contiguous()` each allocated a throwaway node with `new` and immediately discarded it by reassigning the variable to `clone_meta()`'s result on the next line.
- `DenseUniTensor::astype()` returned the pre-wrap raw pointer (`tmp`) instead of the wrapped `out`, leaving `out` computed and discarded — harmless under the old add-ref semantics, but dead code that no longer compiles once `tmp` becomes an `intrusive_ptr`.

`DenseUniTensor::contract()` and the `BlockUniTensor`/`BlockFermionicUniTensor` `contract()` overloads build their result the same unowned-raw-pointer way, but through a direct `new`, not `clone_meta()`; none of the currently failing tests exercise their error paths, so that's left for a follow-up rather than folded into this diff.

## Testing

- `ctest --preset cpu-only` (`debug-openblas-cpu`): all 27 previously-failing tests (`linalg_Test.BkFUt_*`, `Arnoldi_Ut.fermionic_ArnoldiFermionic`, `Lanczos_Gnd.fermionic_LanczosFermionic`, `Lanczos.fermionic_LanczosArpackFermionic`, `DenseUniTensorTest.{set_rowrank_err,relabel,relabels,astype,permute_err,permute__err}`, `BlockFermionicUniTensorTest.{LinAlgElementwise,PermuteInPlaceOnSharedBlockDoesNotCorruptOtherHolder,convert_from_permute_roundtrip}`, `BlockUniTensorTest.{set_rowrank,relabels,relabel,contiguous,same_data}`, `NetworkTest.Network_fermionic_matches_contract`) now pass. Only remaining failure is `SearchTreeTest.BasicSearchOrder2`, a separate pre-existing issue tracked by #853 / #1027, unaffected by this change.
- `ctest --test-dir build/debug-mkl-cpu` (`debug-mkl-cpu`): same result — same 27 tests fixed, same single unrelated pre-existing failure.
- `pytest pytests/`: 75 passed, 1 skipped (unrelated to this change).
- `pre-commit run --files include/UniTensor.hpp src/BlockFermionicUniTensor.cpp src/BlockUniTensor.cpp src/DenseUniTensor.cpp src/linalg/ExpH.cpp src/linalg/ExpM.cpp`: clang-format (v14) clean.

Closes the 27-test leak class reported against #852; `SearchTreeTest.BasicSearchOrder2` itself is tracked separately by #853.

### DIFF
--- a/include/UniTensor.hpp
+++ b/include/UniTensor.hpp
@@ -452,8 +452,8 @@ namespace cytnx {
    public:
     Tensor _block;
     std::vector<Tensor> _interface_block;  // this is serves as interface for get_blocks_();
-    DenseUniTensor *clone_meta() const {
-      DenseUniTensor *tmp = new DenseUniTensor();
+    boost::intrusive_ptr<DenseUniTensor> clone_meta() const {
+      boost::intrusive_ptr<DenseUniTensor> tmp(new DenseUniTensor());
       tmp->_bonds = vec_clone(this->_bonds);
       tmp->_labels = this->_labels;
       tmp->_is_braket_form = this->_is_braket_form;
@@ -514,18 +514,16 @@ namespace cytnx {
     }
 
     boost::intrusive_ptr<UniTensor_base> set_rowrank(const cytnx_uint64 &new_rowrank) const {
-      DenseUniTensor *out_raw = this->clone_meta();
+      boost::intrusive_ptr<DenseUniTensor> out_raw = this->clone_meta();
       out_raw->_block = this->_block;
       out_raw->set_rowrank_(new_rowrank);
-      boost::intrusive_ptr<UniTensor_base> out(out_raw);
-      return out;
+      return out_raw;
     }
 
     boost::intrusive_ptr<UniTensor_base> clone() const {
-      DenseUniTensor *tmp = this->clone_meta();
+      boost::intrusive_ptr<DenseUniTensor> tmp = this->clone_meta();
       tmp->_block = this->_block.clone();
-      boost::intrusive_ptr<UniTensor_base> out(tmp);
-      return out;
+      return tmp;
     };
     bool is_contiguous() const { return this->_block.is_contiguous(); }
     unsigned int dtype() const { return this->_block.dtype(); }
@@ -595,9 +593,8 @@ namespace cytnx {
                                                  const std::string &new_label);
 
     boost::intrusive_ptr<UniTensor_base> astype(const unsigned int &dtype) const {
-      DenseUniTensor *tmp = this->clone_meta();
+      boost::intrusive_ptr<DenseUniTensor> tmp = this->clone_meta();
       tmp->_block = this->_block.astype(dtype);
-      boost::intrusive_ptr<UniTensor_base> out(tmp);
       return tmp;
     }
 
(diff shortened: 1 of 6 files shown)


ianmccul:
Looks good, matches almost exactly the PR I was about to post...

This looks like it covers all of the `clone_meta()` leak class cleanly. I do not see any remaining raw-pointer `clone_meta()` call sites.

The remaining ownership issue is adjacent but broader: several functions still allocate a concrete UniTensor result with raw `new`, do nontrivial work on the raw pointer, and only wrap it in `boost::intrusive_ptr<UniTensor_base>` at the end. That has the same exception-safety shape as the old `clone_meta()` bug, but even if it isn't an acual leak, it is still unnecessarily fragile code.

Known follow-up candidates:

- `DenseUniTensor::contract()` in `src/DenseUniTensor.cpp`
- `BlockUniTensor::contract()` in `src/BlockUniTensor.cpp`
- `BlockFermionicUniTensor::contract()` in `src/BlockFermionicUniTensor.cpp`
- rank-zero `Trace()` conversions that allocate a `DenseUniTensor` after tracing a block/block-fermionic tensor:
  - `BlockUniTensor::Trace(...)` overloads in `include/UniTensor.hpp`
  - `BlockFermionicUniTensor::Trace(...)` overloads in `include/UniTensor.hpp`
  - `UniTensor::Trace_(...)` rank-zero conversion paths in `include/UniTensor.hpp`

Suggested cleanup pattern:

```cpp
boost::intrusive_ptr<DenseUniTensor> tmp(new DenseUniTensor());
// mutate tmp, run validation, fill blocks/labels/bonds
return tmp;
```
The important part is that ownership starts immediately at allocation.

That is all a follow-up PR.  This PR looks good.

ianmccul (approved):

PR #1044, opened 2026-07-11
ci(gpu): native self-hosted GPU build + test workflow

yingjerkao:
## Problem
No GPU CI exists (origin issue #538). `ci-cmake_tests.yml` is CPU-only; the old `gpu-smoke-test` (`main.yml`) targeted a non-existent `v100` runner and shelled out to Docker (not installed), so it never ran a single job. And after #1004 `gpu_test_main` didn't even **compile**: its host-compiled `.cpp` TUs include `Type.hpp` → `<cuda/std/complex>`, but `cytnx` adds the CUDA toolkit + CCCL includes **PRIVATE**, so they never reached the test target (`fatal error: cuda/std/complex: No such file or directory`).

## Fix
- **`tests/gpu/CMakeLists.txt`** — give `gpu_test_main` the CUDA toolkit + CCCL include dirs (the compile fix); tag its cases with the `gpu` ctest label; enumerate `DISCOVERY_MODE PRE_TEST` so a CUDA build links even on a driver-less machine.
- **`CytnxBKNDCMakeLists.cmake`** — expose the detected CCCL dir as `CYTNX_CCCL_INCLUDE_DIR` (single source of truth, reused by the test target rather than re-detected).
- **`CMakePresets.json`** — add a `gpu-only` test preset (filters on the `gpu` label).
- **`.github/workflows/ci-gpu_tests.yml`** — native build on the self-hosted GPU runner using the box's CUDA / cuTENSOR / cuQuantum (roots come from repo `vars.CUTENSOR_ROOT` / `vars.CUQUANTUM_ROOT`, already configured); builds `gpu_test_main` and runs `ctest -L gpu`. Manual dispatch + **same-repo** PRs only — fork PRs never execute on self-hosted hardware.
- Remove the dead `gpu-smoke-test` workflow (`main.yml`).

The `gpu` label + `PRE_TEST` groundwork is adapted from the draft #1023 (which targeted a GitHub-hosted runner and, being unrun, still carried the CCCL compile blocker). This supersedes #1023's direction — suggest closing it once this lands.

## Testing
On the runner box (2× RTX 4070 Ti SUPER, sm_89, CUDA 13.0, cuTENSOR 2.0, cuQuantum):
- `cmake` configure + `cmake --build --target gpu_test_main` succeed — CUDA lib + test binary, no `cuda/std/complex` error.
- `ctest -L gpu` discovers **788** GPU-labeled tests; `AccessorTest` passes **15/15** on-GPU.
- This PR's own `pull_request` run exercises the workflow on the self-hosted runner end-to-end.

### Known runner-env caveat (not a blocker)
The runner's `cuquantum` conda env is CUDA-11-built, so linking warns `libcublas.so.11 … may conflict with libcublas.so.13`. Core tensor tests are unaffected; updating that env to a CUDA-13 cuQuantum build clears the warning.

Toward #538.


.github/workflows/ci-gpu_tests.yml
```diff
@@ -0,0 +1,85 @@
...
+          nvcc --version
+
+      - name: Configure (CUDA Release, C++ tests, no Python)
+        run: >-
+          cmake -S . -B "$BUILD_DIR"
+          -DCMAKE_BUILD_TYPE=Release
+          -DUSE_CUDA=ON -DUSE_CUTENSOR=ON -DUSE_CUQUANTUM=ON
+          -DRUN_TESTS=ON -DBUILD_PYTHON=OFF
+          -DCMAKE_CUDA_ARCHITECTURES="$GPU_CUDA_ARCH"
+
```
IvanaGyro, line 83: Consider to use preset debug-openblas-cuda or add a new test preset like #1023

.github/workflows/ci-gpu_tests.yml
```diff
@@ -0,0 +1,85 @@
...
+    timeout-minutes: 120
+    defaults:
+      run:
+        shell: bash -eo pipefail {0}
+    env:
+      # cuTENSOR / cuQuantum install roots on the runner. Kept as repository
+      # Actions variables so this file carries no machine-specific paths; the
+      # FindCUTENSOR / FindCUQUANTUM modules require both to be set.
+      CUTENSOR_ROOT: ${{ vars.CUTENSOR_ROOT }}
+      CUQUANTUM_ROOT: ${{ vars.CUQUANTUM_ROOT }}
```
IvanaGyro, line 46: If we can control the dependency versions like #1023 will be better. Because that will be the versions we will release.

IvanaGyro (approved):

IvanaGyro:
There are conflicts that need to be resolved.

IvanaGyro (approved):
Will this workflow consider to use predefined preset?

tests/gpu/CMakeLists.txt
```diff
@@ -57,10 +57,29 @@ target_link_libraries(
 )
 target_link_libraries(gpu_test_main cytnx)
 #target_link_libraries(test_main PUBLIC "-lgcov --coverage")
+
+# gpu_test_main's .cpp TUs are host-compiled (g++) but include cytnx public
+# headers (Type.hpp) that pull <cuda/std/complex>. cytnx adds the CUDA toolkit
+# and CCCL include dirs PRIVATEly, so they do not propagate here through
+# linking; add them explicitly, otherwise the host compile fails with
+# "fatal error: cuda/std/complex: No such file or directory".
+target_include_directories(gpu_test_main PRIVATE ${CUDAToolkit_INCLUDE_DIRS})
```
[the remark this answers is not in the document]
yingjerkao: Addressed in e9e692cb — wrapped the `CUDAToolkit_INCLUDE_DIRS` include in an `if()` guard (mirroring the `CYTNX_CCCL_INCLUDE_DIR` guard right below it), so an empty/undefined variable no longer expands to an empty `target_include_directories()` call and fail configuration.

.github/workflows/ci-gpu_tests.yml
```diff
@@ -0,0 +1,85 @@
...
+# toolchain install step.
+#
+# Runs on manual dispatch and on same-repo pull requests. Fork pull requests
+# are NOT executed on the self-hosted runner: on a public repo that would let a
+# fork run arbitrary code on our hardware. The job-level `if` enforces this.
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - master
```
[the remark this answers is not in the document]
yingjerkao: Good catch — addressed in e9e692cb by adding a `push:` trigger for `master`. Code merged from a fork PR (which the `pull_request` run deliberately skips) now runs the GPU suite on the trusted merged revision. A `push` event is never a fork, so the existing job `if` guard runs it unconditionally while still skipping fork `pull_request` events on the self-hosted runner.

.github/workflows/ci-gpu_tests.yml
yingjerkao: Thanks. This job is intentionally a *Release* native build against the runner's own system-installed CUDA/cuTENSOR/cuQuantum (no conda/PyPI toolchain step), which doesn't map cleanly onto the current presets — `debug-openblas-cuda` is Debug+ASan, and it also keeps the build config close to what a from-source user would run on their own GPU box. I agree a dedicated CUDA *test* preset like #1023 is the right long-term home for these flags; moving this workflow onto it once #1023 lands would be a clean follow-up so the config lives in one place.

.github/workflows/ci-gpu_tests.yml
yingjerkao: Agreed, and this pairs with the preset point above — pinning the cuTENSOR/cuQuantum versions to the released set fits naturally once this moves onto the #1023 test preset, rather than hard-coding them in the workflow here. Reasonable to handle both together as the #1023 follow-up.

tests/gpu/CMakeLists.txt
```diff
@@ -63,6 +63,19 @@ target_link_libraries(
...
+# headers (Type.hpp) that pull <cuda/std/complex>. cytnx adds the CUDA toolkit
+# and CCCL include dirs PRIVATEly, so they do not propagate here through
+# linking; add them explicitly, otherwise the host compile fails with
+# "fatal error: cuda/std/complex: No such file or directory".
+if(CUDAToolkit_INCLUDE_DIRS)
+  target_include_directories(gpu_test_main PRIVATE ${CUDAToolkit_INCLUDE_DIRS})
+endif()
+if(CYTNX_CCCL_INCLUDE_DIR)
+  target_include_directories(gpu_test_main PRIVATE ${CYTNX_CCCL_INCLUDE_DIR})
+endif()
```
IvanaGyro, line 77: If the public headers need `<cuda/std/complex>`, the include path of that header should be public for cytnx.

PR #1054, opened 2026-07-15
fix(UniTensor): eliminate overlapping memcpy in combineBonds

### LINKED ISSUE #1052: ASan memcpy-param-overlap in BlockUniTensor::permute_ (via combineBond), CombineBondForceDoesNotRewriteUserHeldBond
## Summary

`BlockUniTensorTest.CombineBondForceDoesNotRewriteUserHeldBond` (`tests/BlockUniTensor_test.cpp:1719`) fails under AddressSanitizer with a `memcpy-param-overlap` abort. The call chain is:

```
UniTensor::combineBond_ (include/UniTensor.hpp:4902)
  -> BlockUniTensor::combineBond (src/BlockUniTensor.cpp:2027)
  -> BlockUniTensor::combineBonds (src/BlockUniTensor.cpp:1932)
  -> BlockUniTensor::permute_ (src/BlockUniTensor.cpp:678)
  -> cytnx::vec_map<unsigned long> (src/utils/vec_map.cpp:20)
```

`vec_map` allocates a `std::vector<unsigned long>` and then does a `memcpy` where the source and destination regions overlap, which ASan flags as UB (`memcpy` requires non-overlapping buffers; overlapping copies must use `memmove`).

## Reproduction

```sh
cmake --preset debug-openblas-cpu
cmake --build --preset debug-openblas-cpu --target test_main
ctest --test-dir build/debug-openblas-cpu -R BlockUniTensorTest.CombineBondForceDoesNotRewriteUserHeldBond --output-on-failure
```

## Scope note

Confirmed pre-existing and unrelated to any in-flight pybind binding work: it reproduces on a branch whose diff touches only `pybind/`, `pytests/`, `tools/generate_stubs.py`, and the committed `.pyi` stubs — no changes under `src/`, `include/`, or `tests/`. Also confirmed not a resource-contention flake: it fails reliably in isolation (`--parallel 1`, single test), unlike a few unrelated `Lanczos_Ut` tests that failed alongside it under 4-way parallel ctest but passed individually.

---
Found by Claude (Anthropic's Claude Code) while running the full CPU test suite as part of unrelated pybind11 stubtest work on this repository.

IvanaGyro:
## Problem

Fixes #1052.

`BlockUniTensorTest.CombineBondForceDoesNotRewriteUserHeldBond` aborts under
ASan with `memcpy-param-overlap`. The issue's own stack trace points at
`vec_map` (`src/utils/vec_map.cpp`) as the allocation site of the corrupted
buffer, but that is only where ASan's shadow memory was allocated — `vec_map`
itself contains no `memcpy` at all and is not the bug.

Running the test directly (rather than relying on the issue's captured trace)
gives the real fault stack:

```
==ERROR: AddressSanitizer: memcpy-param-overlap
    #1 cytnx::BlockUniTensor::combineBonds(...)  src/BlockUniTensor.cpp:1996
    #2 cytnx::BlockUniTensor::combineBond(...)   src/BlockUniTensor.cpp:2027
    #3 cytnx::UniTensor::combineBond_(...)       include/UniTensor.hpp:4902
```

`combineBonds` has sites in three of the four `UniTensor` backends that
left-shift a `std::vector` into itself via `memcpy`, which is undefined
behavior whenever the shift distance is smaller than the region being moved
(source and destination byte ranges overlap):

- **`BlockUniTensor::combineBonds`** (`src/BlockUniTensor.cpp`) — two sites:
  1. **`_inner_to_outer_idx` tail shift** — drops `indicators.size()-1`
     elements starting at `idor+1` and shifts the tail left. It overlaps
     whenever the remaining tail is at least twice the shift distance; the
     existing 2-bond regression test already hits this exactly.
  2. **`cb_stride` shift-by-one** — latent (only overlaps once 3+ bonds are
     combined at once, which no existing test exercised), but the same UB
     class in the same function, with a commented-out backward loop sitting
     next to it as an abandoned alternative.
- **`BlockFermionicUniTensor::combineBonds`** (`src/BlockFermionicUniTensor.cpp`)
  has the identical two sites (its own comment marks it "a copy from
  BlockUniTensor").
- **`DenseUniTensor::combineBonds`** (`src/DenseUniTensor.cpp`) has a
  structurally identical `new_shape` tail shift — same overlap condition,
  same variable names (`idor`, `indicators`). Combining the first two of
  four adjacent bonds (`idor=0`, `indicators.size()=2`) reproduces the exact
  same overlap geometry as the `BlockUniTensor` case.

I grepped all four `combineBonds` implementations for `memcpy` and confirmed
these are the only overlapping sites; `Tensor::combineBonds` (dense array
reshape only, no bond bookkeeping) uses no such shift.

## Fix

Every site is replaced with well-defined standard-library equivalents that
perform the exact same left-shift, across `BlockUniTensor.cpp`,
`BlockFermionicUniTensor.cpp`, and `DenseUniTensor.cpp`:

- `cb_stride` shift → `std::copy(cb_stride.begin() + 1, cb_stride.end(),
  cb_stride.begin())`, well-defined for a left-shift since `std::copy`
  iterates forward, writing to already-read positions.
- `_inner_to_outer_idx` / `new_shape` tail shift → `vector::erase(begin+idor+1,
  begin+idor+indicators.size())`, which removes exactly the same elements
  and leaves the vector at the correct final length, so the trailing
  `resize(this->rank())` is no longer needed.

**This is a memory-safety-only fix.** Every replacement computes the exact
same result the `memcpy` calls intended; `combineBonds`' numerical output is
unchanged across all three backends, only the undefined behavior is removed.

`BlockFermionicUniTensor::combineBonds(vector, force)` is fixed
for source hygiene/consistency with `BlockUniTensor`, but is currently
**unreachable through any public API path** — it starts with an
unconditional `cytnx_error_msg(true, "...not implemented yet.")` behind a
`TODOfermion: signflips need to be included!!!` comment. Lifting that guard
to exercise the fix live would be a functional change to fermionic
sign-flip handling, out of scope here; a new test
(`CombineBondsStillUnimplemented`) pins the current throw so this scope
statement stays true.

## Testing

New tests, each verified to fail pre-fix and pass post-fix:

- `BlockUniTensorTest.CombineBondsThreeBondsPreservesDataAndDimension` — a
  3-bond combine that exercises the `cb_stride` site. Verified pre-fix: aborts
  under ASan at `BlockUniTensor.cpp:1964` with `memcpy-param-overlap`. Checks
  the combined bond's dimension against the independently computed product of
  the three input dimensions, and checks block values against the pre-combine
  block reshaped by hand (row-major merge of its first three axes, matching
  how `Bond_impl::force_combineBond_` orders the combined qnum index) —
  independent of the bookkeeping this PR touches.
- `BlockFermionicUniTensorTest.CombineBondsStillUnimplemented` — documents
  the current unreachability (see above).
- `DenseUniTensorTest.CombineBondsFourBondsPreservesDataAndShape` — a 4-bond
  dense tensor combining the first two adjacent bonds. Verified pre-fix:
  aborts under ASan at `DenseUniTensor.cpp:855` with `memcpy-param-overlap`
  (same failure signature as the `BlockUniTensor` case). Checks the combined
  shape and values against an independently computed reshape of the original
  block, not just absence of a crash.
- The pre-existing `BlockUniTensorTest.CombineBondForceDoesNotRewriteUserHeldBond`
  re-confirmed pre-fix: aborts under ASan at `BlockUniTensor.cpp:1996` with
  `memcpy-param-overlap`; passes post-fix.

Local results, both required CPU debug presets:

| Preset | Suite | Result |
|---|---|---|
| `debug-openblas-cpu` | `ctest` (`test_main`) | 1771/1771 passed |
| `debug-mkl-cpu` | `ctest` (`test_main`) | 1805/1805 passed |

No GPU/CUDA code touched (no `#ifdef UNI_GPU`, no `src/backend/linalg_internal_gpu/`),
so the CUDA compile-check gate does not apply. No `cytnx/`, `pybind/`, or
`pytests/` changes, so `pytest pytests/` does not apply.

Opened as a draft while the fix scope (`BlockUniTensor` +
`BlockFermionicUniTensor` + `DenseUniTensor`) gets a look.

### DIFF
--- a/src/BlockFermionicUniTensor.cpp
+++ b/src/BlockFermionicUniTensor.cpp
@@ -2704,10 +2704,7 @@ namespace cytnx {
 
     // reshape each blocks, and update_inner_to_outer_idx:
     // process stride:
-    memcpy(&cb_stride[0], &cb_stride[1], sizeof(cytnx_uint64) * (cb_stride.size() - 1));
-    // for(int i=cb_stride.size()-2;i>=0;i--){
-    //     cb_stride[i] = cb_stride[i+1];
-    // }
+    std::copy(cb_stride.begin() + 1, cb_stride.end(), cb_stride.begin());
     cb_stride.back() = 1;
     for (int i = cb_stride.size() - 2; i >= 0; i--) {
       cb_stride[i] *= cb_stride[i + 1];
@@ -2735,13 +2732,8 @@ namespace cytnx {
       for (int i = idor + 1; i < idor + indicators.size(); i++) {
         this->_inner_to_outer_idx[b][idor] += this->_inner_to_outer_idx[b][i] * cb_stride[i - idor];
       }
-      if (idor + indicators.size() < this->_inner_to_outer_idx[b].size()) {
-        memcpy(
-          &this->_inner_to_outer_idx[b][idor + 1],
-          &this->_inner_to_outer_idx[b][idor + indicators.size()],
-          sizeof(cytnx_uint64) * (this->_inner_to_outer_idx[b].size() - idor - indicators.size()));
-      }
-      this->_inner_to_outer_idx[b].resize(this->rank());
+      auto &itoi = this->_inner_to_outer_idx[b];
+      itoi.erase(itoi.begin() + idor + 1, itoi.begin() + idor + indicators.size());
     }
 
     // change rowrank:
--- a/src/BlockUniTensor.cpp
+++ b/src/BlockUniTensor.cpp
@@ -1961,10 +1961,7 @@ namespace cytnx {
 
     // reshape each blocks, and update_inner_to_outer_idx:
     // process stride:
-    memcpy(&cb_stride[0], &cb_stride[1], sizeof(cytnx_uint64) * (cb_stride.size() - 1));
-    // for(int i=cb_stride.size()-2;i>=0;i--){
-    //     cb_stride[i] = cb_stride[i+1];
-    // }
+    std::copy(cb_stride.begin() + 1, cb_stride.end(), cb_stride.begin());
     cb_stride.back() = 1;
     for (int i = cb_stride.size() - 2; i >= 0; i--) {
       cb_stride[i] *= cb_stride[i + 1];
@@ -1992,13 +1989,8 @@ namespace cytnx {
       for (int i = idor + 1; i < idor + indicators.size(); i++) {
         this->_inner_to_outer_idx[b][idor] += this->_inner_to_outer_idx[b][i] * cb_stride[i - idor];
       }
-      if (idor + indicators.size() < this->_inner_to_outer_idx[b].size()) {
-        memcpy(
-          &this->_inner_to_outer_idx[b][idor + 1],
-          &this->_inner_to_outer_idx[b][idor + indicators.size()],
-          sizeof(cytnx_uint64) * (this->_inner_to_outer_idx[b].size() - idor - indicators.size()));
-      }
-      this->_inner_to_outer_idx[b].resize(this->rank());
+      auto &itoi = this->_inner_to_outer_idx[b];
+      itoi.erase(itoi.begin() + idor + 1, itoi.begin() + idor + indicators.size());
     }
 
     // change rowrank:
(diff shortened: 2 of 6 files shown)


yingjerkao (approved):
LGTM

Issue #963, opened 2026-07-01
Decide a deliberate strategy for the macOS deployment-target floor and the Linux GCC/libstdc++ floor


IvanaGyro:
## Summary

Two related "how low can our minimum supported toolchain/OS go" problems are currently solved reactively rather than by deliberate choice:

- **macOS deployment target.** `tools/cibuildwheel_before_all_macos.sh` computes `MACOSX_DEPLOYMENT_TARGET` as the *max* `minos` (minimum-OS load command) found across Homebrew's bottled `arpack`/`boost`/`openblas`/`libomp` dylibs and their transitive dependencies (falling back to `10.9` if that computation hasn't run yet — see `pyproject.toml:115`). Homebrew bottles are built by Homebrew's own CI against whatever deployment target Homebrew currently targets for that formula, which trends upward release over release. So Cytnx's actual minimum supported macOS version silently rises over time as a side effect of Homebrew's build policy, not a decision Cytnx makes. `.github/workflows/release_pypi.yml:120-123` already has a comment flagging this: "newer images because Homebrew dylib minos can force a higher wheel deployment target and break delocate compatibility for lower targets."
- **Linux GCC/libstdc++ floor.** PR #961 requires GCC ≥ 13 (for `std::format`) when `BUILD_PYTHON=ON`. Linux wheels are built against the `manylinux_2_28` image (`pyproject.toml`).

## Research: macOS

**Primary recommendation — stop sourcing wheel-shipped dependencies from Homebrew.** The reactive "probe whatever Homebrew happens to bottle" approach can't get lower than Homebrew's current floor by construction. The fix used across the scientific Python ecosystem (numpy, scipy, etc.) for exactly this problem is to source C/C++ dependencies from **conda-forge** instead: its `clang_osx-64`/`clang_osx-arm64` compiler packages bundle a pinned, deliberately old macOS SDK plus a cross `cctools`/`ld64`, so packages built against them (and conda-forge's own prebuilt `arpack`/`boost`/`openblas`/`libomp`) carry a controlled, low `minos` regardless of which OS the CI runner happens to be on. This is likely the cheapest available fix — it replaces the current from-Homebrew probing script with pointing `CMAKE_PREFIX_PATH` at conda-forge-installed packages, with no new from-source build recipes to maintain.

[...]

PR #1057, opened 2026-07-15
build(wheels): source arpack/openblas/boost from conda-forge (Linux + macOS)

IvanaGyro:
## Problem

Two independent problems traced back to where the Linux and macOS wheel builds source their native dependencies from:

- **Linux**: `tools/cibuildwheel_before_all.sh` installs `boost-devel openblas-devel arpack-devel` from EPEL via `dnf` inside the `manylinux_2_28` container. EPEL's `arpack-devel` links whichever OpenBLAS `dnf` resolves — a pthreads-threaded build with its own independent thread pool. HPTT and cytnx's own code use OpenMP (`-fopenmp`, see `CytnxBKNDCMakeLists.cmake`), so a BLAS call made from inside an OpenMP region oversubscribes the CPU (two independent thread pools competing).
- **macOS** (#963): `tools/cibuildwheel_before_all_macos.sh` installs `arpack`/`boost`/`openblas`/`libomp` from Homebrew and computes `MACOSX_DEPLOYMENT_TARGET` from their `minos`. Homebrew's bottles are built against whatever deployment target Homebrew itself currently targets, which trends upward release over release — Cytnx's actual minimum supported macOS version was rising as a side effect of Homebrew's own build policy, not a decision made here. Homebrew's OpenBLAS bottle is also pthreads-only, same issue as Linux.

## Fix

Both scripts now provision a self-contained `micromamba` environment from **conda-forge** instead:

- `openblas=*=*openmp*` selects the OpenMP-threaded OpenBLAS build.
- `arpack=*=nompi*` avoids vendoring an MPI runtime (conda-forge's `arpack` also has `mpich`/`openmpi` variants).
- `liblapacke` / `libboost-headers` replace the corresponding EPEL/Homebrew packages.
- On macOS, `llvm-openmp` replaces Homebrew's `libomp`, and the **existing** dylib-`minos`-probing logic in `cibuildwheel_before_all_macos.sh` is kept unchanged, just repointed at the new conda-forge prefix — `MACOSX_DEPLOYMENT_TARGET` keeps tracking whatever floor the actual vendored dylibs require (conda-forge's `clang_osx-64`/`clang_osx-arm64` toolchain targets a pinned, deliberately old SDK) instead of a value guessed in advance.

A second, small commit pins `ci-cmake_tests.yml`'s own `mamba install` to `libopenblas=*=*openmp*` explicitly. That workflow already pins `_openmp_mutex=*=*_llvm` (so the environment settles on one OpenMP runtime), but `blas=*=openblas` alone doesn't guarantee the solver picks the OpenMP-threaded build over the pthreads one — both are solver-compatible with `_openmp_mutex=llvm` present, since a pthreads OpenBLAS doesn't touch OpenMP at all.

## Threading verification (per repo convention: measure, don't assume)

Sandbox constraints: no docker registry access (egress policy) and no macOS host, so a real `cibuildwheel` run wasn't possible locally. What *was* verified directly, using a real `micromamba`-created environment with this exact package set and real `cmake`/`g++` on the sandbox host:

- `find_library(arpack)`, `find_package(BLAS BLA_VENDOR OpenBLAS)`, `find_package(LAPACK)`, `find_package(LAPACKE)`, and `find_package(Boost)` (forced to MODULE mode, since manylinux has no competing system Boost config) all resolve correctly against the conda-forge prefix via `CMAKE_PREFIX_PATH`.
- `libblas.so.3`/`liblapack.so.3`/`liblapacke.so.3`/`libopenblas.so.0` are confirmed to all symlink to the *same* physical `libopenblasp-r0.3.33.so` — one OpenBLAS copy, not two.
- conda-forge's `openmp`-variant OpenBLAS declares `NEEDED libgomp.so.1`, but the environment's `_openmp_mutex` (llvm variant, conda-forge's current default for this package combination) makes `libgomp.so.1` a **symlink to `libomp.so`** (LLVM's runtime). Verified `libomp.so` genuinely exports the full versioned `GOMP_1.0`–`GOMP_5.0.1` symbol set GCC's `-fopenmp` code needs (`nm -D`), so GCC-compiled cytnx code and OpenBLAS end up sharing one thread pool rather than two.
- Compiled and **ran** a real program linking `arpack` + LAPACKE + Boost + OpenMP, calling `LAPACKE_dgeqrf` from inside an 8-way `#pragma omp parallel for` region — correct output, correct thread count, no crashes, no missing-symbol errors.
- `ci-cmake_tests.yml` already pins `_openmp_mutex=*=*_llvm` for its own mamba-based CPU environment — i.e. this exact OpenMP-runtime combination is already proven in this repo's production CI, not a speculative first use.

The macOS script itself couldn't be executed locally (no Mac available in the sandbox), so it relied on reusing the already-proven Homebrew-era probing logic verbatim, repointed at a new prefix. This PR's own CI caught three real bugs that local verification couldn't have found: the ARM64 micromamba download picking an x86_64 binary, a macOS CMake export-safety error from the deps prefix living inside the checkout, and cibuildwheel's `environment` table not expanding `${VAR:-default}` shell syntax the way a real shell does. All three are fixed (see commit history) and **all four `BuildWheel` legs (`ubuntu-24.04`, `ubuntu-24.04-arm`, `macos-14`, `macos-15-intel`) are now green**, along with every other check on the PR.

## Testing

- `pre-commit run` on all changed files (end-of-file-fixer, trailing-whitespace, large-file check; no C++ touched).
- YAML/TOML syntax validated (`yaml.safe_load`, `tomllib.load`).
- Threading/linkage verification described above.
- This PR's own CI matrix (`release_pypi.yml`'s `BuildWheel-*` jobs, `conda_build.yml`'s `BuildAndTest-*` jobs) — all green as of the latest commit.

---
_Generated by [Claude Code](https://claude.ai/code/session_015cf5bwKYeECGnRN6y9amwk)_


tools/cibuildwheel_before_all.sh
```diff
@@ -1,9 +1,31 @@
...
+# OpenMP code use (-fopenmp, see CytnxBKNDCMakeLists.cmake), so a BLAS call
+# made from inside an OpenMP region oversubscribes the CPU. conda-forge
+# publishes an explicit "openmp" build variant of OpenBLAS that shares an
+# OpenMP runtime's thread pool with the caller instead, closing that gap.
+# ci-cmake_tests.yml's own mamba-based dependency install already relies on
+# this same conda-forge OpenMP runtime for the native test build.
 if command -v dnf >/dev/null 2>&1; then
-    dnf install -y boost-devel openblas-devel arpack-devel ccache
-# musllinux_1_2 images are Alpine-based, so use apk there.
+    dnf install -y ccache curl
```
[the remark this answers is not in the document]
IvanaGyro: Applied in 451997a.

---

tools/cibuildwheel_before_all.sh
```diff
@@ -1,9 +1,31 @@
...
+
+    curl -Ls "https://micro.mamba.pm/api/micromamba/linux-64/latest" | tar -xj -C /usr/local bin/micromamba
```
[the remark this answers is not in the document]
IvanaGyro: Applied in 451997a.

---

tools/cibuildwheel_before_all_macos.sh
```diff
@@ -2,64 +2,72 @@
...
+arch="$(uname -m)"
+if [[ "${arch}" == "arm64" ]]; then
+  conda_subdir="osx-arm64"
+else
+  conda_subdir="osx-64"
+fi
 
-brew update
-brew install "${install_packages[@]}"
+curl -Ls "https://micro.mamba.pm/api/micromamba/${conda_subdir}/latest" | tar -xj bin/micromamba
```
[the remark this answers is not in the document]
IvanaGyro: Applied in 451997a.

---

tools/cibuildwheel_before_all.sh
```diff
@@ -1,9 +1,37 @@
...
+    MAMBA_ROOT_PREFIX=/opt/micromamba /usr/local/bin/micromamba create -y -p "${CYTNX_DEPS_PREFIX}" -c conda-forge \
+      "openblas=*=*openmp*" \
+      liblapacke \
+      "arpack=*=nompi*" \
+      libboost-headers
+
+# musllinux_1_2 images are Alpine-based, so use apk there. conda-forge
+# publishes no musl builds, so this path keeps installing from Alpine's own
+# package index; the OpenBLAS-threading fix above only applies to the
+# manylinux (dnf) wheels.
```
IvanaGyro: We should build our openblas with openmp. Here is the commands for reference. Please verify if each line is needed and correct.

---

OpenBLAS documents that `OMP_NUM_THREADS` controls OpenMP builds, while `OPENBLAS_NUM_THREADS` is for non-OpenMP builds. ([[openmathlib.github.io](https://www.openmathlib.org/OpenBLAS/docs/runtime_variables/?utm_source=chatgpt.com)][2])

Below is a full recipe that:

1. uses a `musllinux_1_2_x86_64` container,
2. installs Alpine `arpack-dev`,
3. builds **OpenMP OpenBLAS**,
4. changes the OpenBLAS SONAME to Alpine-compatible `libopenblas.so.3`,
5. verifies that Alpine `libarpack.so` resolves to your custom OpenMP OpenBLAS.

Alpine’s OpenBLAS package provides `libopenblas.so.3`, so the SONAME step is important if you want Alpine ARPACK to use your replacement cleanly. ([[Alpine Linux Packages](https://pkgs.alpinelinux.org/package/edge/community/x86/openblas?utm_source=chatgpt.com)][3])

```bash
docker run --rm -it \
  -v "$PWD":/io \
  quay.io/pypa/musllinux_1_2_x86_64 \
  /bin/sh
```

Inside the container:

```bash
set -eux

apk add --no-cache \
  build-base \
  gcc \
  g++ \
  gfortran \
  libgomp \
  linux-headers \
  make \
  cmake \
  ninja \
  git \
  perl \
  bash \
  pkgconf \
  patchelf \
  pax-utils \
  arpack-dev \
  openblas-dev

export PREFIX=/opt/openblas-omp
export OPENBLAS_TAG=v0.3.33
export OPENBLAS_MAX_THREADS="256"
export JOBS="$(nproc)"

cd /io
rm -rf OpenBLAS
git clone --depth 1 --branch "${OPENBLAS_TAG}" https://github.com/OpenMathLib/OpenBLAS.git
cd OpenBLAS

make -j"${JOBS}" \
  DYNAMIC_ARCH=1 \
  USE_OPENMP=1 \
  NUM_THREADS="${OPENBLAS_MAX_THREADS}" \
  NO_STATIC=1

make install PREFIX="${PREFIX}"

# Find the real installed OpenBLAS shared object.
OPENBLAS_REAL="$(readlink -f "${PREFIX}/lib/libopenblas.so")"

# Make its SONAME match Alpine's BLAS ABI name expected by arpack-dev.
patchelf --set-soname libopenblas.so.3 "${OPENBLAS_REAL}"

# Ensure all useful linker/runtime names exist in the custom prefix.
ln -sf "$(basename "${OPENBLAS_REAL}")" "${PREFIX}/lib/libopenblas.so.3"
ln -sf "$(basename "${OPENBLAS_REAL}")" "${PREFIX}/lib/libopenblas.so"
ln -sf "$(basename "${OPENBLAS_REAL}")" "${PREFIX}/lib/libblas.so"
ln -sf "$(basename "${OPENBLAS_REAL}")" "${PREFIX}/lib/liblapack.so"

# Verify custom OpenBLAS is OpenMP-linked and has the expected SONAME.
readelf -d "${OPENBLAS_REAL}" | grep -E 'SONAME|NEEDED'
```

Expected important lines:

```text
SONAME               Library soname: [libopenblas.so.3]
NEEDED               Shared library: [libgomp.so.1]
```

Now verify Alpine ARPACK resolves to your custom OpenBLAS:

```bash
export LD_LIBRARY_PATH="${PREFIX}/lib:/usr/lib"

readelf -d /usr/lib/libarpack.so | grep -E 'NEEDED|RPATH|RUNPATH'
ldd /usr/lib/libarpack.so
```

You want `ldd /usr/lib/libarpack.so` to show:

```text
libopenblas.so.3 => /opt/openblas-omp/lib/libopenblas.so.3
```

Also verify with `scanelf`:

```bash
scanelf -n /usr/lib/libarpack.so
scanelf -n "${PREFIX}/lib/libopenblas.so.3"
```



[1]: https://github.com/OpenMathLib/OpenBLAS/blob/develop/USAGE.md?utm_source=chatgpt.com "OpenBLAS/USAGE.md at develop"
[2]: https://www.openmathlib.org/OpenBLAS/docs/runtime_variables/?utm_source=chatgpt.com "Runtime variables - OpenBLAS - openmathlib.github.io"
[3]: https://pkgs.alpinelinux.org/package/edge/community/x86/openblas?utm_source=chatgpt.com "openblas - Alpine Linux packages"

pyproject.toml
```diff
@@ -149,11 +149,30 @@ before-build = "python ./tools/cibuildwheel_before_build.py"
...
-environment = { CC = "clang", CXX = "clang++", CMAKE_PREFIX_PATH = "$(brew --prefix arpack):$(brew --prefix boost):$(brew --prefix libomp):$(brew --prefix openblas)", PATH = "$(brew --prefix ccache)/libexec:$PATH", MACOSX_DEPLOYMENT_TARGET = "$(python -c 'from pathlib import Path; p=Path(\".cibw_macos_deployment_target.txt\"); print(p.read_text().strip() if p.exists() else \"10.9\")')", CCACHE_DEBUG = "true", CCACHE_DEBUGLEVEL = "1", CCACHE_LOGFILE = "/tmp/cytnx-ccache.log" }
+environment = { CC = "clang", CXX = "clang++", CMAKE_PREFIX_PATH = "$(python -c 'from pathlib import Path; p=Path(\".cibw_macos_deps_prefix.txt\"); print(p.read_text().strip() if p.exists() else \"/tmp/cytnx-deps\")')", PATH = "$(python -c 'from pathlib import Path; p=Path(\".cibw_macos_deps_prefix.txt\"); print(p.read_text().strip() if p.exists() else \"/tmp/cytnx-deps\")')/bin:$PATH", MACOSX_DEPLOYMENT_TARGET = "$(python -c 'from pathlib import Path; p=Path(\".cibw_macos_deployment_target.txt\"); print(p.read_text().strip() if p.exists() else \"10.9\")')", CCACHE_DEBUG = "true", CCACHE_DEBUGLEVEL = "1", CCACHE_LOGFILE = "/tmp/cytnx-ccache.log" }
```
IvanaGyro: If we rely on conda ecosystem, we can just set MACOSX_DEPLOYMENT_TARGET by conda's document. 10.9 and 11.0. And please consider to get package paths by invoking conda commands or activating conda env before building.

tools/cibuildwheel_before_all.sh
IvanaGyro: Implemented in 58b787d, adapted from your recipe with one change: instead of hardcoding `libopenblas.so.3` I discover the soname dynamically from `libarpack.so`'s own NEEDED entry via `readelf -d`, so this doesn't silently break if Alpine ever bumps that soname. Also trimmed the apk package list to what a plain `make`-based OpenBLAS build actually needs (dropped `cmake`/`ninja`/`bash`/`pkgconf`/`pax-utils` — OpenBLAS's build system doesn't use them; kept `perl` since its LAPACK-netlib subdirectory does).

Verified the core premise directly on this host: built the conda-forge openmp-variant OpenBLAS I already use for manylinux/macOS, called `openblas_get_num_threads()` under different env vars — `OMP_NUM_THREADS` controls it, `OPENBLAS_NUM_THREADS` alone is ignored, matching the docs you cited exactly.

Not verified end-to-end: no musllinux/docker access in my sandbox, so the actual Alpine build-from-source + patchelf + symlink sequence hasn't run anywhere. Relying on this PR's own musllinux `BuildWheel` CI legs to validate it.

---

pyproject.toml
IvanaGyro: On paths: implemented in 1d4fed9 — switched to a fixed `/tmp/cytnx-deps` prefix instead of deriving it from `$TMPDIR`, so `pyproject.toml` references it as a plain literal with no discovery step at all.

On the deployment-target values: checked today's `conda-forge-pinning` package directly (`c_stdlib_version`, which drives `MACOSX_DEPLOYMENT_TARGET` via the `macosx_deployment_target` `c_stdlib` selector) — it's currently **11.0 uniformly** for both osx-64 and osx-arm64, not a 10.9/11.0 split. I found where the "10.9" is coming from: it's baked into `macos_machine`'s x86_64 target-triple string (`x86_64-apple-darwin13.4.0` — Darwin 13.4.0 corresponds to macOS 10.9), which the pinning file's own comment says is a leftover: *"After we bumped the minimum macOS version to 10.13, this was left unchanged, since it is not essential."* That triple is used to form the compiler target, not the deployment-target value.

Given that, I kept the dynamic dylib-minos-probing approach (unchanged from before) rather than hardcoding either number — it measures what's actually in the installed dylibs, so it stays correct regardless of what conda-forge's pin is today or after a future bump, without needing to track it.

---

pyproject.toml
```diff
@@ -149,11 +149,30 @@ before-build = "python ./tools/cibuildwheel_before_build.py"
...
 before-all = "bash ./tools/cibuildwheel_before_all.sh"
-environment = { CMAKE_INCLUDE_PATH = "/usr/include/openblas", CCACHE_DEBUG = "true", CCACHE_DEBUGLEVEL = "1", CCACHE_LOGFILE = "/tmp/cytnx-ccache.log" }
+# CMAKE_PREFIX_PATH/LD_LIBRARY_PATH point at the conda-forge environment
+# tools/cibuildwheel_before_all.sh creates on manylinux; musllinux still
+# installs boost/openblas/arpack system-wide via apk, so both paths are
+# simply unused (and harmless) there. CMAKE_INCLUDE_PATH stays pointed at
+# EPEL's openblas-devel layout for the musllinux/apk fallback path -- Alpine's
+# own openblas-dev headers were never confirmed to need it, but removing it
+# is unverified and this is a no-op wherever it isn't needed.
+environment = { CMAKE_PREFIX_PATH = "/opt/cytnx-deps", LD_LIBRARY_PATH = "/opt/cytnx-deps/lib", CMAKE_INCLUDE_PATH = "/usr/include/openblas", CCACHE_DEBUG = "true", CCACHE_DEBUGLEVEL = "1", CCACHE_LOGFILE = "/tmp/cytnx-ccache.log" }
```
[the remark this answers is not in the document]
IvanaGyro: Checked this rigorously rather than re-asserting the theory, since it's a real gap worth verifying: `-fopenmp` linking `-lgomp` and `openblas=*=*openmp*` pulling in llvm-openmp are both correct facts, but they don't imply two separate runtimes here — `libgomp.so.1` is *itself* a symlink to `libomp.so` inside this conda-forge environment (conda-forge's `_openmp_mutex=*_llvm` coordination package does this deliberately, so GCC- and clang-built packages in the same environment share one physical runtime). The open question was whether that resolves correctly for the *actual* build (plain system GCC, no CC/CXX override, `LD_LIBRARY_PATH` set the way `pyproject.toml` sets it) and survives `auditwheel repair` specifically.

Reproduced exactly that scenario on this host:
- Compiled a test program with the **system g++ 13.3.0** (not a conda-forge compiler — matches the real manylinux build, which never overrides CC/CXX), `-fopenmp`, linked against the same `openblas=*=*openmp*` package, no RPATH baked in (matching that `INSTALL_RPATH_USE_LINK_PATH` is only `TRUE` on Apple, not Linux, in `CMakeLists.txt`).
- With `LD_LIBRARY_PATH` set exactly as `pyproject.toml`'s `[tool.cibuildwheel.linux].environment` sets it, `ldd` shows `libgomp.so.1` resolving to the identical realpath for both the extension and `libopenblas.so.0`'s own transitive need. Ran it — correct output, and `OMP_NUM_THREADS` controls the thread count as expected.
- Went one level deeper and used **`auditwheel`'s own dependency-resolution code** (`auditwheel.lddtree.ldd`, the exact machinery `auditwheel repair` uses to decide what to vendor) against the same binary: it also resolves `libgomp.so.1` to a single realpath (`libomp.so`), confirming repair would vendor it once, not twice, and wire up both the extension and the vendored OpenBLAS to the same copy.

So for this repo's actual build (no CC/CXX override, `LD_LIBRARY_PATH` set as it is), the runtimes do converge. The mechanism does depend on `LD_LIBRARY_PATH` search order beating any competing system `libgomp.so.1` at repair time — worth a regression test if this ever changes (e.g. an unrelated future edit to the `environment` table), but not a gap in the current state.

---

.github/workflows/ci-cmake_tests.yml
```diff
@@ -72,8 +72,18 @@ jobs:
...
+        # and cytnx's own code use OpenMP (-fopenmp), and an OpenBLAS build
+        # with its own independent (pthreads) thread pool oversubscribes the
+        # CPU when a BLAS call originates from inside an OpenMP region.
+        # _openmp_mutex=*_llvm was already pinned so the environment settles
+        # on one OpenMP runtime; this additionally pins the OpenBLAS build
+        # itself so the solver can't substitute a pthreads build that's
+        # merely compatible with, but not actually using, that runtime.
         run: |
-          mamba install _openmp_mutex=*=*_llvm cmake make boost git compilers blas=*=openblas liblapacke arpack gtest graphviz
+          mamba install _openmp_mutex=*=*_llvm cmake make boost git compilers "blas=*=openblas" "libopenblas=*=*openmp*" liblapacke arpack gtest graphviz
```
IvanaGyro: do we still need _openmp_mutex=*=*_llvm here and .github/workflows/ci-downstream-find-package.yml?

IvanaGyro: Yes, still needed in both, and it's not redundant with `libopenblas=*=*openmp*` — they pin two different things:

- `libopenblas=*=*openmp*` picks the OpenMP-threaded OpenBLAS *build variant* over the pthreads one.
- `_openmp_mutex=*=*_llvm` picks *which concrete OpenMP runtime* (llvm's `libomp` vs. GNU's `libgomp`) that OpenBLAS build and the `-fopenmp`-compiled cytnx/HPTT code both resolve to. Without it the solver could still land on `_openmp_mutex=*_gnu`, and the two OpenMP consumers wouldn't necessarily share one physical runtime even though both are individually "openmp-based" (see the `libgomp.so.1` → `libomp.so` symlink mechanism discussed in the P2 review thread above).

`ci-downstream-find-package.yml` already had `_openmp_mutex=*=*_llvm` but was missing the `libopenblas=*=*openmp*` pin `ci-cmake_tests.yml` has — added it in a0ffe55 for consistency (its own `TODO(deps-single-source)` comment already flags this dependency list as a near-duplicate of `ci-cmake_tests.yml`'s).

---

tools/cibuildwheel_before_all_macos.sh
```diff
@@ -2,64 +2,86 @@
...
+    if [[ "$(printf '%s\n' "${max_minos}" "${minos}" | sort -V | tail -n1)" == "${minos}" ]]; then
+      max_minos="${minos}"
+    fi
   fi
-  echo | tee -a "${report_file}"
-done
+# -L follows conda-forge's own symlinks so dylibs nested a level deeper are
+# still picked up. Keep -maxdepth bounded to guard against symlink-induced
+# recursion.
+done < <(find -L "${deps_prefix}/lib" -maxdepth 4 -name "*.dylib" -type f)
```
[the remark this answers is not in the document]
IvanaGyro: Valid — fixed in 9a073ca by moving ccache into its own `/tmp/cytnx-build-tools` prefix, separate from `/tmp/cytnx-deps` (openblas/arpack/boost/llvm-openmp), so the minos probe only ever scans the wheel's actual runtime dependency closure. `pyproject.toml`'s `PATH` override now points at the new prefix for ccache discoverability; `CMAKE_PREFIX_PATH` is untouched since it was already only `/tmp/cytnx-deps`.

---

.github/workflows/ci-cmake_tests.yml
IvanaGyro: We use libgomp so we can remove _openmp_mutex=*=*_llvm or pick gomp one.

IvanaGyro: You're right that the pin should go — I removed `_openmp_mutex=*=*_llvm` from both `ci-cmake_tests.yml` and `ci-downstream-find-package.yml` in dc316cb. But I need to correct my previous reply here: it was wrong on two counts, and I verified the actual solver behavior with a `micromamba create --dry-run` this time rather than reasoning about it.

1. **"pick gomp one" isn't possible.** conda-forge's OpenMP OpenBLAS build hard-requires the llvm mutex — `_openmp_mutex=*=*_gnu` together with `libopenblas=*=*openmp*` fails to solve outright:
   ```
   libopenblas =* *openmp* is not installable because it requires
     _openmp_mutex =* *_llvm, which conflicts...
   ```
2. **The explicit `_llvm` pin was redundant, not load-bearing** (contra what I said before). Resolving the `ci-cmake_tests.yml` spec with the pin removed still pulls `_openmp_mutex 4.5 7_kmp_llvm` + `llvm-openmp` transitively from `libopenblas=*=*openmp*` alone. So dropping it changes nothing about which runtime is selected.

On "we use libgomp": `libgomp.so.1` is present, but under `_kmp_llvm` it's a symlink to `libomp.so`, so the physical runtime everything (GCC's `-fopenmp -lgomp` and OpenBLAS) resolves to is LLVM's `libomp` — which is exactly what makes the single-thread-pool / no-oversubscription property hold. Comments in both files now record this instead of the old (incorrect) "the pin picks the runtime" rationale.

---

IvanaGyro:
Heads-up on an interaction with #1060 (enable ASan in `ci-cmake_tests`), found while profiling #1058.

This PR pins `ci-cmake_tests` to `libopenblas=*=*openmp*`. That OpenBLAS build has `NEEDED libgomp.so.1` + `RPATH $ORIGIN`, and in the `_openmp_mutex=llvm` env `libgomp.so.1` is a symlink to `libomp.so` (LLVM). So the OpenMP OpenBLAS pulls **LLVM libomp** into every test process via its own RPATH — independent of `LD_LIBRARY_PATH`.

LLVM libomp makes a one-time internal allocation at its first parallel region (`__kmp_allocate_align`, 112 bytes) that it never frees. Under **LeakSanitizer** that's reported as a direct leak, and the process exits non-zero. #1060 turns on ASan and *deliberately keeps LeakSanitizer ON for the gtest/ctest step* (only the Python-hosted steps get `detect_leaks=0`). So once **both** this PR and #1060 are in, the ctest step will fail on a libomp leak in ~every test — a false positive unrelated to Cytnx.

Evidence (local, `debug-openblas-cpu` + ASan):
- OpenMP OpenBLAS + libomp → `LeakSanitizer: detected memory leaks ... __kmp_allocate_align`, exit 1.
- The pthreads OpenBLAS pulls **no** OpenMP runtime (`0` OpenMP `NEEDED`), which is exactly why #1060's current CI (bare `blas=*=openblas` → pthreads) is leak-free apart from the one known `BlockUniTensorTest.CombineBondForceDoesNotRewriteUserHeldBond` (#1052) failure.

Whichever of the two merges second trips this. I've pushed the openmp pin + `OMP_WAIT_POLICY=passive` to #1060's branch to see it under real CI. If the ctest step reddens on the libomp leak, the fix is env-only and keeps genuine leak detection on — no blanket `detect_leaks=0`: preload the system GCC libgomp so it wins the `libgomp.so.1` soname before OpenBLAS's RPATH can pull libomp, ASan runtime first to satisfy its ordering check, e.g. `LD_PRELOAD="$(gcc -print-file-name=libasan.so) /usr/lib/x86_64-linux-gnu/libgomp.so.1"` on the gtest step. Verified locally: leak gone, test passes, real-leak detection preserved. (An LSan suppression `leak:__kmp_allocate` also works but disables detection of any real leak in that call path.)

*Reported by Claude (agent session operated by @IvanaGyro).*

---

ianmccul:
Don't preload GNU libgomp.  Using LSan suppression `leak:__kmp_allocate` is fine here, Cytnx never calls this function, and we're only interested in Cytnx leaks, not allocations in support libraries. `omp_alloc()` seems to use a different allocation path, and does not look like it goes via `__kmp_allocate`. It would nevertheless be worth examining the complete allocation stack. There may be a caller above `__kmp_allocate` that identifies this particular one-time startup allocation more precisely and would make a better suppression target.

Note that one-off allocations are not normally regarded as *leaks* as such anyway.  For example, the dynamic dispatch tables used by Cytnx, I am slightly surprised that these are freed at program termination -- I guess they use some global `std::vector` or some similar mechanism.  But if they were just created with `operator new[]`, or even `malloc`, and never freed, that would be OK, the memory isn't really a leak because it never accumulates. In fact it is slightly more efficient because it doesn't call back into the memory manager at program exit - there is no point freeing the memory on exit because its all about to be returned to the operating system.

“Every allocation must be explicitly freed so LSan passes without suppressions” forbids a legitimate process-lifetime allocation strategy.  The better policy is:

-  “LSan must pass with documented suppressions”

IvanaGyro:
I didn't dig into it. The CI test didn't report LSan errors. Somehow the test workflow links with libgomp instead of libomp from conda-forge, so there is no issue there. However, the message is still good for develops to solve the problem.

yingjerkao (approved):
LGTM

PR #1060, opened 2026-07-15
ci(cmake_tests): run the main test gate under ASan (debug-openblas-cpu)

### LINKED ISSUE #1052: ASan memcpy-param-overlap in BlockUniTensor::permute_ (via combineBond), CombineBondForceDoesNotRewriteUserHeldBond

IvanaGyro:
## Problem

The `ci-cmake_tests` job builds the "Editable install" step with `CMAKE_BUILD_TYPE=Debug` + `RUN_TESTS=ON`, but explicitly leaves `USE_DEBUG=OFF` (the preset default), so `-fsanitize=address` is never applied. AddressSanitizer therefore never runs in the main CI gate. This let a real overlapping-`memcpy` UB bug (`BlockUniTensor::combineBonds`, #1052) sit undetected for a long time — it only surfaces locally when a developer happens to build with a `debug-*` preset.

## Fix

- Switch the "Editable install" step to configure via `--config-settings=cmake.args=--preset=debug-openblas-cpu` instead of individual `cmake.define` overrides. The preset already sets `CMAKE_BUILD_TYPE=Debug`, `RUN_TESTS=ON`, and `USE_DEBUG=ON`, so this job now runs the whole suite (ctest + pytest) under ASan. `build-dir=build` is kept so ctest/gcovr still find a stable path.
- Rewrote the step's comment block, which previously justified why `USE_DEBUG` stayed off; that rationale is now inverted.
- Added an `LD_PRELOAD`/`ASAN_OPTIONS` shim to the two Python-hosted steps only ("Sanity-check the install" and "Run pytest with Python coverage"): ASan-instrumented `cytnx.so` loaded into the uninstrumented `python` binary needs `libasan.so` preloaded, or every native (C++-side) exception crashes the process — ASan's `__cxa_throw` interceptor CHECK-fails because plain `python` doesn't link `libstdc++`, so it can't resolve the real symbol at init. `libstdc++.so` must be preloaded alongside `libasan.so` for the same reason. `ASAN_OPTIONS=detect_leaks=0` is also needed for these two steps — LeakSanitizer's exit-time scan otherwise reports CPython's own deliberately-not-cleaned-up interpreter state (interned strings, static type objects, allocator arenas) as leaks, unrelated to whether Cytnx itself leaks.
- Left "Build the gtest binary" / "Run gtest via ctest" untouched — `test_main` links against ASan natively and needs no shim, and this keeps leak detection on for the one place a genuine Cytnx C++ leak would actually be caught.
- Updated `CLAUDE.md`'s CI-gates section to note this gate now runs under `debug-openblas-cpu` with ASan, in a separate commit from the workflow change. Checked `GEMINI.md` for overlapping review guidance — it does not mention `ci-cmake_tests` or ASan status anywhere, so it was left unchanged.

**This PR does not fix #1052 itself** (that's a separate PR). `ctest` is expected to show exactly one known failure, `BlockUniTensorTest.CombineBondForceDoesNotRewriteUserHeldBond`, until the separate #1052 fix PR merges. Related to #1052 (not closing it).

## Testing

- Local `debug-openblas-cpu` build + `ctest` (via this repo's `build-test-workflow` skill): **1784 tests run, 1783 passed, 1 failed** — the failure is exactly `BlockUniTensorTest.CombineBondForceDoesNotRewriteUserHeldBond`, matching the known pre-existing #1052 bug, and nothing else.
- Local pytest run under the ASan `LD_PRELOAD`/`ASAN_OPTIONS` shim is still in progress (the Python-bindings build for this preset takes a while on this machine) — it will be confirmed as a follow-up comment/push here.
- Opening this as a draft now per plan: since this PR itself modifies `ci-cmake_tests.yml`, the same-repo CI run on this PR branch exercises the *actual* modified workflow (preset switch + `LD_PRELOAD` shim) and is the most direct verification of whether the ASan setup works end-to-end. Watching CI once it starts and will fix anything it catches.


CLAUDE.md
```diff
@@ -102,6 +102,11 @@ Green locally before opening a PR. CI enforces:
 
 - **clang-format-check** (v14) — formatting clean.
 - **ci-cmake_tests** — `ctest` (`test_main`) *and* `pytest pytests/` both pass; Codecov.
+  This gate now builds under the `debug-openblas-cpu` preset, so `USE_DEBUG=ON`
```
IvanaGyro: No need for these nes lines. The new skill for testing will cover the preset and related instructions.

.github/workflows/ci-cmake_tests.yml
```diff
@@ -109,21 +110,48 @@ jobs:
...
+        # CHECK-fails on real___cxa_throw != 0 because plain `python` doesn't
+        # link libstdc++ at all, so the interceptor can't resolve the real
+        # symbol at init. libstdc++.so must be preloaded alongside libasan.so
+        # for the same reason. ASAN_OPTIONS=detect_leaks=0 suppresses
+        # LeakSanitizer's exit-time scan, which otherwise reports CPython's
+        # own deliberately-not-cleaned-up interpreter state (interned
+        # strings, static type objects, allocator arenas) as leaks -- noise
+        # unrelated to whether Cytnx itself leaks.
+        run: |
+          export LD_PRELOAD="$(gcc -print-file-name=libasan.so) $(gcc -print-file-name=libstdc++.so)"
```
IvanaGyro: This may be different than the compiler used to build. And consider to save the vars to GitHub var to reuse them in other steps. #1061 will reconsider to build ASan into pycynx.

CLAUDE.md
IvanaGyro: Agreed, removed in 940f2864 — reverted to the original one-line ci-cmake_tests bullet.

---

.github/workflows/ci-cmake_tests.yml
IvanaGyro: Compiler-mismatch fixed in 2028e8c6, same fix pattern as the one gemini-code-assist requested in #1061 (read `CMAKE_CXX_COMPILER` from the build's own `CMakeCache.txt`, confirm it identifies as GCC via the "Free Software Foundation" marker in `--version`, require both resolved paths to be absolute). This CI job's toolchain is always GCC by construction, so I made an unexpected result here a hard `::error::` failure rather than the skill script's silent-skip — an unresolved compiler here means the environment itself changed, which should fail loudly.

Also computed once now: added a "Resolve ASan runtime paths" step (`id: asan_preload`) that writes the resolved value to `$GITHUB_OUTPUT`, referenced via step-scoped `env:` in both "Sanity-check the install" and "Run pytest with Python coverage" — no more duplicated `gcc -print-file-name` calls. I deliberately didn't hoist it to job-wide `$GITHUB_ENV`: that would leak `ASAN_OPTIONS=detect_leaks=0` into "Run gtest via ctest" too, which would undo the existing design that keeps leak detection on there (the one place a genuine Cytnx C++ leak would actually be caught).

On "build ASan into pycytnx" — I checked before assuming: `cytnx.cpython-311-x86_64-linux-gnu.so` already links `libasan.so.8` as a `DT_NEEDED` dependency and contains 39 `__asan_*` symbols (`-fsanitize=address` is appended to the global `CMAKE_CXX_FLAGS` in `CMakeLists.txt` before `pybind11_add_module(pycytnx ...)` is declared, so it inherits identically to how `test_main` does — this isn't something that needs building in, it's already there). Running `python -c "import cytnx"` with **zero** `LD_PRELOAD`/`ASAN_OPTIONS` set fails immediately with ASan's own message: `"ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD."` That's the actual constraint: ASan requires its runtime to be the first thing loaded into a process, and Python's `import` is a `dlopen()` happening after the (uninstrumented) `python` process has already started — so `LD_PRELOAD` isn't a workaround for missing instrumentation in pycytnx, it's the standard, documented mechanism for using instrumentation that's already dlopen'd into an uninstrumented host (same pattern numpy/PyTorch/etc. use for ASan-testing their own C extensions). I don't think there's a way to avoid it without shipping a custom ASan-linked Python interpreter, which seems like a much bigger cost than the current shim. Posted the same finding on #1061's thread with the same evidence.

---

IvanaGyro:
The failed test is fixed by #1054.

.github/workflows/ci-cmake_tests.yml
```diff
@@ -109,20 +110,85 @@ jobs:
...
+        # `--preset=openblas-cpu` below (cmake.args=--preset=X empirically
+        # wins over the pyproject.toml default; see
+        # .claude/skills/build-test-workflow/scripts/build_preset.sh for the
+        # verified mechanics).
         run: |
           pip install --editable '.[dev]' --verbose \
             --config-settings=build-dir=build \
-            --config-settings=cmake.define.CMAKE_BUILD_TYPE=Debug \
-            --config-settings=cmake.define.RUN_TESTS=ON
+            --config-settings=cmake.args=--preset=debug-openblas-cpu
```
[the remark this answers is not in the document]
IvanaGyro: This is correctly observed, and it's the expected/deliberate state of this PR right now — see the "Problem"/"Testing" sections of the PR description: it explicitly documents that `ctest` shows exactly this one known failure until #1052's fix PR (#1054) merges, and doesn't claim otherwise.

Sequencing: #1054 is the actual fix (already ready for review, all its own CI green), and the maintainer has confirmed on this PR that landing #1054 resolves this failure — not xfail/exclude, since that would mask a real, already-diagnosed memory-safety bug rather than fix it. This PR intentionally makes it visible in CI (that's the whole point — the bug was previously undetected because nothing built with ASan on) rather than hiding it further.

---

PR #1071, opened 2026-07-16
fix(utils): honor non-unit step in GPU arange for real dtypes (#1070)

### LINKED ISSUE #1070: GPU arange ignores non-unit step for real dtypes (missing parentheses in cuSetArange_kernel)
## Summary

`cytnx::arange(start, end, step)` on `Device.cuda` **ignores `step` for real (non-complex) dtypes**. Complex dtypes (ComplexDouble/ComplexFloat) are correct.

## Reproduction

```cpp
auto g = cytnx::arange(10, 40, 10, cytnx::Type.Int64, cytnx::Device.cuda).to(cytnx::Device.cpu);
// g == [10, 11, 12]   (WRONG; expected [10, 20, 30])

auto c = cytnx::arange(10, 40, 10, cytnx::Type.Int64);   // CPU
// c == [10, 20, 30]   (correct)
```

Confirmed for `Int64` and `Double` on an RTX 4070 Ti SUPER (CUDA 13, sm_89):
GPU `arange` gave `[10, 11, 12]`, CPU gave `[10, 20, 30]`, and CPU-then-`.to(cuda)`
gave `[10, 20, 30]` — so it is the GPU `arange` **kernel**, not the host->device transfer.

## Root cause

`src/backend/utils_internal_gpu/cuSetArange_gpu.cu`, the generic real-type
`cuSetArange_kernel` (line ~10) is missing parentheses around the linear index:

```cpp
in[blockIdx.x * blockDim.x + threadIdx.x] =
    start + step * blockIdx.x * blockDim.x + threadIdx.x;   // <-- BUG
```

By C++ precedence this is `start + (step * blockIdx.x * blockDim.x) + threadIdx.x`,
so the per-thread offset `threadIdx.x` is added **without** being multiplied by
`step`. For any array that fits in a single block (`blockIdx.x == 0`) it collapses
to `start + threadIdx.x`, i.e. `step` is effectively 1 — exactly the `[10,11,12]`
above.

The complex-dtype overloads in the *same file* are already correct:

```cpp
make_cuDoubleComplex(start + step * (blockIdx.x * blockDim.x + threadIdx.x), 0);
```

## Fix

Parenthesize the index in the generic kernel to match the complex overloads:

```cpp
in[blockIdx.x * blockDim.x + threadIdx.x] =
    start + step * (blockIdx.x * blockDim.x + threadIdx.x);
```

## Impact

Affects all real dtypes (`Int64/Int32/Int16`, `Uint*`, `Double`, `Float`) with
`step != 1` on GPU. Pre-existing and independent of #1067; found while writing GPU
in-place arithmetic tests there (worked around by building test inputs with CPU
`arange` then `.to(cuda)`).

yingjerkao:
## Problem

`cytnx::arange(start, end, step)` on `Device.cuda` ignores `step` for real
(non-complex) dtypes. `arange(10, 40, 10, Type.Int64, Device.cuda)` produced
`[10, 11, 12]` instead of `[10, 20, 30]` (CPU is correct). See #1070.

## Fix

The generic real-type `cuSetArange_kernel` in
`src/backend/utils_internal_gpu/cuSetArange_gpu.cu` was missing parentheses around
the linear index:

```cpp
start + step * blockIdx.x * blockDim.x + threadIdx.x;   // = start + (step*blk*dim) + thread
```

so the per-thread offset `threadIdx.x` was added un-scaled by `step` (for a single
block this collapses to `start + threadIdx.x`). The complex-dtype overloads in the
same file already had the parentheses. Fixed to:

```cpp
start + step * (blockIdx.x * blockDim.x + threadIdx.x);
```

## Testing

Added `tests/gpu/arange_test.cpp`:
- **independent literal check** — `arange(10,40,10)` on the GPU is `[10,20,30]`;
- real dtypes (`Int64/Int32/Double/Float`) at single-block (n=3) and multi-block
  (n=1000) sizes, checked against the CPU `arange`.

Both pass on an RTX 4070 Ti SUPER (CUDA 13, sm_89). (Verified on the GPU branch
where the bug was found; the kernel and test are independent of that branch's
changes.) GPU tests don't run in CI (no GPU runners), so this was validated
locally.

Closes #1070.

### DIFF
--- a/tests/gpu/arange_test.cpp
+++ b/tests/gpu/arange_test.cpp
@@ -1,0 +1,40 @@
+#include "gtest/gtest.h"
+
+#include "gpu_test_tools.h"
+#include "cytnx.hpp"
+
+// Regression for the GPU arange step bug: cuSetArange_kernel dropped the parentheses
+// around the linear index, so `start + step * blockIdx.x * blockDim.x + threadIdx.x`
+// added the per-thread offset without scaling it by step. For arrays that fit in one
+// block this collapsed to `start + threadIdx.x` (step effectively 1); for larger
+// arrays the per-thread offset was still unscaled. Only real dtypes were affected;
+// the complex overloads already parenthesized the index.
+namespace ArangeTest {
+
+  using namespace cytnx;
+
+  // Independent literal check: arange(10, 40, 10) is [10, 20, 30] on the GPU.
+  TEST(GpuArangeTest, nonunit_step_literal) {
+    Tensor g = arange(10, 40, 10, Type.Int64, Device.cuda).to(Device.cpu);
+    ASSERT_EQ(g.shape()[0], static_cast<cytnx_uint64>(3));
+    EXPECT_EQ(g.at<cytnx_int64>({0}), 10);
+    EXPECT_EQ(g.at<cytnx_int64>({1}), 20);
+    EXPECT_EQ(g.at<cytnx_int64>({2}), 30);
+  }
+
+  // Real dtypes, single-block (n=3) and multi-block (n=1000 > blockDim) sizes, checked
+  // against the (correct, independent) CPU arange implementation.
+  TEST(GpuArangeTest, nonunit_step_real_dtypes) {
+    for (auto dt : {Type.Int64, Type.Int32, Type.Double, Type.Float}) {
+      for (cytnx_uint64 n : {static_cast<cytnx_uint64>(3), static_cast<cytnx_uint64>(1000)}) {
+        SCOPED_TRACE("dtype=" + std::to_string(dt) + " n=" + std::to_string(n));
+        const double end = 2.0 * static_cast<double>(n);
+        Tensor cpu = arange(0, end, 2, dt);
+        Tensor gpu = arange(0, end, 2, dt, Device.cuda).to(Device.cpu);
+        ASSERT_EQ(gpu.shape()[0], n);
+        EXPECT_TRUE(TestTools::AreNearlyEqTensor(gpu, cpu, 1e-6));
+      }
+    }
+  }
+
+}  // namespace ArangeTest
--- a/src/backend/utils_internal_gpu/cuSetArange_gpu.cu
+++ b/src/backend/utils_internal_gpu/cuSetArange_gpu.cu
@@ -8,7 +8,7 @@ namespace cytnx {
                                        cytnx_uint64 Nelem) {
       if (blockIdx.x * blockDim.x + threadIdx.x < Nelem) {
         in[blockIdx.x * blockDim.x + threadIdx.x] =
-          start + step * blockIdx.x * blockDim.x + threadIdx.x;
+          start + step * (blockIdx.x * blockDim.x + threadIdx.x);
       }
     }
     __global__ void cuSetArange_kernel(cuDoubleComplex *in, cytnx_double start, cytnx_double step,
--- a/tests/gpu/CMakeLists.txt
+++ b/tests/gpu/CMakeLists.txt
@@ -11,6 +11,7 @@ add_executable(
   DenseUniTensor_test.cpp
   Accessor_test.cpp
   Tensor_test.cpp
+  arange_test.cpp
   utils_test/vec_concatenate.cpp
   utils_test/vec_unique.cpp
   utils/getNconParameter.cpp


tests/gpu/arange_test.cpp
```diff
@@ -0,0 +1,40 @@
...
+// added the per-thread offset without scaling it by step. For arrays that fit in one
+// block this collapsed to `start + threadIdx.x` (step effectively 1); for larger
+// arrays the per-thread offset was still unscaled. Only real dtypes were affected;
+// the complex overloads already parenthesized the index.
+namespace ArangeTest {
+
+  using namespace cytnx;
+
+  // Independent literal check: arange(10, 40, 10) is [10, 20, 30] on the GPU.
+  TEST(GpuArangeTest, nonunit_step_literal) {
```
IvanaGyro, line 17: `_` is not allowed in GoogleTest's test names and test suite names. See: https://google.github.io/googletest/reference/testing.html#TEST

tests/gpu/arange_test.cpp
```diff
@@ -0,0 +1,40 @@
...
+    EXPECT_EQ(g.at<cytnx_int64>({1}), 20);
+    EXPECT_EQ(g.at<cytnx_int64>({2}), 30);
+  }
+
+  // Real dtypes, single-block (n=3) and multi-block (n=1000 > blockDim) sizes, checked
+  // against the (correct, independent) CPU arange implementation.
+  TEST(GpuArangeTest, nonunit_step_real_dtypes) {
+    for (auto dt : {Type.Int64, Type.Int32, Type.Double, Type.Float}) {
+      for (cytnx_uint64 n : {static_cast<cytnx_uint64>(3), static_cast<cytnx_uint64>(1000)}) {
+        SCOPED_TRACE("dtype=" + std::to_string(dt) + " n=" + std::to_string(n));
```
IvanaGyro, line 30: We can use std::format and DTypeShortName defined here: https://github.com/Cytnx-dev/Cytnx/blob/f64b2f57f61d0e08b54b2b7bea3929e5f090a3ce/tests/Scalar_test.cpp#L586

tests/gpu/arange_test.cpp
```diff
@@ -0,0 +1,40 @@
...
+        const double end = 2.0 * static_cast<double>(n);
```
IvanaGyro, line 31: Is this static_cast needed? Do we need a test for `step * (n -1) < end <= step * n` and also a test that has 1 ULP difference at the boundary?

ianmccul:
Note that the CPU version of `arrange()` is rather .... interesting. It computes the length with integer truncation plus fmod(...) > 1e-14, which is neither a reliable ceil nor a stable half-open-range test. In particular, familiar decimal steps can make it include the nominal endpoint.

The GPU comparison test inherits the front end's incorrect length calculation.

The `arrange()` front end computes:
```c++
Nelem = (end - start) / step;
if (fmod(end - start, step) > 1.0e-14) Nelem += 1;
```
The documented range is [start, end), so both are wrong. `fmod(1.0, 0.1)` is approximately 0.1, not zero, because of floating-point representation.

The fixed absolute 1e-14 threshold causes the opposite error for small scales: `arange(0, 1e-15, 2e-15)` should contain the single value 0, but Cytnx computes zero elements and throws. Empty or direction-mismatched ranges also throw, despite zero-extent tensors now being supported.

A sane implementation starts from something equivalent to:
```c++
const double count = (end - start) / step;
const auto nelem = static_cast<std::uint64_t>(std::ceil(count));
```
with explicit checks for zero/non-finite step, overflow, and empty ranges. Boundary tests should include nextafter(end, ±infinity).

tests/gpu/arange_test.cpp
ianmccul: `for (auto n : {3, 1000}) {`

And *yes* such tests would be quite useful, since they would currently fail.

ianmccul: Assuming that the CPU implementation is correct is a very bold statement!  This is Cytnx!

tests/gpu/arange_test.cpp
yingjerkao: Can this be folded into our naming convention md file?

yingjerkao:
Closing in favor of #1076.

@ianmccul's review here showed that the GPU kernel parenthesization fix in this PR is only half the story. The shared CPU front-end `src/Generator.cpp::arange` computes the element count `Nelem` with a broken length calculation (`fmod(...) > 1e-14`) that is neither a reliable ceil nor a stable half-open `[start, end)` test — it can include the endpoint, breaks at small scales (`arange(0, 1e-15, 2e-15)` throws instead of yielding a single `0`), and throws on empty/reversed ranges instead of returning a zero-extent tensor. That same `Nelem` feeds **both** the CPU and GPU kernels, so:

- fixing the GPU kernel alone leaves the length wrong on both paths, and
- the GPU test here validates against the (also-buggy) CPU result as its oracle ("Assuming that the CPU implementation is correct is a very bold statement! This is Cytnx!").

Rather than land a GPU-kernel-only patch, the CPU front-end and GPU kernel should be corrected **together** against one spec, with independent per-path oracles and proper boundary/ULP tests. I've folded this PR's review — Ian's comments and suggested fixes (ceil-based length, zero/non-finite-step/overflow/empty-range guards, zero-extent for empty ranges, `nextafter` boundary tests) plus Gemini's 64-bit index-overflow finding — into **#1076**, which will carry the final coordinated fix.

#1070 stays open and will be resolved by #1076.

PR #1088, opened 2026-07-19
docs(arange): document numpy-consistency + numpy-comparison tests (#1083)

### LINKED ISSUE #1083: arange: add numpy-comparison pytests + document numpy-consistency (follow-up to #1077)
Follow-up to #1077 (merged), capturing @ianmccul's approval-time requests.

#1077 reworked `arange` to `Nelem = ceil((end - start) / step)`, matching numpy. Ian approved and asked for two things before/after landing:

## 1. numpy-comparison pytests

Add pytests that check `cytnx.arange(...)` against `numpy.arange(...)` for element count and values — including the FP edge cases where the two must agree.

## 2. Document numpy-consistency (and correct the "strictly half-open" framing)

Ian's point: the `ceil`-based count can **include, or even land slightly beyond, the nominal endpoint** in some floating-point cases — contrary to a strict half-open `[start, end)` reading, but consistent with numpy:

```
>>> import numpy as np
>>> np.arange(0.5, 0.8, 0.1)
array([0.5, 0.6, 0.7, 0.8])   # 0.8 (the endpoint) is included
```

The current `src/Generator.cpp::arange` comment and #1077's description describe the behavior as strictly half-open / endpoint-excluded, which is inaccurate for these FP cases. Update the code comment (and any user-facing docstring) to state that Cytnx `arange` follows numpy's implementation, referencing https://numpy.org/doc/stable/reference/generated/numpy.arange.html, and note that the endpoint may be included/exceeded in FP edge cases.

## 3. (minor) Overflow guard

Ian on `Generator.cpp:67`: "If nelem >= 18446744073709551616 I think you've got bigger problems than overflowing a `uint64_t`..." — consider simplifying or dropping the `double -> uint64` overflow guard.

Refs: #1077, #1076. Requested by @ianmccul.

yingjerkao:
## Problem

Follow-up to #1077, capturing @ianmccul's approval-time requests. #1077 reworked `arange` to `Nelem = ceil((end - start) / step)`, which matches `numpy.arange` — but:

- the docs still described the range as strictly half-open with the endpoint "exclusive" / "the last element is the largest one smaller than `end`", which is **inaccurate**: floating-point rounding can make the final element equal or slightly exceed the endpoint (e.g. `np.arange(0.5, 0.8, 0.1) == [0.5, 0.6, 0.7, 0.8]`);
- there were no tests confirming consistency with numpy.

## Fix (docs + tests only — no behavior change)

- **Documentation**: reframed the `Generator.hpp` docstring and the `Generator.cpp` comment to state that `arange` follows [`numpy.arange`](https://numpy.org/doc/stable/reference/generated/numpy.arange.html), with the *nominal* half-open caveat and the endpoint-inclusion example spelled out.
- **Overflow guard**: kept (Ian flagged it as over-defensive), but added a comment explaining it is not paranoia — `count` can overflow to `+inf` for finite `start`/`end`/`step` (a huge span with a tiny step), and casting a non-representable double to `uint64` is UB. So it guards a reachable path, not just an un-allocatable size.

## Testing

- **`pytests/arange_numpy_test.py`** (new): parametrized differential tests against `numpy.arange` (an independent oracle) — element **count and values** across ascending/descending, fractional, the endpoint-inclusion FP cases, small scales, single-element, empty ranges, and the `arange(N)` overload.
- **`Arange.EndpointMayBeIncludedUnderRounding`** (new C++ test): documents the FP endpoint-inclusion (`arange(0.5, 0.8, 0.1)` → 4 elements incl. `0.8`) at the C++ level.

Closes #1083.

### DIFF
--- a/include/Generator.hpp
+++ b/include/Generator.hpp
@@ -93,13 +93,17 @@ namespace cytnx {
   */
   Tensor arange(cytnx_int64 Nelem);
   /**
-  @brief Create a rank-1 Tensor with incremental elements in the range [\p start,\p end) with given
-  step-size \p step between elements.
-  @details The elements are \p start, \p start + \p step, \p start + 2 * \p step, ... The last
-  element is the largest one that is smaller than \p end.
+  @brief Create a rank-1 Tensor with incremental elements \p start, \p start + \p step,
+  \p start + 2 * \p step, ..., following the same convention as
+  [numpy.arange](https://numpy.org/doc/stable/reference/generated/numpy.arange.html).
+  @details The number of elements is `ceil((end - start) / step)`. The range is nominally
+  half-open [\p start, \p end), but -- exactly as with numpy -- this is not guaranteed under
+  floating-point rounding: the final element may equal or slightly exceed \p end (e.g.
+  `arange(0.5, 0.8, 0.1)` yields `[0.5, 0.6, 0.7, 0.8]`). An empty or direction-mismatched range
+  (a non-positive count) yields a zero-extent Tensor.
   @param[in] start start value of the range.
-  @param[in] end end value of the range (exclusive).
-  @param[in] step step-size between subsequent elements in the range.
+  @param[in] end end value of the range (nominally exclusive; see details).
+  @param[in] step step-size between subsequent elements; must be non-zero.
   @param[in] dtype the data type of the Tensor, see cytnx::Type for more information.
   @param[in] device the device type of the Tensor, see cytnx::Device for more information.
   @return
--- a/pytests/arange_numpy_test.py
+++ b/pytests/arange_numpy_test.py
@@ -1,0 +1,62 @@
+"""cytnx.arange must match numpy.arange (#1083).
+
+cytnx.arange follows numpy's ceil-based element count, so the two agree on the
+element count AND the values -- including the floating-point edge cases where the
+nominally half-open [start, end) range includes (or slightly exceeds) the
+endpoint, e.g. np.arange(0.5, 0.8, 0.1) == [0.5, 0.6, 0.7, 0.8].
+"""
+
+import cytnx
+import numpy as np
+import pytest
+
+# Ascending/descending, integer and fractional steps, the endpoint-inclusion FP
+# cases, small scales, and single-element ranges.
+CASES = [
+    (0.0, 10.0, 1.0),
+    (10.0, 40.0, 10.0),
+    (0.0, 1.0, 0.25),
+    (0.5, 0.8, 0.1),  # numpy includes the endpoint here
+    (0.0, 1.0, 0.1),
+    (0.0, 0.3, 0.1),
+    (0.0, 0.4, 0.1),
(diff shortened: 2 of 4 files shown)


yingjerkao:
Added a second commit (4bafb02e) fixing a boundary bug in the overflow guard that surfaced while double-checking the "casting a double >= 2^64 to uint64_t is UB" claim.

**The bug.** The guard was:
```cpp
nelem > static_cast<cytnx_double>(std::numeric_limits<cytnx_uint64>::max())
```
`UINT64_MAX` (`2^64 - 1`) is **not** representable as a `double`, so `static_cast<double>(UINT64_MAX)` rounds **up** to `2^64` (`18446744073709551616.0`). The test was therefore effectively `nelem > 2^64`, which lets `nelem == 2^64` pass through into `static_cast<cytnx_uint64>(nelem)` — and per [conv.fpint] that cast is UB (the truncated value `2^64` doesn't fit in `uint64_t`). So the guard failed to prevent the exact UB it exists for, at the `2^64` boundary.

Quick confirmation:
```
(double)UINT64_MAX = 18446744073709551616.0   # == 2^64, not 2^64-1
nelem >  (double)UINT64_MAX , nelem = 2^64 -> fires? 0   # slips through
nelem >= (double)UINT64_MAX , nelem = 2^64 -> fires? 1
```

**The fix.** Compare with `>=` against `2^64` written explicitly as `0x1p64` (the first `double` not representable as `uint64_t`). It rejects no valid count — the largest `double` strictly below `2^64` is `2^64 - 2048`. Also dropped the now-unused `<limits>` include.

**Test.** New `Arange.OverflowingCountThrows` asserts `arange(0, 2^64, 1)` and a span/step whose count overflows to `+inf` both throw. Arange suite: 13 passed.

So the original comment's UB claim was correct — and checking it turned up this off-by-one at the boundary.

IvanaGyro (approved):

yingjerkao:
Rebased onto current master (`a969a66c`) to clear the merge conflict — the branch was 126 commits behind.

The only conflict was in `tests/arange_test.cpp`, and it was purely mechanical: master reindented that file (the tests sit three namespaces deep, so 6-space indent) while this branch still had the old 2-space form. Resolved by keeping master's indentation and reindenting the two tests this PR adds — `Arange.EndpointMayBeIncludedUnderRounding` and `Arange.OverflowingCountThrows`. No test logic changed; one comment line was rewrapped to stay under 100 columns.

`src/Generator.cpp`, `include/Generator.hpp` and `pytests/arange_numpy_test.py` replayed without conflict, and the `nelem >= 0x1p64` guard from `4bafb02e` is intact.

Verified locally on the rebased base: all 16 `Arange.*` / `DenseUniTensorTest.Arange*` ctest tests pass, and clang-format-14 is clean. The Python test could not be run on this machine (LAPACKE detection fails inside pip's isolated configure, unrelated to this change) — `ci-cmake_tests` covers `pytests/arange_numpy_test.py`.

IvanaGyro (approved):

PR #1091, opened 2026-07-19
refactor(linalg): migrate GPU in-place arithmetic to typed dispatch, remove dead cuAri_ii table (#1003)

yingjerkao:
## Problem

Part of #1003 (consolidating CUDA elementwise arithmetic). Two coupled issues:

1. **GPU in-place `iAdd/iSub/iMul/iDiv` still dispatched through the legacy `lii.cuAri_ii` 12×12 dtype-pair table** — the last "ordinary arithmetic" path reaching it (step 8). That table's non-contiguous kernel writes `out[idx]` while `out` **aliases** the non-contiguous LHS storage and reads `lhs[Lidx]` — a read-at-`Lidx`/write-at-`idx` scatter that **corrupts data** (#988). `iMul`/`iDiv` guarded against it with a hard error; `iAdd`/`iSub` silently corrupted.
2. Once (1) is fixed, the entire `cuAri_ii` table and its generated dispatcher `cuArithmetic_internal.{cu,hpp}` become dead code (step 12).

## Fix

**Commit 1 — step 8 (typed in-place dispatch).** Route the four in-place ops through the already-tested out-of-place `cu{Add,Sub,Mul,Div}_dispatch` kernels (from #1067), mirroring the CPU element-wise semantics (`DispatchInplaceArithmeticCPU`): in-place keeps the LHS dtype — compute in the promoted dtype, store into the LHS dtype. When the promoted dtype equals the LHS dtype the result is written straight into the LHS storage (preserves shared-storage aliasing); a broadcast scalar RHS is applied in place even for a non-contiguous LHS (layout-preserving, the #988 fix). Non-contiguous `tensor⊗tensor` in-place now uses the dispatch's mapper kernel and yields correct values.

**Commit 2 — step 12 (remove the dead table).** Delete `cuArithmetic_internal.{cu,hpp}` + its `bet.py` generator, drop the `cuAri_ii` member/population and the unused `Arithmeticfunc_oii` typedef from `linalg_internal_interface.{hpp,cpp}`, and include the six `cu{Add,Sub,Mul,Div,Cpr,Mod}_internal.hpp` headers directly (the umbrella header had transitively provided the live `cu*_dispatch` decls). Net −3413 lines.

### Behavior notes (flagged for review)
- A **non-contiguous LHS with a tensor RHS becomes contiguous** after the op (CPU keeps it strided). Values are correct; the old path corrupted/errored, so no working behavior regresses.
- **Mixed real/complex in-place computes in the promoted precision** per the #1003 spec, which can differ from CPU's `std::complex::operator op=` narrowing by ~float epsilon.

Full strided/precision parity would need a dedicated in-place kernel (writing `lhs[Lidx]`); noted as a possible follow-up.

## Testing

On RTX 4070 Ti / CUDA 13:
- 64 existing in-place GPU tests (`gpu_tensor_i{add,sub,mul,div}_*`, all shapes × dtype-preserving pairs) — pass
- new `tests/gpu/linalg_test/InplaceArithmetic_test.cpp`: dtype-widening, mixed complex/real, non-contiguous scalar, non-contiguous tensor, and independent hand-computed values — pass
- out-of-place arithmetic sanity slice — pass
- clang-format-14 clean; step 12 changes are `#ifdef UNI_GPU`-guarded so CPU builds are unaffected

> Note: the repo has no automated GPU CI runner, so these GPU results are from a local run.

Advances #1003 (steps 8 and 12); does not close it.


tests/gpu/linalg_test/InplaceArithmetic_test.cpp
```diff
@@ -0,0 +1,223 @@
...
+  // A nonzero singleton (value 3) of the requested dtype, on the GPU. Deterministic so iDiv never
+  // divides by zero.
+  Tensor NonzeroScalar(unsigned int dtype) {
+    return cytnx::arange(3, 4, 1, dtype).to(cytnx::Device.cuda);  // shape {1}, value 3
+  }
+
+  // Relative element-wise comparison. Used for mixed real/complex in-place: the GPU computes in the
+  // promoted dtype (per #1003) then narrows to the LHS dtype, while CPU's std::complex op= narrows
+  // the RHS to the LHS scalar type first, so the two agree only to float rounding.
+  bool NearlyEqRel(const Tensor& a_dev, const Tensor& b_dev, double rel_tol) {
```
IvanaGyro, line 92: It seems that there is a pre-defined tool in gpu_teset_tool.h. And by the coding style we defined in the skill, the function name should be snake case. Additionally, the test name should not use `_`

tests/gpu/linalg_test/InplaceArithmetic_test.cpp
```diff
@@ -0,0 +1,223 @@
+#include <algorithm>
+#include <cmath>
+
+#include "gtest/gtest.h"
+
+#include "gpu_test_tools.h"
+#include "cytnx.hpp"
```
IvanaGyro, line 7: Wrong include order.

yingjerkao:
Superseded by master. Since this PR was opened, master absorbed the same #1003 work through two merges:

- **#1080** already removed the dead `cuAri_ii` table, the `Arithmeticfunc_oii` typedef, and swapped the umbrella `cuArithmetic_internal.hpp` include for the six individual `cu{Add,Sub,Mul,Div,Mod,Cpr}_internal.hpp` headers — i.e. this PR's "remove dead table" commit.
- **#1013 / #1067** migrated the GPU in-place path to a dedicated `cuiArithmeticDispatch` in-place kernel that applies the layout mappers and writes through in place. That is strictly better than this PR's approach (which reused the *out-of-place* `cu*_dispatch` and therefore produced a contiguous result for non-contiguous `tensor (op)= tensor`, diverging from the CPU strided-write-through — the tradeoff flagged in this PR's description). Master now preserves strided layout with full CPU parity.

Test coverage is also covered upstream: master's `tests/gpu/linalg_test/InplacePromote_test.cpp` exercises the promoting + non-contiguous in-place cases (and validates the strided behavior this PR could not provide). Closing as fully superseded; no salvage needed.

PR #1093, opened 2026-07-19
refactor(linalg): constrain free Tensor/UniTensor scalar operators to scalar-like types (#1003)

yingjerkao:
## Problem

Ian's **item 3** from the #1003 review. The free namespace-scope arithmetic/comparison operators for `Tensor` and `UniTensor` (`+ - * / % ==` paired with a scalar) were declared as unconstrained `template <class T>`:

```cpp
template <class T> Tensor operator+(const T&, const Tensor&);
template <class T> cytnx::UniTensor operator*(const T&, const cytnx::UniTensor&);
template <class T> Tensor operator==(const Tensor&, const T&);
```

So an arbitrary `std` / user-defined type could be deduced as `T` and made the operator a viable candidate in overload resolution (e.g. under `using namespace cytnx`), even though the operators are only ever specialized/instantiated for the cytnx scalar surface.

## Fix

Add a `cytnx_scalar_like` concept in `linalg.hpp` admitting exactly that surface — the cytnx dtype scalars (reusing the existing `CytnxType` concept), `cytnx::Scalar`, and the `Tensor::Tproxy` / `Scalar::Sproxy` element proxies — and constrain the **22 free operator primary declarations** with it.

- The `Tensor` operators are **full specializations**, so constraining the primary declaration suffices (no `.cpp` change).
- The `UniTensor` operators are **generic templates**, so their **8 generic definitions** in `Add/Sub/Mul/Div/Mod.cpp` get the matching constraint (a constrained declaration needs a constrained definition).

`linalg::Add/Sub/Mul/Div/Mod/Cpr` (explicitly named — not an overload-resolution pollution vector; pybind uses these with explicit scalar casts) and the `ExpH/ExpM` templates are intentionally left unconstrained.

**Not a breaking change for valid use:** every builtin scalar literal deduces to a cytnx alias (`int` → `cytnx_int32`, `1.0` → `cytnx_double`, …); a non-scalar operand that "compiled" before only ever failed at link (no instantiation exists) and now fails with a clear overload error instead.

Adds `tests/operator_constraint_test.cpp`: compile-time guards that `cytnx_scalar_like` admits the scalars/proxies and rejects other types, that `scalar-on-the-left <op> Tensor/UniTensor` is viable for scalars but **not** for a non-scalar struct (these fail on pre-fix code), plus a runtime check that scalar operators and `Tensor<op>Tensor` still work.

## Testing

CPU (openblas): library + `test_main` build clean (the full library build confirmed no other call site relied on the loose operators); `OperatorConstraint` passes; `pybind/tensor_py.cpp` and `unitensor_py.cpp` compile clean. clang-format-14 clean.

Advances #1003 (Ian's operator-hygiene fold-in, item 3). Independent of the sibling #1003 branches (#1091 GPU in-place, #1092 complex_arithmetic).

### DIFF
--- a/include/Tensor.hpp
+++ b/include/Tensor.hpp
@@ -9,6 +9,7 @@
 #include <fstream>
 #include "utils/dynamic_arg_resolver.hpp"
 #include "Accessor.hpp"
+#include <type_traits>
 #include <utility>
 #include <vector>
 #include <initializer_list>
@@ -25,17 +26,11 @@ namespace cytnx {
 
   class Tensor;
 
-  ///@cond
-  // [Note] these are fwd from linalg.hpp
-  template <class T>
-  Tensor operator+(const Tensor &lhs, const T &rc);
-  template <class T>
-  Tensor operator-(const Tensor &lhs, const T &rhs);
-  template <class T>
-  Tensor operator*(const Tensor &lhs, const T &rhs);
-  template <class T>
-  Tensor operator/(const Tensor &lhs, const T &rhs);
-  ///@endcond
+  // [Note] The free `Tensor <op> T` operators declared in linalg.hpp are forward-declared below
+  // the class definition rather than here, because their constraint names Tensor::Tproxy, which
+  // is not available until Tensor is complete. Declaring them unconstrained here would defeat the
+  // constraint entirely: a differently-constrained declaration is a *distinct* template, so the
+  // unconstrained candidate would stay viable for every non-scalar T (#1003, Ian's review).
 
   /// @brief an tensor (multi-dimensional array)
   class Tensor {
@@ -1787,6 +1782,47 @@ namespace cytnx {
     Tensor Min() const;
 
   };  // class Tensor
+
+  // Scalar-like operand concepts for the free Tensor / UniTensor arithmetic and comparison
+  // operators. They live here, after the class, because cytnx_scalar_like names Tensor::Tproxy
+  // and so needs Tensor to be complete; linalg.hpp includes this header and reuses them.
+  //
+  // Each concept admits exactly the operand types the corresponding operators are instantiated
+  // for, so an unsupported operand is rejected at overload resolution instead of surviving to a
+  // link error. Widening one without adding the matching explicit instantiation reintroduces
+  // precisely that failure (#1003, operator hygiene -- Ian's review).
+
+  /// @cond
+  // Plain scalar values: the cytnx dtype scalars and cytnx::Scalar. This is the operand set of
+  // `operator%` (Tensor and UniTensor) and `operator==` (Tensor), which have no element-proxy
+  // instantiations -- see Mod.cpp and Cpr.cpp.
+  template <class T>
+  concept cytnx_scalar_value =
+    CytnxType<std::remove_cvref_t<T>> || std::is_same_v<std::remove_cvref_t<T>, Scalar>;
+
+  // Adds both element proxies: the operand set of the free Tensor `+ - * /` operators.
+  template <class T>
+  concept cytnx_scalar_like =
+    cytnx_scalar_value<T> || std::is_same_v<std::remove_cvref_t<T>, Tensor::Tproxy> ||
+    std::is_same_v<std::remove_cvref_t<T>, Scalar::Sproxy>;
+
+  // Adds only the Scalar proxy: the operand set of the free UniTensor `+ - * /` operators, which
+  // are instantiated for Scalar::Sproxy but never for Tensor::Tproxy -- see Add.cpp.
+  template <class T>
+  concept cytnx_unitensor_scalar_like =
+    cytnx_scalar_value<T> || std::is_same_v<std::remove_cvref_t<T>, Scalar::Sproxy>;
+
+  // [Note] these are fwd from linalg.hpp; the constraint must match linalg.hpp exactly, or the
+  // two declarations become distinct overloads and the unconstrained one stays viable.
+  template <cytnx_scalar_like T>
+  Tensor operator+(const Tensor &lhs, const T &rc);
+  template <cytnx_scalar_like T>
+  Tensor operator-(const Tensor &lhs, const T &rhs);
+  template <cytnx_scalar_like T>
+  Tensor operator*(const Tensor &lhs, const T &rhs);
+  template <cytnx_scalar_like T>
+  Tensor operator/(const Tensor &lhs, const T &rhs);
+  /// @endcond
 
   Tensor operator+(const Tensor &lhs, const Tensor::Tproxy &rhs);
   Tensor operator-(const Tensor &lhs, const Tensor::Tproxy &rhs);
--- a/include/linalg.hpp
+++ b/include/linalg.hpp
@@ -12,6 +12,7 @@
 #else
 
   #include <functional>
+  #include <type_traits>
 
   #include "backend/Scalar.hpp"
   #include "backend/Storage.hpp"
--- a/include/linalg.hpp
+++ b/include/linalg.hpp
@@ -19,16 +19,12 @@
 
 namespace cytnx {
 
-  // A scalar-like operand for the free Tensor / UniTensor arithmetic and comparison operators
-  // (`+ - * / % ==`). It admits exactly the types those operators are specialized for: the cytnx
-  // dtype scalars (CytnxType), cytnx::Scalar, and the Tensor / Scalar element proxies. Constraining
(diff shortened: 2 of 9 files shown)


yingjerkao:
Thanks — addressed what I safely could:

- **Gemini (`<type_traits>`):** added the explicit include to `linalg.hpp`.

- **Codex (constrain the `Tensor.hpp` forward declarations):** good catch, and it's real — `Tensor + non_scalar` still binds the unconstrained forward-declared free operator and compiles. But it's not a small fix. Those forward declarations are load-bearing for `Tensor`'s inline member templates, e.g.

  ```cpp
  template <class T> Tensor Mul(const T &rhs) { return *this * rhs; }
  ```

  (`operator*`/`operator/` have no member form). Removing them breaks the build at `Tensor.hpp:1357`. And they can't simply be constrained in place: `cytnx_scalar_like` depends on `Tensor::Tproxy`, which is still incomplete at the point of the forward declarations, so the concept isn't available there.

  Closing the scalar-on-the-right gap needs a concept/header reorganization (defining a scalar-like concept early, or reworking those member templates). I've documented the gap in `operator_constraint_test.cpp` and propose it as a scoped follow-up rather than risking the header restructure in this PR.

yingjerkao:
Review status (bot threads) — all three look already addressed by the current branch head:
- `<type_traits>` include for `cytnx_scalar_like` (@gemini-code-assist): present at `include/linalg.hpp:15`.
- Exclude/handle proxies (@chatgpt-codex-connector): the concept already lists `Tensor::Tproxy` and `Scalar::Sproxy` (linalg.hpp:30-31).
- Constrain the Tensor operator forward declarations too (@chatgpt-codex-connector): the forward decls are already `template <cytnx_scalar_like T>` (e.g. linalg.hpp:3159).

Ready to resolve.

ianmccul (changes requested):
1. Tensor RHS remains unconstrained.
https://github.com/Cytnx-dev/Cytnx/blob/bff592b25e5ce5d9a9090d6b491bcb25e5cf506d/include/Tensor.hpp#L30-L37

2.  cytnx_scalar_like does not describe the implemented concept. `Tensor::Tproxy` is a deferred `Tensor` selection, potentially of any rank, not inherently scalar. Some proxy arithmetic works by materializing it as a `Tensor:`
  ```c++
   Tensor operator+<Tensor::Tproxy>(...) {
     return Tensor(lc) + Rt;
   }
  ```
  That should be a proxy-specific overload, not membership in a scalar concept.

Worse, the concept admits combinations with no implementation:
```c++
     Tproxy % Tensor
     Tproxy == Tensor
     Tproxy + UniTensor
     Scalar % UniTensor
     Sproxy % UniTensor
```
These declarations participate in overload resolution but end in undefined references.

3. Implicit Tensor construction still admits non-scalars.
  The test itself avoids containers because std::vector<cytnx_uint64> converts implicitly to `Tensor.` Therefore:
```c++
     std::vector<cytnx_uint64> shape{2, 3};
     auto result = shape + tensor;
```
can bind `Tensor + Tensor`. That directly contradicts the PR’s claim that arbitrary standard-library operands become non-viable. The `Tensor` shape constructors should ultimately be explicit.

4. The tests encode the known hole.
     They test only selected scalar-left expressions and deliberately leave scalar-right untested. They should cover every affected operator, both directions, and all admitted proxy categories.

yingjerkao:
@ianmccul addressed — you were right, and the gap was larger than it looked. Also rebased onto master (the branch was 179 commits behind, so the previous green checks meant very little).

## Your point: the Tensor RHS

> 1. Tensor RHS remains unconstrained.

The constraint was **cosmetic** in that direction. `include/Tensor.hpp` forward-declares the same free operators that `linalg.hpp` declares, and those forward declarations were unconstrained:

```cpp
// Tensor.hpp -- before
template <class T>
Tensor operator+(const Tensor &lhs, const T &rc);
```

A differently-constrained declaration is a **distinct template**, so that unconstrained candidate stayed viable for every non-scalar `T`. `Tensor + NotScalar` still compiled straight through to the undefined reference the constraint was supposed to turn into an overload-resolution failure. Codex independently flagged the same thing.

Those forward declarations now carry the identical constraint. They had to move **below** the class definition, because `cytnx_scalar_like` names `Tensor::Tproxy` and that nested type isn't available until `Tensor` is complete. That's safe: the inline member templates that depend on them — `Tensor Mul(const T &rhs) { return *this * rhs; }`, where `operator*` has no member form — are *dependent* expressions, so they bind by ADL at the point of instantiation, not at the point of definition.

## What auditing the instantiations turned up

Checking which operand types are actually instantiated showed the single concept over-admitted in **two** directions, each accepting an operand at overload resolution only to die at link:

| concept | admits | used by |
|---|---|---|
| `cytnx_scalar_value` | dtypes + `Scalar` | `%` (Tensor + UniTensor), `==` (Tensor) |
| `cytnx_unitensor_scalar_like` | + `Scalar::Sproxy` | UniTensor `+ - * /` |
| `cytnx_scalar_like` | + both element proxies | Tensor `+ - * /` |

- `%` and `==` have **no** proxy instantiations at all — `Mod.cpp` and `Cpr.cpp` specialize only the dtype scalars and `Scalar`. This is Codex's second finding.
- The **UniTensor** operators are instantiated for `Scalar::Sproxy` but never for `Tensor::Tproxy` (`Add.cpp`), so `unitensor + tproxy` had the identical defect. Codex didn't flag this one; it's the same bug in the other object family, so per the repo guardrail on fixing the whole family I closed it too.

I verified by search that no specialization anywhere depends on a type I excluded, and retargeted declarations *and* definitions together — a mismatch between them would silently split the template again, which is the exact failure this PR is about.

## Verification

The new guards are shown to fail on the pre-fix code, not just pass on the fixed code. Reverting **only** the forward-declaration constraint and recompiling:

```
operator_constraint_test.cpp:110: error: static assertion failed
  static_assert(!cy_addable<Tensor, NotScalar>);
operator_constraint_test.cpp:111: error: static assertion failed
  static_assert(!cy_mulable<Tensor, NotScalar>);
operator_constraint_test.cpp:112: error: static assertion failed
  static_assert(!cy_subable<Tensor, NotScalar>);
operator_constraint_test.cpp:113: error: static assertion failed
  static_assert(!cy_divable<Tensor, NotScalar>);
```

Exactly those four, nothing else — so the guard covers the reported defect and no more. The stale `NOTE` that documented this as an open follow-up is gone, replaced by the assertions.

`test_main` builds and passes on `debug-openblas-cpu`; `pre-commit` clean (clang-format v14). Checked separately that pybind doesn't instantiate any of the affected operators.

Gemini's `<type_traits>` note is handled — it now applies to `Tensor.hpp`, which is where the concepts live.

yingjerkao:
@ianmccul re-review requested — your `CHANGES_REQUESTED` is the only thing left blocking this one.

Your item is fixed, and it was load-bearing: the constraint was cosmetic in the Tensor-on-the-left direction because `Tensor.hpp` forward-declared the same operators unconstrained, and a differently-constrained declaration is a *distinct* template, so the unconstrained candidate stayed viable for every non-scalar `T`.

Details, including the two related gaps the instantiation audit turned up, are in https://github.com/Cytnx-dev/Cytnx/pull/1093#issuecomment-5110832364. Short version:

- **Your item** — the `Tensor.hpp` forward declarations now carry the identical constraint. They moved below the class, since `cytnx_scalar_like` names `Tensor::Tproxy`; safe because the inline members that call them are dependent expressions bound by ADL at instantiation.
- **Concept split** — one concept was over-admitting in two directions. `%`/`==` have no proxy instantiations at all, and the UniTensor operators are instantiated for `Scalar::Sproxy` but never `Tensor::Tproxy`. Now three concepts, each admitting exactly what links.
- **The guard is verified against the bug** — reverting only the forward-declaration constraint fails exactly the four new `!cy_*able<Tensor, NotScalar>` assertions and nothing else.

Also rebased onto master (it was 179 commits behind) and now 20/20 green, including the GPU suite on the self-hosted runner.

One thing worth your eye specifically: the **UniTensor / `Tproxy` narrowing** is not something you asked for. `unitensor + tproxy` previously compiled and failed at link; it is now rejected at overload resolution. That is a deliberate behavior change on the same reasoning as your item, applied to the other object family per the repo guardrail — but if you would rather that stayed as-is, it is a two-line revert.

ianmccul:
UniTensor / (rank-0 Tproxy) should arguably work, but C++ users won't care, if Python uses another path then there is no problem.

ianmccul (approved):

PR #1095, opened 2026-07-20
refactor(linalg): pass GPU non-contiguous elementwise layout by value (#1003)

yingjerkao:
Part of #1003 (step 4: pass layout metadata as kernel params instead of per-call `cudaMalloc`).

## Problem
The non-contiguous `tensor(op)tensor` GPU elementwise kernels allocated **five** device arrays (output strides, each operand's original strides, and the two inverse mappers) via `cuMalloc_gpu`/`cuCalloc_gpu`, `cudaMemcpy`'d two of them, and `cudaFree`'d all five **on every call**. Managed-memory alloc/free is expensive and serializes with the launch, and the identical scaffolding was duplicated across the out-of-place (`cuArithmeticDispatch.cuh`) and in-place (`cuiArithmeticDispatch.cuh`) dispatches.

## Fix
New `cuNonContigLayout.cuh`:
- `GpuNonContigLayout` — a POD packing the five rank-sized arrays;
- `MakeGpuNonContigLayout(...)` — host builder (no device allocation);
- `ComputeGpuNonContigIndices(...)` — a shared `__device__` helper for the decompose→recompose index math (previously copy-pasted per kernel).

The layout is passed to the kernels **by value** (kernel constant/param space), removing all per-call device allocation and de-duplicating the triplicated scaffolding. Rank is bounded by `kGpuNonContigMaxRank` (32) with an explicit `cytnx_error_msg`; this is never the binding limit — the kernels' `512*rank` dynamic-shared-memory scratch already caps rank near 12 — and it turns a cryptic launch failure into a clear message.

## Scope
Applied to the **reachable** paths only:
- out-of-place **Add/Sub** (the ops that feed a non-contiguous operand to the shared `tn_kernel_nonconti`);
- the in-place **Add/Sub/Mul/Div** dispatch.

`Mul`/`Div`/`Cpr` still reject a non-contiguous operand at the front end (contiguous-ize first), so their non-contiguous branches are unreachable and are left untouched.

## Testing
- New `tests/gpu/linalg_test/NonContigElementwise_test.cpp`: out-of-place Add/Sub non-contiguous `tensor(op)tensor`, GPU-vs-CPU across all dtypes **and** independent hand-computed literals (including a mixed-dtype `Int32+Double` promotion), per the repo's test-value guidance.
- In-place non-contiguous path covered by the existing `InplacePromote_test`.
- On an RTX 4070 Ti (CUDA 13, sm_89): `gpu_test_main` builds clean; the new + in-place tests pass; the full `Add/Sub/Mul/Div/Cpr` GPU regression (**360 tests**) is green. clang-format-14 clean.

Note: GPU results are from the local RTX 4070 Ti — there is no GPU CI runner. A quantitative before/after benchmark of the removed device allocations is a reasonable follow-up.

### DIFF
--- a/src/backend/linalg_internal_gpu/cuNonContigLayout.cuh
+++ b/src/backend/linalg_internal_gpu/cuNonContigLayout.cuh
@@ -8,45 +8,49 @@
 
 // Shared layout metadata for the non-contiguous tensor(op)tensor GPU elementwise
 // kernels (arithmetic out-of-place, arithmetic in-place, and comparison). Every
-// such kernel needs, per tensor dimension, the output strides (accu_shape), each
-// operand's original strides (old_accu_shape{L,R}) and permutation (invmapper_{L,R})
-// so it can turn a linear index into the two operands' physical offsets.
+// such kernel needs, per output dimension, the output stride (to decompose a
+// linear index into the output coordinate) and each operand's stride in its own
+// physical buffer for that output dimension (to gather that operand's element).
 //
 // Previously each launch cuMalloc_gpu'd five device arrays, cudaMemcpy'd two of
 // them, and cudaFree'd all five on *every* call -- managed-memory alloc/free is
 // expensive and fully serial with the launch (#1003 step 4). Instead we pack the
-// five rank-sized arrays into one POD passed to the kernel BY VALUE, so it rides in
+// per-dimension strides into one POD passed to the kernel BY VALUE, so it rides in
 // the kernel's constant/param space with no device allocation at all.
 //
-// The five arrays are bounded by kGpuNonContigMaxRank, which is the shared-memory
-// limit itself: each non-contiguous kernel stages a per-thread scratch of 512 * rank
-// cytnx_uint64 in dynamic shared memory, so 512 * rank * sizeof(cytnx_uint64) must fit
-// the 48 KB default per-block budget -> rank <= 12. Building a layout for a higher rank
-// now raises a clear error here instead of letting the kernel launch fail cryptically.
+// The layout stores per-output-dimension operand strides directly rather than the
+// operands' permutations + original strides: precomputing strideL/strideR on the
+// host (`strideL[invmapper_L[d]] = old_stride_of_L_at_d`) folds the operand offset
+// straight into the index-decompose loop. That removes the inverse-mapper
+// indirection and, crucially, the per-thread shared-memory multi-index scratch the
+// old recompose needed -- so the kernels launch with no dynamic shared memory and
+// the rank is bounded only by the by-value struct's parameter-space footprint.
 
 namespace cytnx {
   namespace linalg_internal {
     namespace gpu_layout {
 
-      // 512 (threads/block) * kGpuNonContigMaxRank * sizeof(cytnx_uint64) == 48 KB, the
-      // default dynamic-shared-memory budget the non-contiguous kernels stage tmpv in.
-      constexpr int kGpuNonContigMaxRank = 12;
+      // Bounds the three rank-sized arrays carried by value in GpuNonContigLayout.
+      // The struct is 3 * kGpuNonContigMaxRank * sizeof(cytnx_uint64) + 8 B (= 776 B
+      // at 32), well within the >=4 KB CUDA kernel parameter budget. (No longer
+      // bounded by shared memory: the strided decompose uses none.)
+      constexpr int kGpuNonContigMaxRank = 32;
 
       struct GpuNonContigLayout {
-        cytnx_uint64 accu_shape[kGpuNonContigMaxRank];
-        cytnx_uint64 old_accu_shapeL[kGpuNonContigMaxRank];
-        cytnx_uint64 old_accu_shapeR[kGpuNonContigMaxRank];
-        cytnx_uint64 invmapper_L[kGpuNonContigMaxRank];
-        cytnx_uint64 invmapper_R[kGpuNonContigMaxRank];
+        cytnx_uint64 accu_shape[kGpuNonContigMaxRank];  // output stride per output dim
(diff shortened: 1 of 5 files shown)


yingjerkao:
Thanks — both addressed in the latest commit:

- **Rank cap (Gemini):** `kGpuNonContigMaxRank` 32 → **12**. As you noted, the per-thread scratch is `512 × rank × sizeof(cytnx_uint64)`, which hits the 48 KB default block budget at rank 12 — so 12 *is* the real limit, and `MakeGpuNonContigLayout` now rejects a higher rank with a clear error instead of letting the kernel launch fail cryptically (also shrinks the by-value struct to 488 B). This is the same guard the Cpr PRs' bots asked for.

- **invmapper validation (Gemini):** `MakeGpuNonContigLayout` now checks `invmapper_L.size() == rank && invmapper_R.size() == rank` before indexing them, guarding against an out-of-bounds host read.

src/backend/linalg_internal_gpu/cuNonContigLayout.cuh
```diff
@@ -0,0 +1,107 @@
...
+          accuL *= shape[invmapper_L[d]];
+          layout.old_accu_shapeR[d] = accuR;
+          accuR *= shape[invmapper_R[d]];
+        }
+        for (cytnx_uint64 j = 0; j < rank; j++) {
+          layout.invmapper_L[j] = invmapper_L[j];
+          layout.invmapper_R[j] = invmapper_R[j];
+        }
+        return layout;
+      }
```
IvanaGyro: The number of operations needed here is almost the same as calculating strides from invmapper and shape.

src/backend/linalg_internal_gpu/cuNonContigLayout.cuh
```diff
@@ -0,0 +1,107 @@
...
+        cytnx_uint64 invmapper_R[kGpuNonContigMaxRank];
+        cytnx_uint64 shapesize;
+      };
+
+      // Host-side builder: fills the layout from the output shape and the two operand
+      // inverse mappers (identical arithmetic to the old per-launch stride loop, but
+      // writing into the by-value struct instead of five freshly-allocated device
+      // buffers). `shape` has one entry per dimension; empty shape means the operands
+      // are contiguous and the caller must not take the non-contiguous path.
+      inline GpuNonContigLayout MakeGpuNonContigLayout(
```
IvanaGyro: New functions should use snake cases?

tests/gpu/linalg_test/NonContigElementwise_test.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+      // Independent literal expected values (NOT the CPU oracle) for the out-of-place
+      // non-contiguous mapper math, per CLAUDE.md: a bug in the shared gather would
+      // otherwise be able to hide behind a GPU-vs-CPU comparison. Both operands are
+      // permuted, so both Lidx and Ridx are non-trivial.
+      TEST(NonContigElementwiseTest, GpuOutOfPlaceNoncontiguousLiteral) {
+        // l logical = [[1,3,5],[2,4,6]], r logical = [[10,30,50],[20,40,60]]
+        // (arange 3x2 permuted).
+        Tensor l = arange(1, 7, 1, Type.Int64).reshape({3, 2}).permute({1, 0}).to(Device.cuda);
+        Tensor r = arange(10, 70, 10, Type.Int64).reshape({3, 2}).permute({1, 0}).to(Device.cuda);
```
IvanaGyro, line 64: Why it only tests with int64?

IvanaGyro (approved):

src/backend/linalg_internal_gpu/cuNonContigLayout.cuh
yingjerkao: I think moving to strides view will be the optimal solution

src/backend/linalg_internal_gpu/cuNonContigLayout.cuh
yingjerkao: Done in c70270dd — renamed to `make_gpu_non_contig_layout` and `compute_gpu_non_contig_indices` (snake_case per the Cytnx new-code convention).

tests/gpu/linalg_test/NonContigElementwise_test.cpp
yingjerkao: Added Double (8 B) and ComplexDouble (16 B) independent-literal cases to `GpuOutOfPlaceNoncontiguousLiteral` in c70270dd. The gather math (Lidx/Ridx) is dtype-independent, but exercising a wider element and the complex path keeps a typed-pointer / element-size bug from hiding behind Int64.

src/backend/linalg_internal_gpu/cuNonContigLayout.cuh
yingjerkao: Switched to the strides view in c70270dd: the layout now stores per-output-dimension operand strides (`strideL`/`strideR`), precomputed on the host (`strideL[invmapper_L[d]] = operand stride at d`), so the device recompose folds into the index-decompose loop in a single pass — no inverse-mapper indirection and, notably, no per-thread shared-memory scratch. The kernels now launch with **zero dynamic shared memory**, so the old `512 * rank <= 48 KB -> rank <= 12` limit is gone (cap raised to 32, now bounded only by the by-value struct's parameter footprint). Lidx/Ridx are algebraically identical to before.

Note this is an **interim**, kernel-side stride precompute — not `Storage`'s mdspan strided view. The mdspan strided-`Storage` refactor will supersede `GpuNonContigLayout` entirely; this just avoids leaving the reconstruction+shared-memory version in place until then.

PR #1096, opened 2026-07-20
fix(linalg): support non-contiguous tensor⊗tensor for GPU Mul/Div (#1003, #988)

### LINKED ISSUE #1013: Typed Storage dispatch: GPU conversion

yingjerkao:
Part of #1003; closes the GPU non-contiguous gap tracked in #988 for Mul/Div.

## Problem
The GPU out-of-place **Mul** and **Div** front ends rejected a non-contiguous operand with `two tensors must be contiguous. Call Contiguous_() or Contiguous() first` — forcing the caller to contiguous-ize first, even though **Add/Sub accept it** and the **CPU path handles it**. This was a leftover from the pre-#1013 era when `cuMul`/`cuDiv` had no working non-contiguous kernel (#988).

## Fix
#1013 unified all four arithmetic ops onto the shared `cuArithmeticDispatchGPU` kernel, which already applies the layout mappers. `Mul.cpp`/`Div.cpp` now pass the layout (`shape` + inverse mappers) to `cuMul_dispatch`/`cuDiv_dispatch` in the non-contiguous branch — **exactly as `Add.cpp`/`Sub.cpp` already do** — instead of erroring. Div's output is already allocated with the true-division (floating) dtype before the branch.

⚠️ **Behavior change (flagged):** GPU `Mul`/`Div` now **accept** a non-contiguous operand and return the correct result (matching CPU and Add/Sub) rather than throwing. This removes a restriction; it does not change any result that previously succeeded.

## Testing
New non-contiguous `tensor(op)tensor` cases in `Mul_test.cpp` and `Div_test.cpp`:
- **GPU-vs-CPU** across all dtypes (both operands permuted, so both offsets are gathered through the inverse mappers);
- **independent hand-computed literals** with *distinct per-element* results, so a wrong multi-index decomposition is caught, not just a wrong L/R pairing.

Both new tests **fail on the pre-fix code** (which threw "must be contiguous"). On an RTX 4070 Ti (CUDA 13, sm_89): `gpu_test_main` builds clean; the new tests pass; the full **Mul/Div GPU suites (148 tests) are green**. clang-format-14 clean.

Composes with (and is independent of) #1095 — when both land, this newly-live non-contiguous path uses the by-value layout instead of per-call device allocation.

Note: GPU results are from the local RTX 4070 Ti — there is no GPU CI runner. `Cpr` still rejects non-contiguous operands (separate Bool-output kernel); left as a follow-up.

### DIFF
--- a/src/linalg/Div.cpp
+++ b/src/linalg/Div.cpp
@@ -95,7 +95,7 @@ namespace cytnx {
           checkCudaErrors(cudaSetDevice(Rt.device()));
           linalg_internal::cuDiv_dispatch(out._impl->storage()._impl, left._impl->storage()._impl,
                                           right._impl->storage()._impl,
-                                          left._impl->storage()._impl->size(), left._impl->shape(),
+                                          out._impl->storage()._impl->size(), left._impl->shape(),
                                           left._impl->invmapper(), right._impl->invmapper());
   #else
           cytnx_error_msg(true, "[Div] fatal error, the tensor is on GPU without CUDA support.%s",
--- a/src/linalg/Div.cpp
+++ b/src/linalg/Div.cpp
@@ -88,10 +88,15 @@ namespace cytnx {
             Lt._impl->storage().as_storage_variant(), Rt._impl->storage().as_storage_variant());
         } else {
   #ifdef UNI_GPU
-          cytnx_error_msg(true,
-                          "[Div][on GPU/CUDA] error two tensors must be contiguous. Call "
-                          "Contiguous_() or Contiguous() first%s",
-                          "\n");
+          // Non-contiguous tensor(/)tensor on the GPU: the shared cuArithmeticDispatchGPU
+          // kernel applies the layout mappers (as Add/Sub already do), so pass the layout
+          // instead of forcing the caller to contiguous-ize first (#1003, #988). `out` is
+          // already allocated with the true-division (floating) dtype above.
+          checkCudaErrors(cudaSetDevice(Rt.device()));
+          linalg_internal::cuDiv_dispatch(out._impl->storage()._impl, left._impl->storage()._impl,
+                                          right._impl->storage()._impl,
+                                          left._impl->storage()._impl->size(), left._impl->shape(),
+                                          left._impl->invmapper(), right._impl->invmapper());
   #else
           cytnx_error_msg(true, "[Div] fatal error, the tensor is on GPU without CUDA support.%s",
                           "\n");
--- a/src/linalg/Mul.cpp
+++ b/src/linalg/Mul.cpp
@@ -76,7 +76,7 @@ namespace cytnx {
           checkCudaErrors(cudaSetDevice(Rt.device()));
           linalg_internal::cuMul_dispatch(out._impl->storage()._impl, left._impl->storage()._impl,
                                           right._impl->storage()._impl,
-                                          left._impl->storage()._impl->size(), left._impl->shape(),
+                                          out._impl->storage()._impl->size(), left._impl->shape(),
                                           left._impl->invmapper(), right._impl->invmapper());
   #else
           cytnx_error_msg(true, "[Mul] fatal error, the tensor is on GPU without CUDA support.%s",
--- a/src/linalg/Mul.cpp
+++ b/src/linalg/Mul.cpp
@@ -70,10 +70,14 @@ namespace cytnx {
             Lt._impl->storage().as_storage_variant(), Rt._impl->storage().as_storage_variant());
         } else {
   #ifdef UNI_GPU
-          cytnx_error_msg(true,
-                          "[Mul][on GPU/CUDA] error two tensors must be contiguous. Call "
-                          "Contiguous_() or Contiguous() first%s",
-                          "\n");
+          // Non-contiguous tensor(x)tensor on the GPU: the shared cuArithmeticDispatchGPU
+          // kernel applies the layout mappers (as Add/Sub already do), so pass the layout
+          // instead of forcing the caller to contiguous-ize first (#1003, #988).
+          checkCudaErrors(cudaSetDevice(Rt.device()));
+          linalg_internal::cuMul_dispatch(out._impl->storage()._impl, left._impl->storage()._impl,
+                                          right._impl->storage()._impl,
+                                          left._impl->storage()._impl->size(), left._impl->shape(),
+                                          left._impl->invmapper(), right._impl->invmapper());
   #else
           cytnx_error_msg(true, "[Mul] fatal error, the tensor is on GPU without CUDA support.%s",
                           "\n");
--- a/tests/gpu/linalg_test/Div_test.cpp
+++ b/tests/gpu/linalg_test/Div_test.cpp
@@ -238,6 +238,47 @@ namespace cytnx {
 
       INSTANTIATE_TEST_SUITE_P(DivTests, DivTestAllShapes, ::testing::ValuesIn(GetTestShapes()));
 
+      // Non-contiguous tensor(/)tensor on the GPU (#1003, #988): the GPU front end used to
+      // reject a non-contiguous operand ("must be contiguous"); it now feeds the layout to
+      // the shared dispatch kernel like Add/Sub. Both operands are permuted so both operand
+      // offsets are gathered through the inverse mappers. Div is true division (floating
+      // output). Nonzero denominators (arange from 1) avoid divide-by-zero. Compare against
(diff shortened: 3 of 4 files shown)


yingjerkao:
Addressed in the latest commit: the `cuMul_dispatch`/`cuDiv_dispatch` length argument now uses `out._impl->storage()._impl->size()` (the logical element count), matching the CPU non-contiguous path and the contiguous GPU branch, rather than `left._impl->storage()._impl->size()`.

This is the same correctness fix as the critical one flagged on #1097 — see that thread for the verification that it isn't currently triggerable in Cytnx (slices compact to contiguous, so a non-contiguous tensor's storage always equals its logical size) but is applied for correctness and future-proofing. The rank-13+ shared-memory guard is handled centrally in #1095.

ianmccul (approved):
OK, but noting that the current implementation is very slow; much slower than making it contiguous (assuming the algorithm for making it contiguous is effiicient -- I didn't check that)

yingjerkao:
@ianmccul I measured this, because your parenthetical turned out to be the whole question:

> OK, but noting that the current implementation is very slow; much slower than making it contiguous (assuming the algorithm for making it contiguous is effiicient -- I didn't check that)

**The direct path is faster, by 2.8×–23×.** The assumption in your parenthetical is where it breaks: `contiguous()` on the GPU is *not* efficient.

### Numbers

Mean of 5 repetitions, µs. RTX 4070 Ti SUPER (sm_89), CUDA 13.0, cuTENSOR 2.4.1, `Release`, `USE_CUTENSOR=ON USE_CUQUANTUM=ON`. `Double` cubes, permuted `{2,0,1}` so neither operand is contiguous. All four columns are the **same binary at this branch tip (`50e4ef86`)** — the question is about two ways of calling one build, not two revisions.

| n (cube) | elements | **Direct (this PR)** | contiguous-first | already-contiguous floor | `contiguous()` alone |
|---|---:|---:|---:|---:|---:|
| 64  | 262 144    | **313**    | 7 315   | 287    | 3 588 |
| 128 | 2 097 152  | **2 180**  | 11 084  | 2 043  | 5 478 |
| 256 | 16 777 216 | **18 169** | 56 036  | 17 011 | 21 647 |
| 384 | 56 623 104 | **61 768** | 173 608 | 56 579 | 61 078 |

The gather through the inverse mappers costs only **+6.7% to +9.2%** over operating on already-contiguous data. It is nearly free, so it can't be the bottleneck.

### Not fast because it's skipping work

Checked before trusting the timings: direct vs contiguous-first vs an independent CPU reference at n = 8/17/64 — **bit-exact, zero difference** on all elements.

### Why contiguous-first loses structurally

Count array passes. Direct: 2 reads + 1 write = **3**. Contiguous-first: 2 copies (4 passes) + the multiply (3) = **7**. That 2.33× traffic ratio predicts the measured 2.81× asymptote at n=384, and it's bandwidth-independent — no kernel tuning closes it.

On top of that, `contiguous()` has a **~3.4 ms fixed cost per call, independent of size** — an 8-element tensor takes 3 510 µs. Attributed rather than guessed:

```
cutensorCreate + cutensorDestroy only : 3446.6 us/call
full handle + descs + plan + destroy  : 3375.5 us/call
cudaMallocManaged(4KB) + cudaFree     :   24.4 us/call
```

`cuMovemem_cutensor_gpu` builds **and destroys a complete cuTENSOR handle on every call** (`src/backend/utils_internal_gpu/cuMovemem_gpu.cu:234-271`). The descriptors and plan are free; the handle is essentially all of it. Filed separately as #1132 — it's independent of this PR and would speed up every `permute().contiguous()` in the library.

### Caveats

- Both paths run at only ~15–24 GB/s on a ~670 GB/s card, because Cytnx allocates GPU storage with `cudaMallocManaged` (`cuAlloc_gpu.cu:19,26`), so everything here is PCIe-migration-bound. This depresses the absolute numbers but not the ranking, which is set by array passes. With device-resident memory the gap would if anything widen — contiguous-first keeps 5 large buffers live to direct's 3.
- Scope: one dtype (`Double`), one permutation family, one GPU.
- The pre-existing GPU benchmarks in `benchmarks/` don't call `cudaDeviceSynchronize()`, so they time kernel *launches*, not execution. Every timed region above syncs; those numbers aren't comparable to these.

The benchmark is a committed Google Benchmark file (`benchmarks/linalg/Mul_noncontig_bm.cpp`, 4 cases incl. the two controls). I've deliberately **not** pushed it onto this branch, so as not to disturb your approval — happy to land it as its own PR if you'd like it in the tree.

PR #1097, opened 2026-07-20
fix(linalg): support non-contiguous tensor(==)tensor for GPU Cpr (#1003, #988)

yingjerkao:
Part of #1003. With #1096 this closes the GPU non-contiguous gap (#988) for **all five** elementwise ops.

## Problem
The GPU out-of-place **Cpr** front end rejected a non-contiguous operand with `two tensors must be contiguous. Call Contiguous_() or Contiguous() first` — the last GPU elementwise op to do so (Add/Sub already accepted it; Mul/Div were enabled in #1096) — even though the CPU path **and** the `cuCpr_dispatch` non-contiguous Bool-output kernel already handle it.

## Fix
`Cpr.cpp` now passes the layout (`shape` + inverse mappers) to `cuCpr_dispatch` in the non-contiguous branch, mirroring `Add`/`Sub`/`Mul`/`Div`, instead of erroring. `cuCpr` already had a working non-contiguous kernel — it was simply unreachable.

⚠️ **Behavior change (flagged):** GPU `Cpr` now **accepts** a non-contiguous operand and returns the correct `Bool` result (matching CPU) rather than throwing. This removes a restriction; it does not change any result that previously succeeded.

## Testing
New `CprNonContig` tests in `Cpr_test.cpp`:
- **permuted-vs-contiguous in both operand orders**, so each inverse mapper (`invmapper_L` and `invmapper_R`) is exercised;
- **GPU-vs-CPU** across all dtypes;
- an **independent hand-computed identity-pattern literal** — `a[i][j]=3j+i` (permuted) vs `b[i][j]=3i+j` (contiguous) are equal iff `i==j`, a mix a wrong gather would break.

Both new tests **fail on the pre-fix code** (which threw "must be contiguous"). On an RTX 4070 Ti (CUDA 13, sm_89): `gpu_test_main` builds clean; the new tests pass; the full **Cpr GPU suite (82 tests) is green**. clang-format-14 clean.

Note: GPU results are from the local RTX 4070 Ti — there is no GPU CI runner. The `cuCpr` non-contiguous kernel still uses per-call device allocation; applying the #1095 by-value layout to it (now that this path is live) is a natural follow-up.

### DIFF
--- a/src/linalg/Cpr.cpp
+++ b/src/linalg/Cpr.cpp
@@ -76,7 +76,7 @@ namespace cytnx {
           checkCudaErrors(cudaSetDevice(Rt.device()));
           cytnx::linalg_internal::cuCpr_dispatch(
             out._impl->storage()._impl, left._impl->storage()._impl, right._impl->storage()._impl,
-            left._impl->storage()._impl->size(), left._impl->shape(), left._impl->invmapper(),
+            out._impl->storage()._impl->size(), left._impl->shape(), left._impl->invmapper(),
             right._impl->invmapper());
   #else
           cytnx_error_msg(true, "[Cpr] fatal error, the tensor is on GPU without CUDA support.%s",
--- a/src/linalg/Cpr.cpp
+++ b/src/linalg/Cpr.cpp
@@ -70,10 +70,14 @@ namespace cytnx {
             Lt._impl->storage().as_storage_variant(), Rt._impl->storage().as_storage_variant());
         } else {
   #ifdef UNI_GPU
-          cytnx_error_msg(true,
-                          "[Cpr][on GPU/CUDA] error two tensors must be contiguous. Call "
-                          "Contiguous_() or Contiguous() first%s",
-                          "\n");
+          // Non-contiguous tensor(==)tensor on the GPU: cuCpr_dispatch's non-contiguous kernel
+          // applies the layout mappers (as Add/Sub/Mul/Div already do), so pass the layout
+          // instead of forcing the caller to contiguous-ize first (#1003, #988).
+          checkCudaErrors(cudaSetDevice(Rt.device()));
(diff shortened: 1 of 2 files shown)


yingjerkao:
Thanks — addressed in the latest commit.

**Length fix (Gemini, critical):** switched the `cuCpr_dispatch` length argument to `out._impl->storage()._impl->size()` (the logical element count), matching the CPU non-contiguous path (`Add.cpp:120`) and the contiguous GPU branch.

On triggerability, I verified the specific counterexample (a strided view whose storage is larger than its logical size) and it does **not** currently occur in Cytnx:

| construction | contiguous? | storage vs logical |
|---|---|---|
| `permute` | no | equal |
| `slice [1:3,2:5]` | **yes (compacts)** | equal |
| `permute`→`slice` | no | equal |

Slicing compacts to a contiguous tensor, and non-contiguity only arises from `permute`, which never enlarges storage — so `left` and `out` sizes coincide today and there is no live out-of-bounds write. The fix is still correct, consistent with the CPU path, and future-proof, so it's applied. (The same latent pattern exists in Add/Sub on master; I fixed the Mul/Div counterparts in #1096.)

**Rank-13+ shared-memory guard (both bots):** addressed centrally in #1095 — `kGpuNonContigMaxRank` is now 12 (the actual limit: `512 threads × rank × sizeof(uint64) ≤ 48 KB`), so `MakeGpuNonContigLayout` raises a clear error for a rank the kernel cannot launch instead of failing cryptically. The by-value follow-up #1098 carries that guard for `cuCpr` too.

yingjerkao:
Superseded by #1098, which was built on top of this branch and merged to `master` (commit 014a3860) — every commit here is now an ancestor of `master`, including the fix for the non-contiguous-length OOB that @gemini-code-assist flagged (the merged code uses the output's logical size, not `left`'s storage size). Closing as already-merged.

PR #1099, opened 2026-07-20
fix(linalg): register missing Int16/Uint16/Bool rows in CPU Outer dispatch (fixes segfault)

yingjerkao:
## Problem

`linalg::Outer` **segfaults** for `Int16`, `Uint16`, or `Bool` inputs on the CPU.

`Outer.cpp` promotes both operands to a common dtype and dispatches through
`Outer_ii[out_dtype][out_dtype]`. The CPU `Outer_ii` dispatch table in
`linalg_internal_interface.cpp` was never populated with the **Int16, Uint16,
or Bool source rows** — every other source dtype has a full 11-entry row, but
these three had none. So when the promoted dtype is `Int16`/`Uint16`/`Bool`,
`Outer_ii[out_dtype][out_dtype]` is a null function pointer, and the call jumps
to address `0`:

```
AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x000000000000 ...)
    #1 ... TestBody tests/linalg_test/Outer_test.cpp:32   // the linalg::Outer(a, b) call
```

Every sibling dispatch table in the same constructor — `MM_ii`, `Sum_ii`,
`Diag_ii`, `Matmul_ii`, `Matvec_ii`, and even the GPU `cuOuter_ii` — already
registers these dtypes, so the omission is an oversight, not intent. (The GPU
path works; this is CPU-only.)

Reproduces on `master` (86cb96c2). Minimal repro:

```python
import cytnx
a = cytnx.zeros([5], cytnx.Type.Int16); b = cytnx.zeros([7], cytnx.Type.Int16)
cytnx.linalg.Outer(a, b)   # SIGSEGV
```

## Fix

Register the 33 missing rows (`Int16`, `Uint16`, `Bool` sources × 11 target
dtypes). The `Outer_internal_{i16,u16,b}t*` implementations already exist and
are declared — **this is registration only, no kernel or algorithm change.**

## Testing

- New `tests/linalg_test/Outer_test.cpp` (`OuterDtypeCoverage`: `Int16`,
  `Uint16`, `Bool`, `Int16xBool`) checks independently hand-computed values
  (dtype **and** value, with negative and boolean cases).
- **Regression discipline:** on the pre-fix binary these crash with the ASan
  SEGV above at the `linalg::Outer` call site; post-fix all four pass.
- Existing `DtypePromotion.OuterComplexfloatDouble` and `LinalgKronTest` tests
  still pass. `clang-format` (v14) clean.

### DIFF
--- a/src/backend/linalg_internal_interface.cpp
+++ b/src/backend/linalg_internal_interface.cpp
@@ -324,6 +324,47 @@ namespace cytnx {
       Outer_ii[Type.Uint32][Type.Int16] = Outer_internal_u32ti16;
       Outer_ii[Type.Uint32][Type.Bool] = Outer_internal_u32tb;
 
+      // The Int16/Uint16/Bool source rows were omitted from this table, so
+      // Outer_ii[Int16/Uint16/Bool][...] were null; linalg::Outer with such a
+      // left operand dereferenced a null function pointer and crashed. The
+      // Outer_internal_{i16,u16,b}t* implementations already exist -- register
+      // them here so every dtype pair dispatches. (cuOuter_ii already had them.)
+      Outer_ii[Type.Int16][Type.ComplexDouble] = Outer_internal_i16tcd;
+      Outer_ii[Type.Int16][Type.ComplexFloat] = Outer_internal_i16tcf;
+      Outer_ii[Type.Int16][Type.Double] = Outer_internal_i16td;
+      Outer_ii[Type.Int16][Type.Float] = Outer_internal_i16tf;
+      Outer_ii[Type.Int16][Type.Int64] = Outer_internal_i16ti64;
+      Outer_ii[Type.Int16][Type.Uint64] = Outer_internal_i16tu64;
+      Outer_ii[Type.Int16][Type.Int32] = Outer_internal_i16ti32;
+      Outer_ii[Type.Int16][Type.Uint32] = Outer_internal_i16tu32;
+      Outer_ii[Type.Int16][Type.Uint16] = Outer_internal_i16tu16;
+      Outer_ii[Type.Int16][Type.Int16] = Outer_internal_i16ti16;
+      Outer_ii[Type.Int16][Type.Bool] = Outer_internal_i16tb;
+
+      Outer_ii[Type.Uint16][Type.ComplexDouble] = Outer_internal_u16tcd;
+      Outer_ii[Type.Uint16][Type.ComplexFloat] = Outer_internal_u16tcf;
+      Outer_ii[Type.Uint16][Type.Double] = Outer_internal_u16td;
+      Outer_ii[Type.Uint16][Type.Float] = Outer_internal_u16tf;
+      Outer_ii[Type.Uint16][Type.Int64] = Outer_internal_u16ti64;
+      Outer_ii[Type.Uint16][Type.Uint64] = Outer_internal_u16tu64;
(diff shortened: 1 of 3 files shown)


IvanaGyro (approved):

IvanaGyro (approved):
The comments from bots are worth to be resolved. Except that LGTM.

ianmccul:
Outer.cpp casts both operands to `out_dtype,` then calls `Outer_ii[out_dtype][out_dtype]`. Only the three missing diagonal entries are reachable; the other 30 registrations are harmless but dead. Likewise, `Int16TimesBool` does not exercise `Outer_ii[Int16][Bool]`.

I'm wondering why this PR doesn't replace the whole mechanism by a much simpler implementation.  src/linalg/Outer.cpp:24 rejects either operand unless `shape().size() == 1`. There is only a `Tensor` overload. So:
```c++
Tensor Outer(const Tensor &lhs, const Tensor &rhs) {
  cytnx_error_msg(lhs.is_void(), "[ERROR] lhs must be initialized.%s", "\n");
  cytnx_error_msg(rhs.is_void(), "[ERROR] rhs must be initialized.%s", "\n");
  cytnx_error_msg(lhs.device() != rhs.device(),
                  "[ERROR] the two tensors must be on the same device.%s", "\n");
  cytnx_error_msg(lhs.rank() != 1, "[ERROR] tensor #1 should have rank-1.%s", "\n");
  cytnx_error_msg(rhs.rank() != 1, "[ERROR] tensor #2 should have rank-1.%s", "\n");

  return Kron(lhs, rhs).reshape(std::vector<cytnx_uint64>{lhs.shape()[0], rhs.shape()[0]});
}
```

ianmccul (changes requested):
see top-level comment above

yingjerkao:
Addressing @ianmccul's review comment: rather than register the missing Int16/Uint16/Bool rows, #1105 reimplements `Outer` as `Kron(a, b).reshape({m, n})` and retires the per-dtype `Outer_ii`/`cuOuter_ii` dispatch entirely. `Kron`'s variant-based dispatch already covers those dtype pairs on CPU and GPU, so the missing-row segfault this PR fixed can no longer occur. This PR is superseded by #1105.

ianmccul (approved):

ianmccul (commented):
hmm, not sure how to 'unsubmit' approval but it doesn't matter, I assume this PR is no longer relevant

yingjerkao:
Closing as **superseded by #1105**.

This PR registered the missing `Outer_ii[Int16/Uint16/Bool][…]` rows in the CPU `Outer` dispatch table to fix a segfault on those source dtypes. Since it was approved, #1105 landed and reworked `Outer` as `Kron(a, b).reshape(...)`, deleting the `Outer_ii`/`Outerfunc_oii`/`cuOuter_ii` dispatch tables entirely.

As a result, on current `master`:
- **The segfault is already fixed** — `Outer` now routes through `Kron`, whose variant-based `type_promote_from_pointer_t` dispatch covers every dtype pair, so there's no null dispatch entry to crash on. Verified: `Outer(Int16/Uint16/Bool, Double)` and `Outer(Bool, Bool)` all return correct results.
- **The dispatch registration this PR adds has nowhere to go** — the table it patches no longer exists.
- **The dtype coverage is already tested** — `tests/linalg_test/Outer_test.cpp` on `master` (added by #1105) exercises the Int16/Uint16/Bool cases.

Nothing here is left to salvage, so closing rather than rebasing. Thanks!

PR #1100, opened 2026-07-20
test(linalg): GPU correctness coverage for Kron and Outer (#1003, confirms #999)

yingjerkao:
## Problem

`Kron` and `Outer` had **no GPU tests** — their CUDA paths (`cuKron_internal.cuh`, `cuOuter_internal.cu`) were entirely uncovered. This is the coverage the #1003 Outer/Kron typed-dispatch consolidation needs as a safety net, and it also confirms #999 is resolved.

## Fix

Add `tests/gpu/linalg_test/Kron_test.cpp` and `Outer_test.cpp`. Expected values are computed **independently** from the operation definition (literals or per-element recompute), not by comparing against another Cytnx path that could share the same bug:

- **Double / ComplexDouble** — hand-written literal results (Kron also covers fractional + negative values).
- **Mixed `ComplexFloat × Double → ComplexDouble`** — the promotion discriminator behind #999/#984: the product must be computed and stored through the promoted output type. Checks output **dtype and value**.
- **Broad GPU-vs-CPU cross-check** over every real/complex dtype as *secondary* coverage. `ComplexFloat` uses a looser tolerance because ~1e6-magnitude float32 complex products diverge from the CPU at the ULP; a single real multiply is bit-identical.

Inputs are built on the CPU and moved with `.to(Device.cuda)` (GPU `arange` ignores a non-unit step for real dtypes, #1070).

## Testing

All **9 tests pass** on an RTX 4070 Ti SUPER (CUDA 13.0, `sm_89`):

```
[  PASSED  ] 9 tests.   (GpuKron: 5, GpuOuter: 4)
```

Verified against a build with the parent commit's `Outer_ii` fix applied (the Outer cross-check's CPU reference needs the Int16/Uint16 dispatch rows). The `cuKron`/`cuOuter`/`Kron.cpp`/`Outer.cpp` sources under test are identical to `origin/master`.

## Notes

- **Stacked on** #1099 (the CPU `Outer` Int16/Uint16/Bool dispatch fix). The Outer GPU cross-check's CPU reference crashes without it. Will retarget to `master` once #1099 merges.
- No GPU CI runner exists, so these compile/run only under a local `USE_CUDA` build (as done here).

Part of #1003. Confirms #999.

### DIFF
--- a/tests/gpu/CMakeLists.txt
+++ b/tests/gpu/CMakeLists.txt
@@ -19,8 +19,10 @@ add_executable(
   linalg_test/Add_test.cpp
   linalg_test/Cpr_test.cpp
   linalg_test/Div_test.cpp
+  linalg_test/Kron_test.cpp
   linalg_test/Mod_test.cpp
   linalg_test/Mul_test.cpp
+  linalg_test/Outer_test.cpp
   linalg_test/Sub_test.cpp
   linalg_test/InplacePromote_test.cpp
   linalg_test/Det_test.cpp
--- a/tests/gpu/linalg_test/Kron_test.cpp
+++ b/tests/gpu/linalg_test/Kron_test.cpp
@@ -1,0 +1,167 @@
+#include "gtest/gtest.h"
+
+#include "cytnx.hpp"
+#include "gpu_test_tools.h"
+
+// GPU Kron correctness. Every expected value here is computed independently
+// from the Kronecker-product definition
+//
+//     out[i, j] = A[i / bR, j / bC] * B[i % bR, j % bC]
+//
+// (A is aR x aC, B is bR x bC, out is (aR*bR) x (aC*bC)), written out as
+// literals or recomputed from the raw inputs -- never by comparing against
+// another Cytnx path that could share the same bug. The mixed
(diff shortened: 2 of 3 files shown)


yingjerkao:
Closing — its GPU test coverage has been **salvaged into #1108**, and the rest is superseded by #1105/#1106.

Background: this PR was stacked on the (now-closed) #1099 branch, and the `Outer`→`Kron` rewrite + GPU test files it depended on already landed on `master` via #1105/#1106 — so it can't merge as-is (it would try to re-apply changes already present and conflict).

But its tests were more thorough than what landed, so rather than lose them, #1108 folds this PR's stronger cases into `master`'s existing `tests/gpu/linalg_test/{Kron,Outer}_test.cpp`:
- 2-D hand-computed Kron/Outer (fractional + negative values),
- complex × complex hand-computed cases,
- the ComplexFloat × Double → ComplexDouble promotion discriminator,
- the broad GPU-vs-CPU all-dtype differential (`GpuMatchesCpuAllDtypes`),

while keeping master's Int16 hand-computed Kron and the #1099 Int16-diagonal Outer segfault guard. Verified on GPU (CUDA 13, RTX 4070 Ti): 11/11 pass. Thanks!

Issue #916, opened 2026-06-18
Type Scalar-accepting stub parameters as a shared ScalarLike union


IvanaGyro:
## Summary

Several `linalg` functions accept a `cytnx::Scalar` parameter that, at runtime,
also accepts plain Python and NumPy scalars via implicit `Scalar` conversions.
pybind11-stubgen renders these as bare `cytnx.cytnx.Scalar`, so a static type
checker rejects working calls like `Axpy(1, x)` even though they run fine.

`ExpH`/`ExpM` were given a local `ScalarLike` alias in #915 (because collapsing
their per-dtype overloads to a single `Scalar` overload would otherwise have
narrowed their stubs). The remaining bare-`Scalar` stubs are pre-existing from
the stubs PR (#912) and should be unified.

## Affected (bare `Scalar` in current stubs)

- `linalg.Axpy`, `linalg.Axpy_`
- `linalg.Gemm`, `linalg.Gemm_`
- `linalg.Ger`
- (any future bindings rebound to a single `Scalar` overload, e.g. the planned
  Tensor/UniTensor operator rebind)

## Proposal

Promote the `ScalarLike` alias introduced in `tools/generate_stubs.py` to a
single shared definition and apply it uniformly to every `Scalar`-typed
parameter, so the committed stubs describe the real Python API. The alias should
hold exactly the implicit-conversion set (`complex`, `typing.SupportsFloat`,
`typing.SupportsInt`, and the NumPy scalar types) — not widened beyond it.

Validation should confirm `mypy`/`pyright` accept plain/NumPy scalar arguments
and that `mypy.stubtest` stays clean.

## Notes

pybind11 cannot emit a Python union for a `Scalar` parameter from C++ (stubgen
only sees the C++ type, not the conversion set), so this lives in the stub
generator. Tracking the unification separately from #915 to keep that PR scoped.

IvanaGyro:
Still need to investigate if there is a method to not live in the stub generator.

Issue #1066, opened 2026-07-16
Storage.from_pylist/UniTensor.get_block reject mixed Python-int/numpy-scalar sequences (and can't preserve the old empty-list default) under the per-dtype overload model


IvanaGyro:
## Problem

PR #1053's keep-set consolidation replaced `Storage.from_pylist`'s and `UniTensor.get_block`/`get_block_`'s per-C++-dtype raw overloads with one `.def()` overload per Python-distinguishable numpy scalar width, plus a `py::int_`-typed overload for plain Python ints. Each of these is a `std::vector<T>` binding, and pybind11's vector caster requires **every element** of the list to satisfy the *same* element type caster — there is no way to register an overload that accepts a **heterogeneous** list mixing plain Python `int` elements with numpy scalar elements (e.g. `[np.int32(2), 3]`).

Before this PR, this worked by accident: the raw arithmetic-typed overloads (e.g. `vector<cytnx_int64>`) used pybind11's arithmetic integral caster, which accepts *any* object satisfying `__index__` even in the no-convert pass — both a plain Python `int` and a numpy scalar independently satisfy that per-element check, so a mixed list matched fine. After the consolidation, no single overload's element type accepts both, so the list falls through to the real-valued (`double`) overload or raises `TypeError`.

Confirmed via `chatgpt-codex-connector`'s review on PR #1053:

```python
>>> cytnx.Storage.from_pylist([np.int32(2), 3]).dtype()
Type.Double   # silently wrong; was some integer Type.* before PR #1053

>>> B.get_block([np.int32(0), 0], False)
TypeError: get_block(): incompatible function arguments...
```
(`get_block` regression applies to all 4 call sites: labeled/unlabeled × `get_block`/`get_block_`.)

## Related: the empty-list default (already discussed on PR #1053)

The same per-dtype-overload model also can't preserve `Storage.from_pylist([])`'s old `ComplexDouble` default (it's now `ComplexFloat`) without breaking `mypy.stubtest` — moving `complex128` to be registered first restores the runtime default but makes mypy see an earlier "broadest" stub signature shadowing every later, narrower one (14 fresh `overload-cannot-match` errors), which is exactly the class of bug PR #1053 exists to fix. See [this PR #1053 review thread](https://github.com/Cytnx-dev/Cytnx/pull/1053#discussion_r3593175533) for the detailed writeup.

[...]

Issue #1065, opened 2026-07-16
Keep-set numpy_scalar overloads (PR #1053) don't cover np.int8/np.uint8, silently widening to a real dtype


IvanaGyro:
## Problem

PR #1053's keep-set consolidation registers a `numpy_scalar<T>` overload per Python-distinguishable numpy integer width (`int64`/`uint64`/`int32`/`uint32`/`int16`/`uint16`/`bool`), but cytnx has no `Int8`/`Uint8` dtype, so `np.int8`/`np.uint8` were never included in that set. Before that PR, these values were (accidentally) intercepted by a raw arithmetic-typed overload registered ahead of the `numpy_scalar` block — pybind11's arithmetic integral caster accepts any object implementing `__index__` even in the no-convert pass, and numpy int8/uint8 both do — so they landed on *some* integer dtype (wrong width, but still an integer). After the consolidation, they fall through past the `int_`/numpy_scalar candidates entirely and match the real-valued (`double`/`float`) overload instead, which silently produces a **floating-point** dtype from an **integer** input.

Confirmed for `Scalar`'s constructor specifically (fixed directly in PR #1053, commit 34b19c1 — TODO: update once merged):

```python
>>> cytnx.Scalar(np.int8(5)).dtype()
Type.Double   # was some integer Type.* before PR #1053's Scalar.__init__ consolidation
```

`chatgpt-codex-connector`'s review on PR #1053 flagged this, and noted "the same copied keep-set affects the other scalar-dispatch bindings" — confirmed by grep, the same `numpy_scalar<int16_t>`/`numpy_scalar<uint16_t>` keep-set pattern (with no int8/uint8 counterpart) appears at roughly:

- `pybind/tensor_py.cpp` — ~32 sites (Tensor's arithmetic operators, `__setitem__`, `fill`, etc.)
- `pybind/unitensor_py.cpp` — ~52 sites (UniTensor's arithmetic operators, `Helpclass.set_elem`, etc.)
- `pybind/storage_py.cpp` — ~6 sites (`fill`/`append`/`from_pylist`)
- `pybind/linop_py.cpp` — ~2 sites (`set_elem`)

Each is a candidate for the same fix, but fixing all of them was out of scope for PR #1053 (which was itself already a large, review-heavy consolidation) and is deferred here.

## Suggested fix

[...]

PR #1104, opened 2026-07-20
refactor(pybind)!: remove the cytnx.Scalar Python binding (#1045)

### LINKED ISSUE #1045: Remove the cytnx.Scalar Python binding (Python should derive type from dtype)
## Goal
Stop exposing `cytnx.Scalar` on the Python surface. `Scalar` is a C++ type-erasure helper; in Python it is redundant and awkward — there is no native float32, and a `cytnx.Scalar` object is worse ergonomics than a plain Python number whose precision the caller already knows from the tensor's `dtype`. Python code should derive the numeric type from `dtype`, and bindings should return/accept native Python numbers.

## Background
- Direction confirmed by @yingjerkao: the Python API will not expose `cytnx.Scalar`; Python derives the type from `dtype`.
- Precedent already in place: `Tensor.item()` / `UniTensor.item()` dispatch on `dtype` and return native Python types (not `Scalar`). PR #1000 changed `norm()` to return a native `float` in Python (C++ returns `Scalar`, the binding converts).

## Scope / checklist
- [ ] Remove `pybind/scalar_py.cpp` and its registration in `pybind/cytnx.cpp` (`void scalar_binding(py::module&);` decl ~line 47, `scalar_binding(m);` call ~line 119).
- [ ] Audit & migrate every Python-facing binding that **returns** `Scalar` to a native Python number (dtype-dispatched, like `item()`). ~55 `Scalar` references across `pybind/{linalg,tensor,unitensor}_py.cpp` to review.
- [ ] Audit & migrate every binding that **accepts** a `Scalar` argument to accept native Python numbers instead (pybind's numeric casters already handle int/float/complex).
- [ ] Migrate Python call sites / tests that construct `cytnx.Scalar(...)`, e.g. `pytests/binding_dtype_test.py:241,246`.
- [ ] Update `cytnx/*.py` continuation modules and any docs/examples referencing `cytnx.Scalar`.
- [ ] Regenerate the pybind11 `.pyi` stubs after removal.

## Notes
- C++ `Scalar` is unaffected — this is purely about the Python binding surface.
- Should land as its own PR, not folded into #1000.

Refs #1000.

yingjerkao:
## Problem
`cytnx.Scalar` is a C++ type-erasure helper that adds nothing on the Python surface — there is no native float32, and a `cytnx.Scalar` object is worse ergonomics than a plain Python number whose precision the caller already knows from the tensor's `dtype`. Direction confirmed by @yingjerkao in #1045.

## Fix
- Remove `pybind/scalar_py.cpp` and its declaration/registration in `pybind/cytnx.cpp` and the CMake source list.
- Drop the redundant `const Scalar&` overloads from every `Tensor`/`UniTensor` arithmetic/comparison/`__setitem__` keep-set. Native int/float/complex and numpy scalars are already handled by sibling overloads, and with `Scalar` unregistered nothing in Python can produce a `cytnx.Scalar` to match them — so removal is behaviour-preserving for all native inputs.
- Migrate `linalg` `Ger`/`Gemm`/`Gemm_`/`Lanczos_Exp` from `Scalar` coefficient args to native `double`/`complex128` overloads. `Ger` keeps a no-`a` form so the C++ `Void`-default (which keeps `a` out of output-dtype promotion, see `src/linalg/Ger.cpp`) is preserved exactly.
- Migrate pytests off `cytnx.Scalar(...)`; regenerate `linalg.pyi` and surgically drop `Scalar` from `__init__.pyi`.

**Scope note:** the rest of `__init__.pyi` is independently stale (it predates the operator keep-set collapse, #916/#928, and was never regenerated). A full regeneration produces a ~2600-line diff of unrelated churn, so this PR only removes the `Scalar` entries and leaves the broader stub reconciliation for a dedicated stub-hygiene pass.

C++ `Scalar` is unaffected — this is purely the Python binding surface.

## Testing
- `openblas-cpu` `pycytnx` build clean.
- `pytest pytests/` → **139 passed, 1 skipped**.
- Verified `cytnx.Scalar` is gone and native int/float/complex/numpy arithmetic (both directions), floordiv, and `Ger`/`Gemm` real+complex coefficients still work.

Closes #1045. Refs #1000.


pybind/linalg_py.cpp
```diff
@@ -521,57 +897,64 @@ void linalg_binding(py::module &m) {
...
+    },
+    py::arg("a"), py::arg("x"), py::arg("y"), py::arg("b"), py::arg("c"),
+    py::call_guard<py::gil_scoped_release>());
+  m_linalg.def(
+    "Gemm_",
+    [](cytnx_complex128 a, const Tensor &x, const Tensor &y, cytnx_complex128 b, Tensor &c) {
+      cytnx::linalg::Gemm_(a, x, y, b, c);
+    },
+    py::arg("a"), py::arg("x"), py::arg("y"), py::arg("b"), py::arg("c"),
+    py::call_guard<py::gil_scoped_release>());
```
IvanaGyro: Need to wrap cytnx_complex128 with https://github.com/Cytnx-dev/Cytnx/blob/014a386047d52dce0f0831186303f8b2950984fb/pybind/complex_arg.hpp#L27-L32 to support python 3.10.

yingjerkao:
Addressed @IvanaGyro's review in 5893f30c: the complex overloads of `Gemm`/`Ger`/`Gemm_`/`Lanczos_Exp` now take `complex_arg.hpp`'s `ComplexArg` instead of a raw `cytnx_complex128`, so the generated stub parameter annotation includes builtin `complex` (fixing the Python 3.10 type-check rejection of e.g. `Gemm(2+1j, x, y)`), matching the `ExpH`/`ExpM` bindings in the same file. Runtime behaviour is unchanged — verified all four accept plain Python complex literals; `linalg.pyi` regenerated for the four signatures.

Heads-up: this branch is ~117 commits behind `master` now (it predates the `complex_arg.hpp` infrastructure landing), so it'll need a rebase/merge before it lands — I kept this fix focused rather than folding in a large merge.

yingjerkao:
Rebased onto current `master` (`2e96b248`). The branch was ~117 commits behind, so instead of a literal 3-way rebase (which silently dropped some of the removals — `master` had moved the accept-`Scalar` overloads around) I **re-derived the change against master's current code** and **squashed the two prior commits into one** (the earlier `ComplexArg` follow-up is folded in). Result: the branch is now exactly `master` + one clean commit.

What changed vs. the previous version of this PR, because `master` moved underneath it:

- **Overload removals re-verified against master.** Each removed accept-`Scalar` arithmetic/compare overload was confirmed to still have a native `double`/`complex128` sibling (plus the `numpy_scalar`/`py::int_` keep-set), so plain-Python-number dispatch is unchanged. In particular master's UniTensor scalar `//` and `%` keep-sets retain their native siblings, so `ut // 2.0` still works (#1049). Tensor `//` stays deliberately unbound (#941).
- **linalg migration matches master's `ExpH`/`ExpM` pattern:** `Ger`/`Gemm`/`Gemm_`/`Lanczos_Exp` now take a native real overload first, then a `complex_arg.hpp` `ComplexArg` fallback (accepts builtin `complex` on 3.10, keeps real→real dtype).
- **New master-era Python-Scalar tests removed.** `master` grew `pytests/scalar_variant_test.py` (+ Scalar-constructor cases in `keep_set_dispatch_test.py`) via the #847 variant refactor — all exercising the now-removed Python `cytnx.Scalar` surface. These are deleted; the underlying variant behaviour stays covered by C++ `tests/Scalar_test.cpp`.
- **Stubs regenerated** (pybind11-stubgen): `class Scalar` gone from `__init__.pyi`; `linalg.pyi` gains the real/complex overloads.

Testing on this box (openblas-cpu): `pytest pytests/` → **260 passed, 1 skipped**; smoke-checked that `cytnx.Scalar` is gone and Tensor/UniTensor scalar arithmetic + `Ger`/`Gemm`/`Gemm_`/`Lanczos_Exp` accept native int/float/complex with the expected result dtypes. `clang-format` v14 clean. (No GPU/C++-ctest impact: the change is confined to the `pycytnx` pybind target, not `libcytnx`.)

pybind/linalg_py.cpp
```diff
@@ -561,17 +561,55 @@ void linalg_binding(py::module &m) {
...
                py::call_guard<py::gil_scoped_release>());
 
+  // Ger's scaling coefficient `a` was formerly a cytnx.Scalar; the Python
+  // surface no longer exposes Scalar, so split it into native real/complex
+  // overloads plus a no-`a` form. The no-`a` form calls the C++ default
+  // (Scalar()), whose Void dtype keeps `a` out of the output-dtype promotion
+  // (see src/linalg/Ger.cpp) -- passing 1.0 explicitly would not reproduce it.
+  m_linalg.def(
+    "Ger", [](const Tensor &x, const Tensor &y) { return cytnx::linalg::Ger(x, y); }, py::arg("x"),
+    py::arg("y"));
```
IvanaGyro: BTW, this is not really the BLAS `ger`.

IvanaGyro (approved):

pybind/linalg_py.cpp
```diff
@@ -561,17 +561,55 @@ void linalg_binding(py::module &m) {
...
-               py::arg("b"), py::arg("c"), py::call_guard<py::gil_scoped_release>());
+  // Gemm_ (in-place c = a*(x@y) + b*c): native real/complex overloads for the
+  // scaling coefficients a and b, replacing the former cytnx.Scalar arguments.
+  // A mixed real/complex (a, b) call resolves to the complex overload; that
+  // only widens the real coefficient's internal dtype and does not change the
+  // result, since the complex coefficient already forces a complex output.
+  m_linalg.def(
+    "Gemm_",
+    [](cytnx_double a, const Tensor &x, const Tensor &y, cytnx_double b, Tensor &c) {
+      cytnx::linalg::Gemm_(a, x, y, b, c);
```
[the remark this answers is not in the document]
yingjerkao: Verified and fixed in 62881865. You're right — reproduced against a from-scratch build of `origin/master`: `Gemm(5, float, float)` was `Float`, and my earlier `cytnx_double` overload made it `Double` (on master an integer coefficient resolved to an `int64`/`uint64` Scalar, which doesn't upcast a Float result).

Rather than a `py::int_` overload, the coefficients now go through a single **`ScalarArg`** dispatcher (`pybind/scalar_arg.hpp`) that infers the coefficient's cytnx dtype from the caller's actual Python/numpy type at runtime — the single-dispatcher model @IvanaGyro proposes in #1066, not a per-dtype overload ladder. Result, all matching master again: `Gemm(5,f,f)` / `Gemm(np.int64(5),…)` / `Gemm(np.int32(5),…)` / `Ger(f,f,5)` → `Float`; `2.0` → `Double`; `2+1j` → `ComplexDouble`. It also closes the `np.int8`/`np.uint8` gap (#1065) and now *preserves* narrow numpy floats/complex that the old accept-`Scalar` overload rejected outright.

pybind/linalg_py.cpp
```diff
@@ -905,17 +905,56 @@ void linalg_binding(py::module &m) {
...
+  // surface no longer exposes Scalar, so split it into native real/complex
+  // overloads plus a no-`a` form. The no-`a` form calls the C++ default
+  // (Scalar()), whose Void dtype keeps `a` out of the output-dtype promotion
+  // (see src/linalg/Ger.cpp) -- passing 1.0 explicitly would not reproduce it.
+  m_linalg.def(
+    "Ger", [](const Tensor &x, const Tensor &y) { return cytnx::linalg::Ger(x, y); }, py::arg("x"),
+    py::arg("y"));
+  m_linalg.def(
+    "Ger",
+    [](const Tensor &x, const Tensor &y, cytnx_double a) { return cytnx::linalg::Ger(x, y, a); },
```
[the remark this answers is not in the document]
yingjerkao: Fixed in 62881865, together with the integer case. The coefficients now go through a single runtime `ScalarArg` dispatcher (`pybind/scalar_arg.hpp`) that keeps the caller's numpy width: `np.float32` → `Float`, `np.complex64` → `ComplexFloat`, and `np.int8`/`np.uint8` → `Int16`/`Uint16` (the widths pybind's `numpy_scalar<T>` set doesn't register, #1065). One overload → one stub signature, per @IvanaGyro's #1066.

One correction to the premise: on `origin/master` a bare `np.float32`/`np.complex64` coefficient actually raised `TypeError` here (the accept-`Scalar` overload's implicit conversions didn't accept them), rather than preserving Float32/Complex64 — so this is now accepted *and* preserved, not a regression that existed before.

pybind/linalg_py.cpp
yingjerkao: Done. The coefficients are now taken as `ScalarArg` (`pybind/scalar_arg.hpp`), whose caster handles a builtin Python `complex` directly (`PyComplex_Check`) and whose stub annotation includes `complex` (the same corrected `io_name` as `complex_arg.hpp`), so `Gemm(2+1j, …)` / `Lanczos_Exp(…, 2+1j)` type-check and run under the 3.10 floor. `complex_arg.hpp` is still used by `ExpH`/`ExpM`.

pybind/linalg_py.cpp
```diff
@@ -905,17 +905,56 @@ void linalg_binding(py::module &m) {
...
                py::call_guard<py::gil_scoped_release>());
 
+  // Ger's scaling coefficient `a` was formerly a cytnx.Scalar; the Python
+  // surface no longer exposes Scalar, so split it into native real/complex
+  // overloads plus a no-`a` form. The no-`a` form calls the C++ default
+  // (Scalar()), whose Void dtype keeps `a` out of the output-dtype promotion
+  // (see src/linalg/Ger.cpp) -- passing 1.0 explicitly would not reproduce it.
+  m_linalg.def(
+    "Ger", [](const Tensor &x, const Tensor &y) { return cytnx::linalg::Ger(x, y); }, py::arg("x"),
+    py::arg("y"));
```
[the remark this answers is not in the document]
yingjerkao: For context: `Ger`'s binding didn't release the GIL on `master` either, so this isn't introduced by the Scalar removal — I kept the existing behavior to keep the PR scoped. `Ger` is pure C++ compute (no Python callbacks), so adding `py::call_guard<py::gil_scoped_release>()` to both `Ger` overloads is safe; I'd prefer to land it as a small separate consistency change (a couple of other linalg bindings are in the same boat) rather than widen this PR. `Gemm`/`Gemm_`/`Lanczos_Exp` already release it.

pybind/linalg_py.cpp
yingjerkao: Agreed — the binding just forwards to `cytnx::linalg::Ger`; the name predates this PR and its semantics are unchanged here. Happy to track a rename separately if it's a misnomer worth fixing.

yingjerkao:
Pushed `62881865` addressing the review. **Heads-up since the approval was on the previous commit (`2e96b248`): this commit changes coefficient-dtype behavior**, so it's worth a re-glance.

Codex correctly caught that binding the `Ger`/`Gemm`/`Gemm_`/`Lanczos_Exp` scaling coefficients as `cytnx_double` + `ComplexArg` widened an **integer** coefficient to `Double` (verified against a from-scratch `origin/master` build: `Gemm(5, float, float)` was `Float`, became `Double`). Fixed by taking each coefficient as a single **`ScalarArg`** runtime dispatcher (`pybind/scalar_arg.hpp`) instead of a per-dtype overload — the model @IvanaGyro describes in #1066. It restores the exact old integer/float/complex dtypes, additionally fixes the `np.int8`/`np.uint8` gap (#1065), and preserves narrow numpy float/complex coefficients (which the old accept-`Scalar` overload rejected with `TypeError`). One stub signature per function; `pytest pytests/` → 260 passed, 1 skipped. Per-thread replies inline.

pybind/scalar_arg.hpp
```diff
@@ -0,0 +1,168 @@
...
+        return pybind11::handle();
+      }
+    }();
+    return np_generic.ptr() != nullptr && pybind11::isinstance(src, np_generic);
+  }
+
+  // Build a cytnx::Scalar from an arbitrary Python number (Python
+  // bool/int/float/complex, or any numpy scalar), inferring the dtype at
+  // runtime. Returns false for anything that is not a number.
+  inline bool load_scalar_arg(pybind11::handle src, cytnx::Scalar &out) {
```
IvanaGyro: I feel that this dispatcher is slow. Will explicitly overload for each numpy type be faster?

pybind/linalg_py.cpp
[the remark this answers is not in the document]
IvanaGyro: #1066 is for list args only. There should be more efficient way for scalar args.

pybind/scalar_arg.hpp
yingjerkao: You're right — reverted. I've dropped `scalar_arg.hpp` and put the branch back to `2e96b248` (the commit you approved), so `Ger`/`Gemm`/`Gemm_`/`Lanczos_Exp` are the simple `double` + `ComplexArg` binding again. Agreed the runtime numpy reflection isn't the right tool for a single scalar coefficient.

pybind/linalg_py.cpp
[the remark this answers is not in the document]
yingjerkao: Thanks for the correction — I over-applied #1066 (which, as you say, is about **list** args and their mixed-element / empty-list problems) to scalar args, where those problems don't exist. Reverted to the minimal `double` + `ComplexArg` binding in `2e96b248`.

The only thing left is Codex's note that a bare integer coefficient now widens a `Float` result to `Double` (correctness-neutral, and only for the rare Float-tensor + integer-coefficient case). If you do want to preserve the coefficient's dtype, I'll do it with explicit `numpy_scalar<T>` overloads (the `bind_exp_scalar` pattern) as a separate, consistent change rather than a dispatcher — just let me know if it's worth it.

pybind/scalar_arg.hpp
IvanaGyro: > You're right — reverted. I've dropped `scalar_arg.hpp` and put the branch back to `2e96b248` (the commit you approved), so `Ger`/`Gemm`/`Gemm_`/`Lanczos_Exp` are the simple `double` + `ComplexArg` binding again. Agreed the runtime numpy reflection isn't the right tool for a single scalar coefficient.

Sorry. It seems that I was wrong.

IvanaGyro: I investigated pybind11 3.0.4's scalar casters and overload resolution, and benchmarked alternatives with the same shape as `Ger(x, y, a)` (two bound arguments followed by the scalar).

**Conclusion:** the concern about the original dispatcher is valid, but explicit external pybind11 overloads for every NumPy scalar type are **not generally faster**. The dispatcher can be made substantially cheaper without an overload ladder. More importantly, the current reverted head (`2e96b248`) has a correctness problem and should not be merged unchanged.

### Correctness problem in the current `double` + `ComplexArg` bindings

The current binding registers the `double` overload before `ComplexArg` for `Ger`, `Gemm`, `Gemm_`, and `Lanczos_Exp`. For `np.complex64(3 + 4j)`:

1. Neither overload matches during pybind11's no-conversion pass.
2. During the conversion pass, the earlier `double` overload invokes `PyFloat_AsDouble`.
3. NumPy converts the value to `3.0`, emits `ComplexWarning`, and discards `4j`.
4. The later `ComplexArg` overload is never tried.

I reproduced this with pybind11 3.0.4. It is especially risky for `Lanczos_Exp`, where a complex time step can silently become real. The same binding also maps `np.float32`/integer coefficients to `Double` and `np.complex64` to `ComplexDouble`; because `Ger`/`Gemm` include coefficients in dtype promotion, that can change the result dtype and select double-precision storage/kernels.

This behavior follows from pybind11's [floating-point caster](https://github.com/pybind/pybind11/blob/v3.0.4/include/pybind11/cast.h#L231-L315), [complex caster](https://github.com/pybind/pybind11/blob/v3.0.4/include/pybind11/complex.h#L39-L70), and registration-order-based [overload resolution](https://pybind11.readthedocs.io/en/stable/advanced/functions.html#overload-resolution-order).

### Why a per-dtype overload ladder is not the efficient solution

Each `py::numpy_scalar<T>` caster is efficient by itself: it checks a cached NumPy scalar type and extracts with `PyArray_ScalarAsCtype` ([implementation](https://github.com/pybind/pybind11/blob/v3.0.4/include/pybind11/numpy.h#L778-L801)). The cost comes from pybind11's outer overload dispatcher:

- overloads are attempted sequentially in registration order, potentially in both no-conversion and conversion passes;
- for `Ger(x, y, a)`, every failed scalar overload also reloads the preceding Tensor arguments;
- later scalar types become progressively slower;
- `Gemm_` has two independently typed coefficients, so complete explicit coverage becomes combinatorial.

A representative pybind11 3.0.4 microbenchmark gave:

| Strategy | Approximate total call time |
|---|---:|
| Baseline bound call | 0.10 µs |
| Original `dtype.kind`/`itemsize` reflection | 0.12 µs for Python numbers; 0.40–0.46 µs for NumPy scalars |
| One caster backed by exact `std::variant` alternatives | 0.15–0.31 µs |
| One caster using cached NumPy type-pointer dispatch | 0.11–0.15 µs |
| External overload ladder | 0.21 µs for the first type; 0.8–2.2 µs for later types |

The absolute values are machine-dependent, but the ordering was stable: an overload ladder only favors whichever type is registered first. The original reflection overhead is only a few tenths of a microsecond and is normally negligible relative to allocating and computing a GER/GEMM result; incorrect promotion to a wider kernel is likely more expensive.

### Recommended implementation

Restore a single input-only `ScalarArg`, but remove Python-level `dtype.kind`/`itemsize` reflection. Implement its `load()` using either:

1. a `std::variant` of exact `py::numpy_scalar<T>` casters followed by `py::bool_`, `py::int_`, `double`, and `ComplexArg` (simpler and already fast); or
2. cached exact NumPy scalar type-pointer checks with direct extraction (fastest in the microbenchmark).

The caster should immediately convert the selected value to `cytnx::Scalar` while the GIL is held; the computational lambda then receives only C++ data, so the existing GIL-release guards remain safe. Put exact NumPy alternatives before broad Python conversions, with Boolean before integer.

This keeps one binding and one union stub signature, preserves `np.float32`/`np.complex64`, handles `np.int8`/`np.uint8` by widening them to Cytnx `Int16`/`Uint16`, avoids repeated Tensor conversions, and handles the two `Gemm_` coefficients independently.

So my recommendation is:

1. Do fix the correctness/dtype issue; do not optimize it with many pybind11 overloads.
2. Use one exact-type `ScalarArg` caster, preferably variant-backed unless profiling justifies cached-pointer code.
3. Add regression tests for nonzero-imaginary `np.complex64`, narrow result dtypes, native integers, and mixed `Gemm_` coefficient types.
4. Treat releasing the GIL for `Ger` as a separate, more meaningful performance improvement.

— **GPT-5.6 Sol Max**

IvanaGyro: A correction to my earlier recommendation: **yes, for this PR I would prefer the cached NumPy type-pointer caster over the variant-backed caster.**

I originally favored the variant because it composes pybind11 casters and avoids depending directly on `pybind11::detail`. That was too conservative here: Cytnx pins `pybind11 == 3.0.4` exactly, the cached implementation is small and isolated, and a future pybind11 bump will compile-test it. The variant remains reasonable, but it is not the best-performing choice shown by the benchmark.

This is worth implementing primarily to restore correct coefficient dtype/complex-value behavior; the dispatch-speed improvement is secondary. I compiled and tested the following core against pybind11 3.0.4 + NumPy 2.3.5 for all 13 standard NumPy scalar types, including `np.complex64(3+4j)`:

```cpp
#include <pybind11/numpy.h>

template <typename NumpyT, typename CytnxT = NumpyT>
bool try_numpy_scalar_exact(pybind11::handle src, cytnx::Scalar &out) {
  // pybind11's numpy_scalar<T> caster already owns a lazily cached type pointer.
  using caster = pybind11::detail::type_caster<pybind11::numpy_scalar<NumpyT>>;
  auto *expected = reinterpret_cast<PyTypeObject *>(caster::target_type().ptr());
  if (Py_TYPE(src.ptr()) != expected) return false;

  NumpyT value{};
  pybind11::detail::npy_api::get().PyArray_ScalarAsCtype_(src.ptr(), &value);
  out = cytnx::Scalar(static_cast<CytnxT>(value));
  return true;
}

inline bool load_standard_numpy_scalar(pybind11::handle src, cytnx::Scalar &out) {
  using namespace cytnx;
  // Put the common coefficient types first; each failed check is one pointer comparison.
  return try_numpy_scalar_exact<float, cytnx_float>(src, out)
         || try_numpy_scalar_exact<double, cytnx_double>(src, out)
         || try_numpy_scalar_exact<std::complex<float>, cytnx_complex64>(src, out)
         || try_numpy_scalar_exact<std::complex<double>, cytnx_complex128>(src, out)
         || try_numpy_scalar_exact<bool, cytnx_bool>(src, out)
         || try_numpy_scalar_exact<std::int8_t, cytnx_int16>(src, out)
         || try_numpy_scalar_exact<std::uint8_t, cytnx_uint16>(src, out)
         || try_numpy_scalar_exact<std::int16_t, cytnx_int16>(src, out)
         || try_numpy_scalar_exact<std::uint16_t, cytnx_uint16>(src, out)
         || try_numpy_scalar_exact<std::int32_t, cytnx_int32>(src, out)
         || try_numpy_scalar_exact<std::uint32_t, cytnx_uint32>(src, out)
         || try_numpy_scalar_exact<std::int64_t, cytnx_int64>(src, out)
         || try_numpy_scalar_exact<std::uint64_t, cytnx_uint64>(src, out);
}
```

`load_scalar_arg()` should call this before broad `PyFloat_Check`/`PyComplex_Check` branches, because under NumPy 2.x `np.float64` and `np.complex128` are subclasses of Python `float` and `complex`. Keep the existing Python `bool`/`int`/`float`/`complex` handling for native values.

The exact `Py_TYPE` comparison is the hot path. If NumPy scalar subclasses must also be accepted, add a rare fallback pass using `PyObject_TypeCheck(src.ptr(), expected)` after the exact/native paths. That preserves subclass support without slowing standard scalar calls.

The deliberate tradeoff is the two `pybind11::detail` references. They mirror pybind11 3.0.4's own [`numpy_scalar<T>` caster](https://github.com/pybind/pybind11/blob/v3.0.4/include/pybind11/numpy.h#L778-L801), require no NumPy build-time headers, and are reasonable here because pybind11 is exactly pinned. This dependency should be noted beside the helper and revalidated whenever the pin changes.

— **GPT-5.6 Sol Max**

PR #1101, opened 2026-07-20
refactor(linalg): typed GPU dispatch for Outer + retire Kron GPU promotion trait (#1003)

yingjerkao:
## Problem

The last of #1003: `Outer` and `Kron` were the two remaining GPU ops still on legacy dispatch.

- **Outer** dispatched through a generated **12×12 `cuOuter_ii` dtype-pair table** (144 wrapper functions + a 2D lookup) — exactly the kind of "old generated dtype table" #1003 wants gone.
- **Kron** was the **last user** of the bespoke GPU promotion trait hierarchy (`type_promote_gpu_t` / `type_promote_from_gpu_pointer` / `_t`) in `Type.hpp`.

## Fix

Two commits:

1. **Outer → typed dispatch.** Replace `cuOuter_ii` with a single `std::variant`-based `cuOuter_dispatch` (the elementwise `cu*_dispatch` design). `Outer.cpp` already promotes both operands to the output dtype, so `out`/`Lin`/`Rin` share `out->dtype()`; the dispatch visits that one type via `as_storage_variant` and routes **exactly as the old diagonal table entries did** — floating/complex → cuBLAS **GER** (`A = 1·x·yᵀ`), integer/bool → the custom `cuOuter_kernel`. Deletes the 144 wrappers + the table + its population. **Net ≈ −1600 lines.**

2. **Kron → retire the trait.** Cast both operands to the promoted output dtype up front (a no-op ref when it already matches) and dispatch on the single output type via `out.gpu_ptr()`. This is numerically identical to the kernel's per-element `TO(Lin)*TO(Rin)` cast, cuts kernel instantiations from 11×11 → 11, and lets the trait hierarchy be **deleted from `Type.hpp`** (no remaining users — only historical comments mention it).

Both changes are **GPU-only and behaviour-preserving** — they change *how* a kernel is selected, not the math. The CPU `Outer_ii`/`type_promote_from_pointer_t` paths are untouched (out of #1003's CUDA scope).

> ⚠️ Per the repo guardrails, flagging this as a numerical/dtype-adjacent change for human review — though it is a pure dispatch refactor that reuses the existing kernels unchanged.

## Testing

On an RTX 4070 Ti SUPER (CUDA 13.0, `sm_89`):

- **GPU Kron/Outer suite: 9/9 pass** — including the mixed `ComplexFloat × Double → ComplexDouble` promotion discriminator (output **dtype and value**) and GPU-vs-CPU cross-checks over every dtype (from #1100).
[...]

PR #1105, opened 2026-07-20
refactor(linalg): implement Outer as Kron+reshape, retire Outer dispatch (#1003)

yingjerkao:
## Problem
`Outer(a, b)` for rank-1 vectors was implemented with its own per-dtype dispatch tables (`Outer_ii` on CPU, `cuOuter_ii` on GPU). Because `Outer` casts both operands to the promoted dtype **before** dispatching, only the diagonal `Outer_ii[t][t]` entries were ever reachable — the ~30 off-diagonal entries were dead code, and three diagonal rows (`Int16`, `Uint16`, `Bool`) were never registered, so `Outer` on those dtypes dereferenced a `nullptr` and segfaulted (#1099).

## Fix
Per @ianmccul's review of #1099/#1101: `Outer(a, b)` is exactly the Kronecker product of the two vectors reshaped to `{m, n}` — `Kron(a, b)` is a length-`m*n` vector whose element `(i*n + j)` is `a_i * b_j`. Reimplement `Outer` as `Kron(a, b).reshape({m, n})` and delete the `Outer_ii` / `cuOuter_ii` tables and their internal kernels.

`Kron` already uses variant-based typed dispatch (`type_promote_from_pointer_t` / `type_promote_from_gpu_pointer_t`) covering **every** dtype pair on both CPU and GPU, and shares the same `Type.type_promote` result dtype — so this both fixes the #1099 segfault and removes the dead dispatch machinery. `cuKron` compiles under CUDA 13 (#999, closed), so the GPU path is covered.

`type_promote_gpu_t` is intentionally **left in place**: it still underpins the `type_promote_from_gpu_pointer_t` trait that `cuKron` uses, so it is not dead (this is why the Type_test.cpp break flagged on #1101 does not apply here).

### Removed
- `src/backend/linalg_internal_cpu/Outer_internal.{cpp,hpp}`
- `src/backend/linalg_internal_gpu/cuOuter_internal.{cu,hpp}`
- `Outerfunc_oii` typedef, `Outer_ii` / `cuOuter_ii` members + their population in `linalg_internal_interface.{hpp,cpp}`, and the CMake source entries.

Net: **+216 / −3369**.

## Testing
Added `tests/linalg_test/Outer_test.cpp` and `tests/gpu/linalg_test/Outer_test.cpp` with independent hand-computed values: rectangular shape, single-element rank-1 operands, `ComplexFloat × Double → ComplexDouble` promotion, and the `Int16`/`Uint16`/`Bool` diagonal cases that segfaulted (#1099).
- **CPU** (`debug-openblas-cpu`, ASan): all new `Outer.*` tests + existing `DtypePromotion.OuterComplexfloatDouble` and `Zero_extent` Outer cases pass.
- **GPU** (`build_gpu` `gpu_test_main`, CUDA 13, RTX 4070 Ti): all 3 `Outer.*` GPU tests pass.

Refs #1003. **Supersedes #1099 and #1101** (the missing dispatch rows no longer exist; there is no `cuOuter` to add typed dispatch for).


ianmccul (approved):
Looks good

PR #1113, opened 2026-07-22
ci(windows): build and publish PyPI wheels

IvanaGyro:
## Stack

`master → #1110 → codex/windows-msvc-cuda-stack (#1116 patch) → #1117 → #1118 → #1119 → #1113`

This workflow-only PR depends on #1119. Its own diff is limited to the two existing PyPI workflow files.

## Summary

- add Windows Server 2022 CPU and CUDA wheel jobs using the locked `wheel-cpu` and `wheel-cuda` Pixi environments;
- run `prepare-windows` before CPU wheels and `prepare-cuda` before CUDA wheels, creating ARPACK/CUDA import libraries before CMake configures;
- build CPython 3.10 through 3.14 `win_amd64` wheels;
- cache compilation output and upload Windows artifacts separately;
- include Windows artifacts in stable PyPI publication while leaving nightly/Conda behavior unchanged;
- retain the existing Linux and macOS jobs.

## Prior failure and fix

The split Windows CUDA job reached native linking but failed in device LTO against the PyPI CUDA/NVVM layout. #1119 now disables IPO only for Windows CUDA release wheels while preserving Linux LTO. The earlier Windows `MODULE` install issue is fixed in #1118.

## Validation

- both workflow files pass repository-pinned hooks and `git diff --check`; 
- the top-layer diff contains only the two workflow files;
- local locked CPU/CUDA environments, all-target MSVC builds, wheel tooling, downstream package consumption, and release-manifest tests pass in the implementation layers;
- fresh GitHub checks were triggered by the rebase onto current `master`.

Part of #1114.

### DIFF
--- a/.github/workflows/release_pypi.yml
+++ b/.github/workflows/release_pypi.yml
@@ -60,6 +60,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
+permissions:
+  contents: read
+
 jobs:
   DetermineChannel:
     # Resolves a single source of truth for the rest of the workflow:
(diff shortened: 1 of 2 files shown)


yingjerkao:
I thought we don't support native Windows?

IvanaGyro:
> I thought we don't support native Windows?

It is just a proposal. If we really exclude supporting Windows, the PR chain can be close. Only partially #1109 and partially #1116 are needed.

yingjerkao:
> > I thought we don't support native Windows?
> 
> It is just a proposal. If we really exclude supporting Windows, the PR chain can be close. Only partially #1109 and partially #1116 are needed

I will approve this PR chain as a whole.

Issue #1129, opened 2026-07-28
[build] FindCUTENSOR/FindCUQUANTUM: exported CUTENSOR_ROOT/CUQUANTUM_ROOT silently overrides an explicit -D


yingjerkao:
## Summary

`FindCUTENSOR.cmake` and `FindCUQUANTUM.cmake` prefer the **environment** variable over an explicit
`-DCUTENSOR_ROOT=` / `-DCUQUANTUM_ROOT=` on the command line. An explicit `-D` is silently
discarded whenever the corresponding variable happens to be exported — which conda/miniforge
environments do by default (`CUQUANTUM_ROOT=$CONDA_PREFIX`).

This is the reverse of the CMake convention (cache variable wins, environment is a fallback
default), and it is silent: the cache still holds the value the user asked for, so
`CMakeCache.txt` and `cmake -L` both report the *requested* path while the build actually used a
different one.

## Reproducer

```console
$ cmake -S . -B b -DCMAKE_MODULE_PATH=<cytnx>/cmake/Modules -DCUQUANTUM_ROOT=/tmp/decoy_root
  # with CUQUANTUM_ROOT=/home/me/miniforge3 exported in the environment

-- Looking for cuQuantum in /home/me/miniforge3      <-- environment won
-- REPRO effective CUQUANTUM_ROOT = /home/me/miniforge3
-- REPRO cache     CUQUANTUM_ROOT = /tmp/decoy_root  <-- what the user asked for
```

The cache and the effective value disagree, with no warning.

## Cause

`cmake/Modules/FindCUQUANTUM.cmake` (and the identical block in `FindCUTENSOR.cmake`):

```cmake
if(NOT DEFINED ENV{CUQUANTUM_ROOT} AND NOT DEFINED CUQUANTUM_ROOT)
  message(FATAL_ERROR "CUQUANTUM_ROOT not set!")
else()
  if(DEFINED ENV{CUQUANTUM_ROOT})
    set(CUQUANTUM_ROOT "$ENV{CUQUANTUM_ROOT}")   # <-- unconditional; shadows the cache entry
  endif()
```

The `set()` creates a normal variable that shadows the cache variable for the rest of the module,
so the `-D` value is never consulted again.

Note this *undoes* what `CMakeLists.txt` already does correctly. `require_dependent_variable`
(`CMakeLists.txt:135`) seeds the cache from the environment only as a default:

```cmake
if(NOT DEFINED ${variable})
[...]

Issue #970, opened 2026-07-05
Deployment strategy — status board


pcchen:
# Deployment strategy — status board

Umbrella tracker grouping the existing build/packaging issues and PRs into a
phased deployment roadmap. This references existing issues/PRs; it does not
duplicate or re-litigate them. Checkboxes track whether each line's
underlying issue/PR is still open, not a separate approval step.

## Phase gate — packaging model (Phases 2–3 depend on this)
- [ ] Decide PyTorch-style single-package vs JAX-style cpu/cuda12/cuda13 variants — #969 (+ #966)

## Phase 0 — wheel size / PyPI quota
- [x] Exclude C++ SDK (`libcytnx.a`, headers, CMake config) from the Python wheel — #947 (PR #967)
- [x] Dedup OpenBLAS across manylinux arpack+cytnx wheel builds — #947 (PR #967)

Both are implemented in **PR #967**, pending merge. The dedup path was built and
inspected on the real `manylinux_2_28` image (aarch64): the wheel vendors a single
OpenBLAS and excludes the SDK. Note (per #971): the dedup is a size optimization,
not a hard guarantee — two OpenBLAS variants are still permitted when arpack's
build lacks LAPACKE, so no single-OpenBLAS CI gate is enforced.

## Phase 1 — build-time discovery
**PR #950 is being split into focused PRs** (see the comment thread on #950); the
lines below map to those split PRs.
- [ ] Fix the finder `*_FOUND` dead-code bug in FindCUTENSOR/FindCUQUANTUM — #945/#946 (#950 split: finder-correctness PR)
- [ ] Version-aware cuTENSOR library-subdir search (drop dead `lib/10.2`,`lib/11`) — #946 (#950 split: finder-layout PR)
- [ ] Modernize CUDA configure diagnostics **and** remove the stale legacy `CUDA_*_LIBRARY`/`linkflags.tmp` dead block — #949 (#950 split: diagnostics PR; #950's current diff fixes only the display half)
- [ ] Enforce CUDA/cuTENSOR version floors (CUDA ≥ 12, cuTENSOR ≥ 2.0) — #962 (#950 split: version-policy PR, pending the #962 decision)
- [ ] Correct the stale/incomplete compiling-options table in the install docs (missing `USE_CUTENSOR`/`USE_CUQUANTUM`/`BACKEND_TORCH`, wrong default install prefix) — ← open gap (untouched by #950/#967)
[...]

PR #1130, opened 2026-07-28
docs(install): document cuTENSOR/cuQuantum as required for CUDA builds

yingjerkao:
## Problem

`docs/source/adv_install.rst` presented cuTENSOR and cuQuantum as optional extras and
never mentioned `CUTENSOR_ROOT` / `CUQUANTUM_ROOT` at all. There was **no path from the
documentation to a working CUDA build**: follow the install steps, run a documented CUDA
preset, and configure fails with

```
Error: Cache variable 'CUTENSOR_ROOT' is required under dependency conditions
(USE_CUTENSOR) but its value is empty.
```

I hit exactly this on a machine that already had CUDA installed.

In practice both libraries are mandatory. Every CUDA configuration the project ships turns
them on:

| configuration | cuTENSOR | cuQuantum |
| --- | --- | --- |
| `openblas-cuda` preset (and `mkl-cuda`, `debug-openblas-cuda`, `debug-mkl-cuda`, which inherit it) | ON | ON |
| GPU CI (`ci-gpu_tests.yml:80`) | ON | ON |
| CUDA release wheels (`prepare_cuda_release.py` → `--preset=openblas-cuda`) | ON | ON |

## Fix

- **State the requirement up front** — a CUDA build needs cuTENSOR and cuQuantum, not just
  the CUDA toolkit — with the two root variables, a worked conda example, and the expected
  `include/` + `lib/` layout (including that both the flat and per-CUDA-major `lib/`
  layouts are searched, post-#993).
- **New CUDA options table** covering `USE_CUTENSOR`, `USE_CUQUANTUM`, `CUTENSOR_ROOT`,
  `CUQUANTUM_ROOT`.
- **`BACKEND_TORCH`** added to the main options table.
- **Default install prefix corrected**: `~/.local/cytnx`, not `/usr/local/cytnx`
  (`CMakeLists.txt:166` and the `default` preset).
- **Copy-paste bug fixed**: the cuQuantum paragraph gave `-DUSE_CUTENSOR=ON` as its compile
  option, and misspelled "cuqauantum".
- **Note that presets override** the plain-`cmake` defaults the table lists.
- **Warning about #1129** — an exported `CUTENSOR_ROOT`/`CUQUANTUM_ROOT` silently takes
  precedence over `-D`, which bites conda users because conda exports `CUQUANTUM_ROOT`.

## Testing

Documentation only; no build or code changes.

RST validated with docutils 0.23 against the file before and after: the three grid tables
parse and render (verified in the generated HTML), and the file gains **no new parse
errors** — the two reported are pre-existing `code-block` directives that plain docutils
does not implement (Sphinx-only) and are present on `master` too.

Every factual claim was checked against the tree: option defaults from `CMakeLists.txt`
(L166, L168–186), preset values from `CMakePresets.json`, CI flags from
`ci-gpu_tests.yml:80`, wheel flags from `tools/prepare_cuda_release.py`.

Addresses the install-docs gap tracked on #970 (Phase 1).

### DIFF
--- a/docs/source/adv_install.rst
+++ b/docs/source/adv_install.rst
@@ -132,7 +132,7 @@ There are two methods how you can set-up all the dependencies before starting th
         3. Make sure libblas=mkl (you can check using *$conda list | grep libblas*)
 
 
-3. In addition, if you want to have GPU support (compile with -DUSE_CUDA=on), then additional packages need to be installed:
+3. In addition, if you want to have GPU support (compile with ``-DUSE_CUDA=ON``), then additional packages need to be installed:
 
 .. .. code-block:: shell
 
@@ -140,19 +140,33 @@ There are two methods how you can set-up all the dependencies before starting th
 
 .. code-block:: shell
 
-    $conda install -c nvidia cuda
+    $conda install -c nvidia cuda cutensor cuquantum
 
-If cutensor shall be used as well (compile option -DUSE_CUTENSOR=ON), then further install the following:
+.. note::
 
-.. code-block:: shell
+    **A CUDA build needs cuTENSOR and cuQuantum as well, not just the CUDA toolkit.**
+    Every CUDA configuration Cytnx ships enables all three: the ``openblas-cuda`` preset
+    (and ``mkl-cuda`` / ``debug-openblas-cuda`` / ``debug-mkl-cuda``, which inherit it)
+    sets ``USE_CUDA``, ``USE_CUTENSOR`` and ``USE_CUQUANTUM`` to ``ON``, and so do the
+    GPU CI job and the published ``cytnx-cuda`` wheels.
 
-    $conda install -c nvidia cutensor
+    ``-DUSE_CUTENSOR=ON`` and ``-DUSE_CUQUANTUM=ON`` each additionally **require** the
+    corresponding install root to be set, either in the environment or on the command
+    line. Configuring without them fails with::
 
-Similarly, cuqauantum (compile option -DUSE_CUTENSOR=ON), requires:
+        Error: Cache variable 'CUTENSOR_ROOT' is required under dependency conditions
+        (USE_CUTENSOR) but its value is empty.
 
-.. code-block:: shell
+    With the conda packages above, both roots are your conda prefix:
 
-    $conda install -c nvidia cuquantum
+    .. code-block:: shell
+
+        $export CUTENSOR_ROOT=$CONDA_PREFIX
+        $export CUQUANTUM_ROOT=$CONDA_PREFIX
+
+    Each root is expected to contain ``include/`` and ``lib/`` subdirectories. cuTENSOR
+    2.x is required; ``lib/`` may either hold the libraries directly or use a
+    per-CUDA-major subdirectory (``lib/12``, ``lib/13``) — both layouts are searched.
 
 **Option B. Install dependencies via system package manager**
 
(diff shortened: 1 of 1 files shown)


docs/source/adv_install.rst
```diff
@@ -241,6 +255,54 @@ The following are the available compiling option flags that you can specify in *
...
+    them — in particular the CUDA presets turn ``USE_CUTENSOR`` and ``USE_CUQUANTUM``
+    on. ``cmake -L <build dir>`` shows the values actually in effect.
+
+Additional options for CUDA if -DUSE_CUDA=on:
+
++-------------------------+-------------------+------------------------------------+
+|       options           | default           |          description               |
++-------------------------+-------------------+------------------------------------+
+| -DUSE_CUTENSOR          |  OFF              | Accelerate tensor contraction with |
+|                         |                   | cuTENSOR (>= 2.0). Requires        |
```
IvanaGyro, line 277: I feel we should not teach users to manipulate CMAKE flags manually because we don't guarantee that configuration will work. Maintaining so many configurations cost too much time and resources, and it is easy to miss some combinations.

docs/source/adv_install.rst
```diff
@@ -241,6 +255,54 @@ The following are the available compiling option flags that you can specify in *
...
++-------------------------+-------------------+------------------------------------+
+| -DCUTENSOR_ROOT         |  (unset)          | Install root of cuTENSOR. Required |
+|                         |                   | when USE_CUTENSOR=ON. May also be  |
+|                         |                   | given as an environment variable.  |
++-------------------------+-------------------+------------------------------------+
+| -DCUQUANTUM_ROOT        |  (unset)          | Install root of cuQuantum.         |
+|                         |                   | Required when USE_CUQUANTUM=ON.    |
+|                         |                   | May also be given as an            |
+|                         |                   | environment variable.              |
++-------------------------+-------------------+------------------------------------+
```
IvanaGyro, line 297: This comment is out of scope. I feel we should remove these two cmake flags. The usual way for cmake to find a package which is not installed on the default path is to append `CMAKE_PREFIX_PATH`. Users should append entries to `CMAKE_PREFIX_PATH` if they have unusual system settings.

docs/source/adv_install.rst
```diff
@@ -132,27 +132,41 @@ There are two methods how you can set-up all the dependencies before starting th
...
 .. code-block:: shell
 
-    $conda install -c nvidia cuda
+    $conda install -c nvidia cuda cutensor cuquantum
 
-If cutensor shall be used as well (compile option -DUSE_CUTENSOR=ON), then further install the following:
+.. note::
 
-.. code-block:: shell
+    **A CUDA build needs cuTENSOR and cuQuantum as well, not just the CUDA toolkit.**
```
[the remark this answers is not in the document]
IvanaGyro: The setting in CMakeLists.txt is for backward compatibility.

docs/source/adv_install.rst
```diff
@@ -226,7 +240,7 @@ The following are the available compiling option flags that you can specify in *
 +------------------------+-------------------+------------------------------------+
 |       options          | default           |          description               |
 +------------------------+-------------------+------------------------------------+
-| -DCMAKE_INSTALL_PREFIX | /usr/local/cytnx  | Install destination of the library |
+| -DCMAKE_INSTALL_PREFIX | ~/.local/cytnx    | Install destination of the library |
```
IvanaGyro, line 243: Also an out-of-scope comment. [Cmake official document mentions](https://cmake.org/cmake/help/latest/guide/tutorial/Installation%20Commands%20and%20Concepts.html)

> Note. We have discouraged setting CMAKE_ variables inside the project. Setting CMAKE_INSTALL_PREFIX is particularly bad practice without very good reasoning for doing so, since it prevents users from ever overriding it. When providing a default, projects should check CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT.

We should consider to leave CMAKE_INSTALL_PREFIX as default in the preset file and the cmake files.

IvanaGyro:
I found the reviews for workflow related and doc related PR are usually ignored recently.

yingjerkao:
> I found the reviews for workflow related and doc related PR are usually ignored recently.

Your review comments are taken. I will revert the changes. However, I still think requiring cuQuantum/cuTensor at the same time is a better solution since QR only lives in cuQuantum, not in cuSolve

Issue #1144, opened 2026-07-29
cuBLAS/cuSOLVER handles created and destroyed on every call (52 live sites)


yingjerkao:
Same anti-pattern as #1132, one library down: `linalg_internal_gpu/` creates and destroys a cuBLAS or cuSOLVER handle around **every single call**, at **52 live sites across 13 files** (4 per file, one per dtype specialisation).

```cpp
// cuMatmul_internal.cu:30 — and 51 more like it
checkCudaErrors(cublasCreate(&cublasH));
...
cublasDestroy(cublasH);
```

| library | files | sites |
|---|---|---|
| cuBLAS | `cuGemm`, `cuGemm_Batch`, `cuGer`, `cuMatmul`, `cuMatmul_dg`, `cuMatvec`, `cuNorm`, `cuVectordot` | 32 |
| cuSOLVER | `cuDet`, `cuEigh`, `cuGeSvd`, `cuInvM_inplace`, `cuSvd` | 20 |

All 13 verified live: present in `linalg_internal_gpu/CMakeLists.txt`, with object files in a CUDA build and 5–12 undefined references each from other translation units. (I checked, having just been burned by #1142.)

## Measured cost

RTX 4070 Ti SUPER, CUDA 13, steady state over 200 iterations after warmup:

| | first call | steady state |
|---|---|---|
| `cublasCreate` + `cublasDestroy` | 2761.9 µs | **326.5 µs** |
| `cusolverDnCreate` + `cusolverDnDestroy` | 647.6 µs | **456.4 µs** |

Smaller than cuTENSOR's ~3.4 ms (#1132), but on far hotter operations. Against the work it wraps — `cublasDgemm` on an n×n double matrix, handle cost measured in the same process:

| n | dgemm | handle create+destroy | ratio |
|---|---|---|---|
| 32 | 9.5 µs | 186.5 µs | **19.6×** |
| 64 | 16.8 µs | 186.3 µs | **11.1×** |
| 128 | 31.1 µs | 195.3 µs | **6.3×** |
| 256 | 59.9 µs | 207.0 µs | **3.5×** |
| 512 | 462.4 µs | 299.3 µs | 0.6× |

Below roughly n=400 the handle costs more than the matrix multiply it exists to perform. Tensor-network workloads spend much of their time on small-to-medium contractions, so this is the common case rather than the corner case.

## Fix

[...]

PR #1147, opened 2026-07-29
fix(linalg): scope GPU cuSOLVER resources so error paths cannot leak (#1145, #1146)

### LINKED ISSUE #1145: cuGeSvd leaks a gesvdjInfo_t on every GPU SVD (~2 KB host per call)
`cuGeSvd_internal.cu` creates a `gesvdjInfo_t` on every call and never destroys it. All four dtype specialisations are affected.

```cpp
gesvdjInfo_t gesvdj_params = nullptr;
checkCudaErrors(cusolverDnCreateGesvdjInfo(&gesvdj_params));   // :53
...
checkCudaErrors(cusolverDnZgesvdj(..., gesvdj_params));        // :72
...
checkCudaErrors(cudaFree(devinfo));                            // :94
checkCudaErrors(cusolverDnDestroy(cusolverH));                 // :95  <- end of function
```

The cleanup block releases `d_work`, `Mij`, `UMem`, `vTMem`, `devinfo` and the cuSOLVER handle, but not the Jacobi params object. `cusolverDnDestroyGesvdjInfo` appears **zero** times in the repo, though it exists immediately after its Create in the CUDA header (`cusolverDn.h:3800-3802`).

This is reachable from user code: `cuGeSvd_internal_*` is wired into the `cuGeSvd_ii` dispatch table for all four dtypes (`linalg_internal_interface.cpp:337-340`).

## Measured

Creating 100,000 `gesvdjInfo_t` without destroying, versus the identical loop with the destroy added:

| | per object |
|---|---|
| host RSS | **2064 bytes** |
| device memory | 0 bytes |
| control, with `cusolverDnDestroyGesvdjInfo` | 0 bytes |

100k leaked objects grew host RSS by ~197 MB; the control loop grew it by 4 KB in total. So the leak is exactly the missing call, and it is host memory only — nothing leaks on the device.

## Impact

~2 KB of host memory per GPU SVD, unbounded. A DMRG-style run doing 100k SVDs leaks roughly 200 MB. Slow enough not to look like a crash — it would present as a long-running simulation gradually consuming RAM.

## Fix

One line per specialisation, alongside the existing `cusolverDnDestroy(cusolverH)`:

```cpp
checkCudaErrors(cusolverDnDestroyGesvdjInfo(gesvdj_params));
```

Note that this line will not run on the error path — `cytnx_error_msg` at :82 throws before the cleanup block is reached. That is a separate and larger problem, filed separately; a fix here should be placed so it survives the throw rather than simply appended to the tail of the function.

Found while auditing handle lifetimes for #1144.

### LINKED ISSUE #1146: GPU cuSOLVER wrappers leak device memory on their error paths (throw before cleanup, 5 files)
Every live cuSOLVER wrapper in `linalg_internal_gpu/` raises its LAPACK-`info` error **before** freeing the device memory and handle it allocated. `cytnx_error_msg` expands to a `[[noreturn]]` call (`cytnx_error.hpp:92`), so the stack unwinds past the cleanup block and everything allocated in the function leaks — including **device** memory.

`cuGeSvd_internal.cu`, first specialisation:

```cpp
cytnx_error_msg(info != 0, "%s %d %s",
                "Error in cuBlas function 'cusolverDnZgesvdj': cuBlas INFO = ", info,
                "If info>0, possibly svd not converge, ...");   // :82  THROWS
checkCudaErrors(cudaFree(d_work));                              // :86  never runs
checkCudaErrors(cudaFree(Mij));
if (UMem != nullptr and U->dtype() == Type.Void) cudaFree(UMem);
if (vTMem != nullptr and vT->dtype() == Type.Void) cudaFree(vTMem);
checkCudaErrors(cudaFree(devinfo));
checkCudaErrors(cusolverDnDestroy(cusolverH));                  // :95
```

Note what the error message says the failure mode is: *"possibly svd not converge"*. Non-convergence is a normal runtime outcome on ill-conditioned input, not a programming error — so this is a leak on a path that is expected to be taken.

## All five live cuSOLVER files have the same shape

| file | throws at | first cleanup | device allocations live at the throw |
|---|---|---|---|
| `cuGeSvd_internal.cu` | :82 | :86 | `Mij`, `d_work`, `devinfo`, optionally `UMem`, `vTMem` |
| `cuSvd_internal.cu` | :104 | :107 | `Mij`, `d_work`, `devinfo`, optionally `UMem`, `vTMem` |
| `cuEigh_internal.cu` | :55 | :58 | `tA`, `work`, `devinfo` |
| `cuDet_internal.cu` | :37 | :51 | `devIpiv`, `devInfo`, `workspace`, `_in` |
| `cuInvM_inplace_internal.cu` | :30, :46 | :52 | `ipiv`, `devinfo`, `d_work`; plus `d_I` at the second throw |

Each also leaks its `cusolverDnHandle_t`, and `cuGeSvd` additionally leaks the `gesvdjInfo_t` covered in the companion issue.

`cuInvM_inplace_internal.cu` has two throw sites; the second (:46) is after `d_I` is allocated at :36, so it leaks more than the first.

## Magnitude

Unlike the `gesvdjInfo_t` leak, this is VRAM. `Mij` alone is `M * N * sizeof(dtype)` — 16 MB for a 1000×1000 `complex128` — and `UMem`/`vTMem` are `max² * sizeof(dtype)` each when allocated internally, so a single failed SVD of that size can leak on the order of 48 MB plus workspace.

`cytnx::error` surfaces in Python as `CytnxError`, which is catchable. So this is straightforwardly reachable:

```python
for m in matrices:            # ill-conditioned input
    try:
        cytnx.linalg.Svd(m)   # leaks tens of MB of VRAM per failure
    except cytnx.CytnxError:
        continue              # ... until the GPU is out of memory
```

A user handling SVD failures — the obvious thing to do with a catchable convergence error — will exhaust VRAM.

## Fix

Appending the frees is not sufficient, since the throw is what skips them. These need the allocations to release themselves during unwinding: an RAII guard for device buffers and one for each library handle, so that cleanup is not on the happy path only. That would also fix the `gesvdjInfo_t` leak structurally rather than by adding one more line that the throw can skip.

If a minimal change is preferred, freeing before the throw works but has to be repeated at every throw site in every specialisation — 5 files × 4 dtypes, and `cuInvM` has two sites each.

This interacts with #1144: if the cuSOLVER handle becomes a cached per-device handle, it stops being one of the leaked resources, which removes part of the problem but none of the device-memory part.

Found while auditing handle lifetimes for #1144.

yingjerkao:
Closes #1145. Closes #1146.

Fixes both together because #1146's fix subsumes #1145: adding the missing `cusolverDnDestroyGesvdjInfo` at the tail of `cuGeSvd_internal` would have been skipped by the very throw that #1146 is about.

## Problem

Every live cuSOLVER wrapper in `linalg_internal_gpu/` raised its LAPACK-`info` error *before* the cleanup block at the end of the function:

```cpp
cytnx_error_msg(info != 0, ... "possibly svd not converge" ...);   // cuGeSvd_internal.cu:82
checkCudaErrors(cudaFree(d_work));                                 // :86  never runs
checkCudaErrors(cudaFree(Mij));
```

`cytnx_error_msg` expands to a `[[noreturn]]` call, so unwinding skipped every `cudaFree` and the handle. Note what the message says the trigger is — non-convergence is a normal outcome on ill-conditioned input, not a programming error, so this leaked on an expected path. `cytnx::error` surfaces in Python as the catchable `CytnxError`, making the obvious retry loop a VRAM exhaustion loop.

Separately, `cuGeSvd_internal.cu` never destroyed its `gesvdjInfo_t` at all, on any path (#1145).

Five files, four dtype specialisations each: `cuGeSvd`, `cuSvd`, `cuEigh`, `cuDet`, `cuInvM_inplace`. `cuInvM` has two throw sites with *different* sets of live allocations, and `cuSvd` has a third exit via `throw std::runtime_error`.

## Measured, before and after

The same 200-call loop over a singular matrix (`InvM` on a rank-deficient 4×4, which makes `getrf` report `info > 0`):

| | leaked | per call |
|---|---|---|
| **pre-fix** | **2,465,792 KiB (~2.4 GB)** | ~12.6 MB |
| **post-fix** | −16,384 KiB (16 MB *more* free) | none |

Device free memory went 15393 MB → 12985 MB pre-fix, and did not fall at all post-fix. The negative post-fix figure is the allocator releasing cached blocks.

**12.6 MB per call for a 4×4 matrix is the interesting part**: the dominant leak is the leaked `cusolverDnHandle_t` itself, not the small work buffers. I had expected the buffers to matter most and was wrong — the handle carries the cost, which also means #1144 (caching these handles) and this fix interact.

## Fix

A new `utils_internal/cuScopedResource_gpu.hpp` with three owning types — `DeviceBuffer<T>`, `CusolverDnHandle`, `GesvdjInfo` — and all 20 functions converted so cleanup is independent of the exit path. No manual `cudaFree`, `cusolverDnCreate`/`Destroy`, or `CreateGesvdjInfo` remains in any of the five files.

Reordering the frees before the throw was the smaller change, but it does not work here: `cuInvM`'s first throw site must free only what has been allocated *so far*, and hand-tracking that per site is exactly the bug class being fixed.

Three details worth review:

**Conditional ownership is preserved exactly.** `cuEigh`'s `tA` aliases `v`'s storage when eigenvectors are wanted and is a temporary otherwise; `cuSvd`/`cuGeSvd`'s `UMem`/`vTMem` likewise. The old code expressed this as `if (dtype() == Type.Void) cudaFree(...)`. The converted code keeps a separate `owned_*` buffer that is non-empty only in the temporary case, so ownership matches the old conditional rather than approximating it.

**Zero-size allocations are handled deliberately.** `DeviceBuffer(0)` allocates nothing and stays null. `lwork` legitimately comes back as 0 from the cuSOLVER buffer-size queries, and `cudaMalloc(ptr, 0)` yields a pointer that cannot be freed — the defect fixed for `Storage` in #1126. This is a behaviour change in that edge case, in the safe direction.

**`checkCudaErrors` does not throw** — it prints and calls `exit(EXIT_FAILURE)` (`cytnx_error.hpp:526`). I checked this before choosing the fix: it means `cytnx_error_msg` is the *only* exception source in these functions, so the throw sites above are the complete set rather than a sample.

`cuDet` also allocates via `cuMalloc_gpu` (managed memory, since it reads the factorized diagonal from the host), so `DeviceBuffer::adopt` takes ownership of a pointer allocated elsewhere. `cudaFree` releases both kinds, so ownership is uniform even though allocation is not.

Host allocations on the same paths are now `std::vector` rather than `malloc`/`new[]`, which also removes their leak on the throw.

## Testing

`tests/gpu/linalg_test/CusolverResourceLeak_test.cpp`:

- **`GpuInvMErrorPathDoesNotLeakDeviceMemory`** — the regression test. Shown to fail pre-fix: the pre-fix code leaks ~2.4 GB where the threshold is 64 MiB, a margin of ~40×, so it is not threshold-sensitive.
- **`GpuCusolverSuccessPathsDoNotLeakDeviceMemory`** — passed before the fix, so not a regression test; it guards the refactor itself against a missed free or a double free via a stale alias, across all five converted paths.
- **`GpuInvMStillCorrectAfterScopedResources`** — checks the inverse of a diagonal matrix against the independently known reciprocal diagonal, not against another Cytnx path. The conversion rewired which pointers the cuSOLVER calls receive, so numerical confirmation is warranted.

Full GPU suite locally, `ctest -L gpu`:

```
100% tests passed, 0 tests failed out of 820
Total Test time (real) = 1686.34 sec
```

817 pre-existing plus the 3 added here. The 4 skips are pre-existing `GTEST_SKIP`s in files this PR does not touch.

**One gap, stated rather than papered over:** #1145's leak is ~2 KB of *host* memory per call, so a CI-robust threshold test would need ~20,000 GPU SVDs to clear allocator noise — too slow to be worth it. I verified it directly instead: 100,000 `gesvdjInfo_t` created without destroy grew host RSS by 2064 bytes each, and the same loop with `cusolverDnDestroyGesvdjInfo` grew it by 4 KB in total. In this PR it is covered structurally by the `GesvdjInfo` type and by the device-memory test exercising the same code path.

### DIFF
--- a/src/backend/linalg_internal_gpu/cuDet_internal.cu
+++ b/src/backend/linalg_internal_gpu/cuDet_internal.cu
@@ -17,9 +17,8 @@ namespace cytnx {
       cytnx_complex128* od = (cytnx_complex128*)out;  // result on cpu!
       // Scoped resources (#1146): the info check below throws, so ownership -- not a cleanup block
       // at the tail of the function -- is what keeps these from leaking.
-      auto _in = utils_internal::DeviceBuffer<cuDoubleComplex>::adopt(
-        (cuDoubleComplex*)utils_internal::cuMalloc_gpu(in->size() *
-                                                       sizeof(cuDoubleComplex)));  // unify mem.
+      // Managed (unified) memory: the diagonal is read from the host below.
+      auto _in = utils_internal::DeviceBuffer<cuDoubleComplex>::managed(in->size());
       checkCudaErrors(cudaMemcpy(_in.get(), in->data(), sizeof(cytnx_complex128) * in->size(),
                                  cudaMemcpyDeviceToDevice));
 
@@ -57,9 +56,8 @@ namespace cytnx {
     void cuDet_internal_cf(void* out, const boost::intrusive_ptr<Storage_base>& in,
                            const cytnx_uint64& L) {
       cytnx_complex64* od = (cytnx_complex64*)out;  // result on cpu!
-      auto _in = utils_internal::DeviceBuffer<cuFloatComplex>::adopt(
-        (cuFloatComplex*)utils_internal::cuMalloc_gpu(in->size() *
-                                                      sizeof(cuFloatComplex)));  // unify mem.
+      // Managed (unified) memory: the diagonal is read from the host below.
+      auto _in = utils_internal::DeviceBuffer<cuFloatComplex>::managed(in->size());
       checkCudaErrors(cudaMemcpy(_in.get(), in->data(), sizeof(cytnx_complex64) * in->size(),
                                  cudaMemcpyDeviceToDevice));
 
@@ -97,9 +95,8 @@ namespace cytnx {
     void cuDet_internal_d(void* out, const boost::intrusive_ptr<Storage_base>& in,
                           const cytnx_uint64& L) {
       cytnx_double* od = (cytnx_double*)out;  // result on cpu!
-      auto _in = utils_internal::DeviceBuffer<cytnx_double>::adopt(
-        (cytnx_double*)utils_internal::cuMalloc_gpu(in->size() *
-                                                    sizeof(cytnx_double)));  // unify mem.
+      // Managed (unified) memory: the diagonal is read from the host below.
+      auto _in = utils_internal::DeviceBuffer<cytnx_double>::managed(in->size());
       checkCudaErrors(cudaMemcpy(_in.get(), in->data(), sizeof(cytnx_double) * in->size(),
                                  cudaMemcpyDeviceToDevice));
 
@@ -137,8 +134,8 @@ namespace cytnx {
     void cuDet_internal_f(void* out, const boost::intrusive_ptr<Storage_base>& in,
                           const cytnx_uint64& L) {
       cytnx_float* od = (cytnx_float*)out;  // result on cpu!
-      auto _in = utils_internal::DeviceBuffer<cytnx_float>::adopt((
-        cytnx_float*)utils_internal::cuMalloc_gpu(in->size() * sizeof(cytnx_float)));  // unify mem.
+      // Managed (unified) memory: the diagonal is read from the host below.
+      auto _in = utils_internal::DeviceBuffer<cytnx_float>::managed(in->size());
       checkCudaErrors(cudaMemcpy(_in.get(), in->data(), sizeof(cytnx_float) * in->size(),
                                  cudaMemcpyDeviceToDevice));
 
--- a/src/backend/linalg_internal_gpu/cuDet_internal.cu
+++ b/src/backend/linalg_internal_gpu/cuDet_internal.cu
@@ -4,6 +4,9 @@
 #include "backend/lapack_wrapper.hpp"
 
 #include "backend/utils_internal_gpu/cuAlloc_gpu.hpp"
+#include "backend/utils_internal_gpu/cuScopedResource_gpu.hpp"
+
+#include <vector>
 
 namespace cytnx {
 
@@ -12,28 +15,28 @@ namespace cytnx {
     void cuDet_internal_cd(void* out, const boost::intrusive_ptr<Storage_base>& in,
                            const cytnx_uint64& L) {
       cytnx_complex128* od = (cytnx_complex128*)out;  // result on cpu!
-      cuDoubleComplex* _in = (cuDoubleComplex*)utils_internal::cuMalloc_gpu(
-        in->size() * sizeof(cuDoubleComplex));  // unify mem.
-      checkCudaErrors(cudaMemcpy(_in, in->data(), sizeof(cytnx_complex128) * in->size(),
+      // Scoped resources (#1146): the info check below throws, so ownership -- not a cleanup block
+      // at the tail of the function -- is what keeps these from leaking.
+      auto _in = utils_internal::DeviceBuffer<cuDoubleComplex>::adopt(
+        (cuDoubleComplex*)utils_internal::cuMalloc_gpu(in->size() *
+                                                       sizeof(cuDoubleComplex)));  // unify mem.
+      checkCudaErrors(cudaMemcpy(_in.get(), in->data(), sizeof(cytnx_complex128) * in->size(),
                                  cudaMemcpyDeviceToDevice));
 
-      cusolverDnHandle_t cusolverH;
-      cusolverDnCreate(&cusolverH);
+      utils_internal::CusolverDnHandle cusolverH;
 
-      int* devIpiv;
-      int* devInfo;
-      checkCudaErrors(cudaMalloc((void**)&devIpiv, L * sizeof(int)));
-      checkCudaErrors(cudaMalloc((void**)&devInfo, sizeof(int)));
+      utils_internal::DeviceBuffer<int> devIpiv(L);
+      utils_internal::DeviceBuffer<int> devInfo(1);
 
       int workspace_size = 0;
-      cuDoubleComplex* workspace = nullptr;
-      cusolverDnZgetrf_bufferSize(cusolverH, L, L, _in, L, &workspace_size);
-      checkCudaErrors(cudaMalloc((void**)&workspace, workspace_size * sizeof(cuDoubleComplex)));
+      cusolverDnZgetrf_bufferSize(cusolverH.get(), L, L, _in.get(), L, &workspace_size);
+      utils_internal::DeviceBuffer<cuDoubleComplex> workspace(workspace_size);
 
-      cusolverDnZgetrf(cusolverH, L, L, _in, L, workspace, devIpiv, devInfo);
+      cusolverDnZgetrf(cusolverH.get(), L, L, _in.get(), L, workspace.get(), devIpiv.get(),
+                       devInfo.get());
 
       int info;
-      checkCudaErrors(cudaMemcpy(&info, devInfo, sizeof(int), cudaMemcpyDeviceToHost));
+      checkCudaErrors(cudaMemcpy(&info, devInfo.get(), sizeof(int), cudaMemcpyDeviceToHost));
       cytnx_error_msg(info < 0, "[ERROR] cusolverDnZgetrf fail with info= %d\n", info);
       // TODO: info > 0 means U[info - 1, info - 1] is zero, which implies the determinant is
       // zero. The steps below can be skipped.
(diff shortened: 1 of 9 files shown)


src/backend/utils_internal_gpu/cuScopedResource_gpu.hpp
```diff
@@ -0,0 +1,130 @@
...
+#include <cstddef>
+#include <utility>
+
+#include "cytnx_error.hpp"
+#include "Type.hpp"
+
+#if defined(UNI_GPU)
+  #include <cuda_runtime.h>
+  #include <cusolverDn.h>
+#endif
```
IvanaGyro: How about using

```cpp
#if !define(UNI_GPU)
  static_assert(...)
#endif
```

src/backend/utils_internal_gpu/cuScopedResource_gpu.hpp
```diff
@@ -0,0 +1,130 @@
...
+          if (ptr) cudaFree(ptr);
+          ptr = std::exchange(other.ptr, nullptr);
+        }
+        return *this;
+      }
+
+      /**
+       * @brief Takes ownership of a pointer allocated elsewhere, e.g. by `cuMalloc_gpu`.
+       *
+       * Needed where the buffer must be *managed* memory rather than plain device memory --
```
IvanaGyro: Is this mean the input ptr must be managed? The comment is a little bit confused.

src/backend/utils_internal_gpu/cuScopedResource_gpu.hpp
```diff
@@ -0,0 +1,130 @@
...
+     * in the function (#1146). Ownership makes cleanup independent of the exit path, which manual
+     * `cudaFree` calls at the tail of a function cannot be -- especially where a function has more
+     * than one throw site and each has a different set of live allocations.
+     *
+     * A zero-element request leaves the buffer empty and allocates nothing. `cudaMalloc(ptr, 0)`
+     * yields a pointer that cannot be freed, which is the defect fixed for `Storage` in #1126;
+     * `lwork` legitimately comes back as 0 from the cuSOLVER buffer-size queries, so this case is
+     * reachable here rather than hypothetical.
+     */
+    template <class T>
```
IvanaGyro, line 45: Does it require cuda type?

src/backend/linalg_internal_gpu/cuDet_internal.cu
```diff
@@ -12,188 +15,161 @@ namespace cytnx {
...
                            const cytnx_uint64& L) {
       cytnx_complex128* od = (cytnx_complex128*)out;  // result on cpu!
-      cuDoubleComplex* _in = (cuDoubleComplex*)utils_internal::cuMalloc_gpu(
-        in->size() * sizeof(cuDoubleComplex));  // unify mem.
-      checkCudaErrors(cudaMemcpy(_in, in->data(), sizeof(cytnx_complex128) * in->size(),
+      // Scoped resources (#1146): the info check below throws, so ownership -- not a cleanup block
+      // at the tail of the function -- is what keeps these from leaking.
+      auto _in = utils_internal::DeviceBuffer<cuDoubleComplex>::adopt(
+        (cuDoubleComplex*)utils_internal::cuMalloc_gpu(in->size() *
+                                                       sizeof(cuDoubleComplex)));  // unify mem.
```
IvanaGyro: How about let DeviceBuffer also handle manged memory creation?

IvanaGyro (approved):

src/backend/utils_internal_gpu/cuScopedResource_gpu.hpp
yingjerkao: Adopted the intent in `b7fbfe4c`, with `#error` rather than `static_assert`.

You are right that the guard was wrong as written. CMake only adds `backend/utils_internal_gpu` and `backend/linalg_internal_gpu` under `USE_CUDA`, so `#if defined(UNI_GPU)` could never be false in a real build — all it did was let an accidental include from a CPU translation unit expand to nothing and fail later as "DeviceBuffer is not a member of `cytnx::utils_internal`", pointing at the use site instead of the include.

I went with `#error` because the failure is about *inclusion*, not instantiation: it reports during preprocessing, before the `<cuda_runtime.h>` / `<cusolverDn.h>` includes below it are parsed, so the diagnostic is the message rather than a pile of missing-header errors. Verified it fires:

```
$ g++ -std=gnu++20 -fsyntax-only -I src -I include guardcheck.cpp
src/backend/utils_internal_gpu/cuScopedResource_gpu.hpp:10:4: error: #error
  "cuScopedResource_gpu.hpp is CUDA-only; include it from a .cu under backend/*_gpu."
```

(Also noting the small typo for anyone copying the snippet later: `#if !defined(...)`, not `#if !define(...)`.)

src/backend/utils_internal_gpu/cuScopedResource_gpu.hpp
yingjerkao: It did not — and the comment was describing a constraint on the *caller*, which is exactly the confusing part. `adopt()` accepted any pointer; managed memory was simply the only reason anyone used it.

`b7fbfe4c` removes the ambiguity by removing `adopt()` (see the `cuDet_internal` thread): allocation now happens inside the class, so there is no caller-supplied pointer to have a rule about. The comment now states which allocator each entry point uses and why `cuDet_internal` needs the host-accessible one.

src/backend/utils_internal_gpu/cuScopedResource_gpu.hpp
yingjerkao: No — `T` fixes `sizeof(T)` for the allocation and the pointer type `get()` returns, nothing more. In practice these buffers hold `int` (`devIpiv`, `devInfo`), `cytnx_double`, and `cuDoubleComplex` side by side.

Documented on the class in `b7fbfe4c`, since the question is a fair one to have when the type is named `DeviceBuffer` and sits in a CUDA-only header. The one caveat now recorded with it: nothing in the buffer is ever safe to dereference on the host unless the storage came from `managed()`.

src/backend/linalg_internal_gpu/cuDet_internal.cu
yingjerkao: Good call — done in `b7fbfe4c`, and it let `adopt()` go away entirely.

Before, the allocation sat at the call site and ownership was bolted on afterwards:

```cpp
auto _in = utils_internal::DeviceBuffer<cuDoubleComplex>::adopt(
  (cuDoubleComplex*)utils_internal::cuMalloc_gpu(in->size() * sizeof(cuDoubleComplex)));
```

Now:

```cpp
// Managed (unified) memory: the diagonal is read from the host below.
auto _in = utils_internal::DeviceBuffer<cuDoubleComplex>::managed(in->size());
```

That removes the cast, the manual `count * sizeof(T)`, and the escape hatch — all four dtype overloads here were `adopt()`'s only callers, so it is deleted rather than left as a way to bypass the class's own allocation. Only the allocator differs between `managed()` and the constructor; `cudaFree` releases both kinds, so teardown, move semantics and the zero-element case are shared.

`DetGpu.*` (all four sites, including `GpuHandleAllDtype`) and `CusolverResourceLeak.*` pass.

src/backend/linalg_internal_gpu/cuGeSvd_internal.cu
```diff
@@ -22,77 +23,72 @@ namespace cytnx {
...
 
       cytnx_int64 min = std::min(M, N);
       cytnx_int64 max = std::max(M, N);
       cytnx_int64 ldA = N, ldu = N, ldvT = M;
 
+      // UMem/vTMem alias U's and vT's storage when it exists and are temporaries otherwise; the
+      // owned_* buffers are non-empty only in the temporary case, matching the old conditional
+      // cudaFree on `dtype() == Type.Void`.
       void *UMem = nullptr, *vTMem = nullptr;
+      utils_internal::DeviceBuffer<data_type> owned_UMem, owned_vTMem;
```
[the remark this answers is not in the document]
yingjerkao: Done in `9d5aa2a1`:

```
owned_UMem  -> owned_u_mem
owned_vTMem -> owned_vt_mem
owned_tA    -> owned_ta
```

`owned_d_work` and `owned_h_work` were already snake_case. The pointers these own (`UMem`, `vTMem`, `tA`) keep their names — they are pre-existing identifiers, and renaming them would spread the diff into untouched code for no benefit, which the touched-code policy explicitly warns against.

Rename only, no behaviour change. `CusolverResourceLeak.*`, `DetGpu.*`, `Svd.*` and `GeSvd.*`: 20 passed, 1 pre-existing skip.

PR #1169, opened 2026-08-02
fix(network): validate ORDER expressions

### LINKED ISSUE #832: Network ORDER-line parser accepts inputs that violate the documented API
## Summary

The `ORDER` line parser used by `Network::Launch` accepts a much wider grammar than the API documents. The doc promises "standard mathematical bracket rules", but the parser only checks for a forbidden-character set and balanced parentheses — almost any comma/paren soup gets through. We should tighten the parser to match the spec (or, if the spec is intentionally loose, tighten the spec to match a clearly defined grammar).

## Documented API

From `include/Network.hpp:298-302`:

> About [ORDER]:
> - The contraction order, it can be specify using the standard mathmetical bracket ruls.
> - Without specify this line, the default contraction order will be from the first line to the last line

The example file `example/Network/example.net:6` demonstrates the intended form:

```
ORDER: (A,B),(C,D)
```

i.e. tensor names separated by `,`, optionally grouped by matched `(` `)` pairs, forming a well-formed bracket expression.

## Actual parser behavior

`_parse_ORDER_line_` in `src/RegularNetwork.cpp:18-38` (duplicated in `src/RegularGncon.cpp:58-78`) only does the following:

1. Rejects any line containing characters in `"\t;\n:"`.
2. Requires at least one occurrence of `(`, `)`, or `,`.
3. Verifies that the `(` and `)` counts match.
4. Tokenizes via `str_findall(line, "(),")` (defined in `src/utils/str_utils.cpp:50-78`), which only splits on those delimiters and drops empty segments.

There is no check that the bracket expression is well-formed, no check that operators appear in a meaningful order, no check that tensor-name characters are restricted to the documented `A-Z, a-z` set, and no check that the parenthesization expresses a binary contraction tree.

## Examples that the parser currently accepts but the spec forbids

All of the following are accepted today on `claude/tensornet-launch-cuda-issues-kPaHw` (matching the documented format only requires "standard mathematical bracket rules"):

- `A,B,C,D` — no parentheses at all, but "bracket rules" implies brackets express the order.
- `(A,B,C,D)` — single flat group, no contraction order specified.
- `((A,B)),(C,D)` — redundant nested parens.
- `(A,B,)` — trailing comma producing an empty operand.
- `),A,B,(` — `(` and `)` counts match, but the expression is not well-formed.
- `()(),A,B,C,D` — empty groups.
- `(A,(B,C,D))` — mixes a binary node with a ternary one; downstream `contraction_tree` only handles binary trees per documented "bracket rules".
- Names with characters outside `A-Z, a-z` (e.g. digits, underscores) — accepted by the ORDER parser even though `include/Network.hpp:288` says "the name can be any alphabets A-Z, a-z".

## Proposal

Tighten `_parse_ORDER_line_` to enforce the documented grammar. A reasonable definition:

```
order   ::= expr
expr    ::= name | '(' expr ',' expr ')'
name    ::= [A-Za-z]+              # matches Network.hpp doc
```

Concretely:
- Reject empty tokens between delimiters.
- Reject characters outside `[A-Za-z(),\s]`.
- After tokenizing, run a single-pass recursive-descent (or shunting-yard) check that the bracket expression is well-formed and produces a binary tree whose leaves are all the input tensors exactly once.
- Surface a clear error message pointing at the offending position.

Also de-duplicate the parser: `src/RegularNetwork.cpp:18-38` and `src/RegularGncon.cpp:58-78` are copy-pasted; both should call a single shared implementation so future fixes don't drift.

## Tests to add

- Round-trip valid expressions (`(A,B)`, `((A,B),C)`, `(A,(B,C))`, …) parse to the expected tree.
- Each example listed above produces a parse error with a useful message.
- Existing `tests/testNet.net` and `example/Network/example.net` still parse.

IvanaGyro:
## Problem

The Network `ORDER` parser only rejected a small character set and compared parenthesis counts. Malformed comma/parenthesis expressions could therefore reach the stack-based contraction-tree builder, where they could be accepted incorrectly or trigger unsafe stack operations.

## Fix

- add a shared recursive-descent parser for binary contraction expressions
- preserve the documented optional outermost parentheses form, including `(A,B),(C,D)`
- reject flat n-ary groups, empty operands, redundant unary groups, misplaced delimiters, and unmatched parentheses with line/column diagnostics
- require every network tensor to appear exactly once before building the contraction tree
- route both RegularNetwork and the legacy RegularGncon parser through the shared implementation
- document the grammar and canonicalize the flat test fixture as `((B,C),A)`

Nonalphabetic tensor names such as digits and underscores remain supported as opaque leaf names.

## Testing

- added focused C++ regression tests for valid trees, whitespace, opaque names, every malformed example from #832, exact error columns, and duplicate/missing/undefined leaves through `setOrder`
- pinned pre-commit hooks pass, including clang-format 14
- local build and test execution intentionally skipped as requested; CI will verify the change

Closes #832

### DIFF
--- a/src/utils/order_parser.cpp
+++ b/src/utils/order_parser.cpp
@@ -1,8 +1,10 @@
 #include "utils/order_parser.hpp"
 
 #include <cctype>
+#include <cstddef>
 #include <cstdlib>
 #include <string>
+#include <utility>
 #include <vector>
 
 #include "cytnx_error.hpp"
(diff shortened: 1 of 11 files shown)


src/utils/order_parser.cpp
```diff
@@ -0,0 +1,121 @@
...
+          }
+          return parsed_tokens;
+        }
+
+       private:
+        const std::string& input_line;
+        std::size_t source_line_num;
+        std::size_t cursor = 0;
+        std::size_t leaf_count = 0;
+        std::vector<std::string> parsed_tokens;
```
IvanaGyro: wrong declaration order. please refer to https://google.github.io/styleguide/cppguide.html#Declaration_Order
And private member var should be followed by `_`.

src/utils/order_parser.cpp
```diff
@@ -0,0 +1,121 @@
...
+                          "[ERROR][ORDER] line:%zu column:%zu invalid contraction order: %s\n",
+                          source_line_num, error_cursor + 1, reason);
+          std::abort();
+        }
+      };
+
+    }  // namespace
+
+    std::vector<std::string> parse_order_line(const std::string& line, std::size_t line_num) {
+      return OrderParser(line, line_num).parse();
```
IvanaGyro, line 126: Will this copy `std::vector<std::string> parsed_tokens;`?

IvanaGyro: Addressed in f309a0d4. `parse()` now returns `std::move(parsed_tokens_)`, so the parser's member vector is moved into the result rather than copied.

src/utils/order_parser.cpp
```diff
@@ -0,0 +1,121 @@
...
+          if (input_line[cursor] != '(') {
+            parse_name();
+            return;
+          }
+
+          parsed_tokens.emplace_back("(");
+          ++cursor;
+          parse_tree();
+          consume(',', "expected ',' between the two contraction operands");
+          parse_tree();
```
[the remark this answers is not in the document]
IvanaGyro: Addressed in f309a0d4. The parser now rejects the 1,025th nested group with an ORDER error before it can recurse further; `NetworkOrderLimitsNestingDepth` covers the limit and its diagnostic.

src/utils/order_parser.cpp
IvanaGyro: Addressed in f309a0d4. I reordered the class so its helper methods precede private data members and renamed every private member with the required trailing underscore.

Issue #1171, opened 2026-08-02
[build] Default CUDA architectures: GPU kernels silently return zeros on sm_80, and a wrong-arch build never reports an error


yingjerkao:
## Summary

A stock `cmake --preset debug-openblas-cuda` build, using the default `CMAKE_CUDA_ARCHITECTURES`, produced a `gpu_test_main` in which **every GPU kernel silently returned zeros** on an NVIDIA A100-SXM4-80GB (sm_80). Not an error, not a crash — wrong numbers. Rebuilding with `-DCMAKE_CUDA_ARCHITECTURES=80` made all the same tests pass.

Two separable problems, the second arguably worse than the first.

## 1. The multi-arch default produces a binary with no runnable code for sm_80

#870 added the portable fat-binary default at `CMakeLists.txt:281`:

```cmake
set(CMAKE_CUDA_ARCHITECTURES 75-real 80-real 86-real 89-real 90-real 90-virtual)
```

A fresh configure confirms all six reach the compile line, so the ordering in `CMakeLists.txt` is right (architectures set at :281, `add_library(cytnx)` at :320, `add_subdirectory(tests)` at :492). But the emitted codegen is **LTO IR only**:

```
--generate-code=arch=compute_75,code=[lto_75]
--generate-code=arch=compute_80,code=[lto_80]
--generate-code=arch=compute_86,code=[lto_86]
--generate-code=arch=compute_89,code=[lto_89]
--generate-code=arch=compute_90,code=[compute_90,lto_90]
```

because `CMAKE_INTERPROCEDURAL_OPTIMIZATION` is forced `TRUE` on every non-Apple host (`CMakeLists.txt:312-317`), so `-real` architectures yield `lto_XX` and the actual SASS is deferred to the device-link step. Only `90-virtual` keeps PTX, and `compute_90` PTX cannot JIT down to sm_80.

**Confirmed mechanism** (thanks to @IvanaGyro below): that deferred device link never happens. Neither `CUDA_SEPARABLE_COMPILATION` nor `CUDA_RESOLVE_DEVICE_SYMBOLS` is set anywhere in the repo, so the `lto_XX` IR is never linked into SASS. The result is a binary containing, for sm_80, neither SASS nor PTX — nothing the driver can run.

[...]

Issue #870, opened 2026-06-02
CUDA build should default to a portable multi-arch binary, overridable via CMAKE_CUDA_ARCHITECTURES


IvanaGyro:
> _Filed by Claude (Claude Code) while preparing an Ubuntu environment-setup script for the `openblas-cuda` preset._

## Context

The CUDA-enabled package is built in **GitHub Actions** and released to **PyPI**. That means:

- The build runner has **no NVIDIA GPU**.
- The resulting wheel must **run on end users' GPUs of many different architectures** (Volta → Hopper and beyond), since we ship one binary to everyone.

## Problem

`CMakeLists.txt` unconditionally forces the CUDA architecture to `native` when `USE_CUDA` is enabled:

```cmake
if(USE_CUDA)
  set(CMAKE_CUDA_ARCHITECTURES native)   # forced, ignores user/preset/cache/env value
  enable_language(CUDA)
  ...
```

This is wrong for a redistributable build in three ways:

1. **`native` requires a GPU at configure time** (nvcc queries the local device), so it fails outright on the GPU-less CI runner.
2. **Even if it succeeded, `native` bakes in only the build host's architecture** — the wheel would not run on any other GPU generation, defeating the point of a PyPI release.
3. Because it is a plain (non-cache) `set()`, it **overrides** anything passed via `-DCMAKE_CUDA_ARCHITECTURES=...`, the cache, a preset, or the environment, so it cannot even be worked around without editing the file.

## Proposed solution

Make `CMakeLists.txt` **default to a portable multi-architecture "fat" binary** when the user has not specified an architecture, and let any user-supplied value take precedence. The default embeds compiled SASS for each supported real architecture plus **PTX for the newest** one (PTX lets the CUDA driver JIT-compile for future/newer GPUs we didn't pre-compile for). This single binary runs across all targeted architectures, and the build needs no GPU.

```cmake
if(USE_CUDA)
  # Default to a portable fat binary: SASS for each supported GPU + PTX of the
  # newest for forward-compatible JIT. Any user/preset/cache/CUDAARCHS value wins.
  if(NOT CMAKE_CUDA_ARCHITECTURES)
    set(CMAKE_CUDA_ARCHITECTURES 70-real 75-real 80-real 86-real 89-real 90-real 90-virtual)
  endif()
  enable_language(CUDA)
  ...
```

Key properties:

[...]

PR #1173, opened 2026-08-04
fix(build): keep the portable CUDA architecture list across reconfigures

yingjerkao:
Relates to #1171 (deliberately does not close it — see *Scope* below).

## Problem

`enable_language(CUDA)` writes CMake's own single-architecture default into the
cache on the first configure. The guard at `CMakeLists.txt:262` that detects a
caller-supplied `CMAKE_CUDA_ARCHITECTURES` then sees that cached value on every
later run and mistakes it for the caller's choice, so the portable fat-binary
default added in #870 is skipped.

Measured on this tree (CUDA 13.0, CMake 4.0.2), configuring the same build dir
repeatedly with identical flags:

| configure | emitted codegen |
|---|---|
| 1st (fresh dir) | `lto_75 lto_80 lto_86 lto_89 lto_90 compute_90` |
| 2nd (same dir) | `compute_75` only |
| 3rd (bare re-run) | `compute_75` only |

Any reconfigure triggers it — a preset re-run, an IDE, or a changed flag — so in
practice a build dir holds the portable list only until the next `cmake` run.
After that the device link emits a single `sm_75` cubin, and every GPU other than
Turing falls back to JIT'ing PTX instead of running the SASS #870 meant to ship.

This also explains the `CMAKE_CUDA_ARCHITECTURES:STRING=75` cache entry noted in
#1171. It is not cosmetic: it is the value that becomes authoritative on the
next configure.

## Fix

Record the list this project installs in an internal cache marker
(`CYTNX_CUDA_ARCHITECTURES_APPLIED`). A cached value that still matches it is
recognized as ours and re-derived; any other value — including a `-D` supplied on
a later configure — is honored as the caller's. The list is written to the cache
rather than shadowing it with a normal variable, so `CMakeCache.txt` and anything
reading it agree with the compile lines.

## Scope

This restores the portable fat binary. It is **not** the wrong-answers bug in
#1171's title, so that issue stays open.

I could not reproduce silent zeros from any build state on sm_89: even the
collapsed single-architecture build computes correctly, because the device link
emits `code=[compute_75,sm_75]` and the driver JITs the PTX. The failure
signature itself does reproduce in isolation — a fatbin carrying SASS with no
compatible PTX gives `value=0, launch=cudaErrorNoKernelImageForDevice`, i.e.
wrong numbers and no error — which is why the launch-error checking in part 2 of
#1171 is the change that would actually have surfaced the A100 session.

Two premises in the #1171 thread are also contradicted by this tree, noted here
so nobody builds on them: `CUDA_SEPARABLE_COMPILATION` and
`CUDA_RESOLVE_DEVICE_SYMBOLS` **are** set (`CytnxBKNDCMakeLists.cmake:253`,
`:255`), a device-link step **is** generated (221 rule lines in
`CMakeFiles/cytnx.dir/build.make`), and the pinned-arch build's archive contains
a real `sm_89.cubin` — so that build executed native SASS rather than JIT'd PTX.
CMake 3.28.3 (the version in the report) generates byte-identical flags.

## Testing

- Configure sequence 1/2/3 above now yields the six-architecture list every time,
  and `CMakeCache.txt` reads
  `75-real;80-real;86-real;89-real;90-real;90-virtual`.
- `-DCMAKE_CUDA_ARCHITECTURES=80` on a later configure is honored and survives a
  subsequent bare re-run.
- Built after a reconfigure, `cmake_device_link.o` contains
  `sm_75/80/86/89/90.cubin` — pre-fix it contained only `sm_75.cubin`.
- Full GPU suite on an RTX 4070 Ti SUPER (sm_89): `ctest -L gpu` → **841/841
  passed**, 4 pre-existing skips.
- `pre-commit run --files CMakeLists.txt` clean.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -258,9 +258,22 @@ if(USE_CUDA)
   # a preset's cacheVariables, or the CUDAARCHS environment variable, before
   # enable_language(CUDA) fills CMAKE_CUDA_ARCHITECTURES with its own default
   # and makes the "not specified" case impossible to detect.
+  #
+  # A *re*configure of an existing build dir needs more than that check: the
+  # first configure leaves a CMAKE_CUDA_ARCHITECTURES cache entry behind, so
+  # every later run sees a non-empty value and would mistake it for a caller's
+  # choice, silently collapsing the portable fat binary below onto whatever
+  # single architecture CMake defaulted to. CYTNX_CUDA_ARCHITECTURES_APPLIED
+  # remembers the list this project installed, so a cache entry that still
+  # matches it is recognized as ours and re-derived, while any other value --
+  # including a -D given on a later configure -- is honored as the caller's.
   set(_cytnx_user_set_cuda_architectures FALSE)
   if(CMAKE_CUDA_ARCHITECTURES OR DEFINED ENV{CUDAARCHS})
     set(_cytnx_user_set_cuda_architectures TRUE)
+    if(DEFINED CYTNX_CUDA_ARCHITECTURES_APPLIED
+       AND CMAKE_CUDA_ARCHITECTURES STREQUAL CYTNX_CUDA_ARCHITECTURES_APPLIED)
+      set(_cytnx_user_set_cuda_architectures FALSE)
+    endif()
   endif()
   # enable_language(CUDA) runs its own internal compiler ABI-detection compile
   # using whatever CMAKE_CUDA_ARCHITECTURES is set to at this point, so our own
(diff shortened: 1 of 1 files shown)


ianmccul:
The underlying reconfigure bug is real, but I think this fix introduces more cache state and provenance inference than is necessary.

`CYTNX_CUDA_ARCHITECTURES_APPLIED` infers ownership by comparing values. That cannot distinguish Cytnx choosing its default from a caller explicitly choosing the same list, and it requires `FORCE`-overwriting a standard user-facing CMake cache variable. It also does not repair build directories created before the marker existed.

`FindCUDAToolkit` is explicitly designed to work without the CUDA language being enabled. We can therefore discover the toolkit version, initialize `CMAKE_CUDA_ARCHITECTURES` only when neither the cache nor first-run `CUDAARCHS` supplied it, and then enable CUDA:

```cmake
if(USE_CUDA)
  find_package(CUDAToolkit REQUIRED)

  if(NOT DEFINED CMAKE_CUDA_ARCHITECTURES
     AND NOT DEFINED ENV{CUDAARCHS})
    if(CUDAToolkit_VERSION VERSION_GREATER_EQUAL 13.0)
      set(cytnx_cuda_architectures
          75-real 80-real 86-real 89-real 90-real 90-virtual)
    else()
      set(cytnx_cuda_architectures
          70-real 75-real 80-real 86-real 89-real 90-real 90-virtual)
    endif()

    set(CMAKE_CUDA_ARCHITECTURES "${cytnx_cuda_architectures}"
        CACHE STRING "CUDA architectures to generate device code for")
  endif()

  enable_language(CUDA)
endif()
```

This follows the documented CMake lifecycle: `CUDAARCHS` initializes the cache on the first configuration, subsequent configurations use the cache, and a later `-DCMAKE_CUDA_ARCHITECTURES=...` naturally replaces it. No project-private ownership marker or `FORCE` is needed.

This PR also exposes a broader structural problem. CUDA setup currently has split ownership: `enable_language(CUDA)` and `find_package(CUDAToolkit)` are repeated in `CytnxBKNDCMakeLists.cmake`, which then mutates global `CMAKE_CUDA_FLAGS`. The build mixes modern target-based CMake with global flags, manually serialized `cxxflags.tmp`/`linkflags.tmp`, duplicated dependency discovery, and cache manipulation. That makes configuration-order bugs such as this one much easier to create.

I suggest simplifying this PR to initialize the architecture cache before `enable_language(CUDA)`. The wider CMake cleanup should be separate.

— Codex

IvanaGyro (commented):
`-DCMAKE_CUDA_ARCHITECTURES` and `ENV{CUDAARCHS}` always take the highest priority in cache, so the solution can be simpler. We just need to change `set(CMAKE_CUDA_ARCHITECTURES 75-real 80-real 86-real 89-real 90-real 90-virtual)` to `      set(CMAKE_CUDA_ARCHITECTURES "75-real;80-real;86-real;89-real;90-real;90-virtual" CACHE STRING "CUDA architectures for CUDA 13+" FORCE)
`

Update: @ianmccul 's solution is clearer.

PR #1174, opened 2026-08-04
test(linalg): GPU mixed real/complex coverage for Sub/Mul/Div/Cpr (#1003)

yingjerkao:
## Problem

#1003 requires that *"tests cover dtype and value correctness for mixed
real/complex cases"*, and names the binary matrix explicitly:

```
ComplexFloat op Double  -> ComplexDouble
Double op ComplexFloat  -> ComplexDouble
...
```

Auditing the GPU suite against that list, only two ops actually covered it:

- `Kron`/`Outer` — `GpuMixedComplexFloatDoublePromotesToComplexDouble`, independent oracle.
- `Add` — `AddTestAllShapes.GpuTensorAddTensorMixedTypes` sweeps the full
  ldtype × rdtype cross product (and `AreNearlyEqTensor` does assert dtype),
  though against the CPU path rather than an independent one.

`Sub`, `Mul`, `Div` and `Cpr` had **no** mixed real/complex GPU coverage at all —
their `*TestAllShapes` sweeps pair a dtype with itself, and their existing
`*MixedDtypeTest` cases cover only signed/unsigned promotion. That is precisely
the #984/#999 discriminator the issue was written around.

## Fix

One test per op, with expected values evaluated in host double arithmetic rather
than taken from the CPU path, so a promotion bug shared by both backends cannot
hide (per `CLAUDE.md`'s independent-expected-values rule):

| Op | What it adds |
|---|---|
| `Sub` | both operand orders — subtraction does not commute |
| `Mul` | both orders, plus a `ComplexFloat` **scalar** against a `Double` tensor, which takes the `lconst`/`rconst` kernels rather than the tensor⊗tensor one |
| `Div` | both orders; these exercise *different* arithmetic — dividing by a real-valued complex reduces to a componentwise divide, while dividing by one with a non-zero imaginary part goes through the full complex-division formula |
| `Cpr` | its promotion is invisible in the output dtype (always `Bool`), so the test pins it with values whose equality *depends* on the comparison type: `0.5` is float-exact and must compare equal, `0.1` is not and must compare unequal |

`Mod` is deliberately absent — `cuMod_dispatch` rejects complex dtypes, so the
mixed real/complex row does not exist for it.

Also corrects the `Storage_base::data()` comment, which still claimed *"GPU
kernel dispatch … the `cuAri_ii` table — ruling 3 keeps GPU on the legacy
table"*. That table was removed earlier in #1003; the dtype dispatch is typed
now, and the erased accessor survives only because CUDA launches need the raw
device address.

## Testing

All four new tests pass on an A100 (`debug-openblas-cuda`), as do the 39 tests
in the surrounding `*MixedDtype*`, `*NonContig*`, `Kron.*` and `Outer.*` suites.

These are coverage tests for already-correct behavior, so "fails before the fix"
does not apply — the meaningful check is whether they would *catch* a narrowed
computation. The `Double` operands are deliberately float32-inexact (`0.1`,
`3.3`); measured off-line, each test's largest element-wise gap under a
ComplexFloat-narrowed computation is **1.2e-8 to 1.9e-7**, four to five orders of
magnitude above the asserted `1e-12` tolerance. Individual components that happen
to be float-exact (e.g. `1.5 / 0.1`) are harmless, since `AreNearlyEqTensor`
fails if *any* element differs.

Completes the last open acceptance criterion of #1003.

### DIFF
--- a/include/backend/Storage.hpp
+++ b/include/backend/Storage.hpp
@@ -75,9 +75,11 @@ namespace cytnx {
     //     call sites than the six arithmetic families converted in #941's
     //     first pass (Add/Sub/Mul/Div/Cpr/Mod) and is out of scope for that
     //     task.
-    //   - GPU kernel dispatch (src/backend/linalg_internal_gpu/*, the
-    //     cuAri_ii table) -- ruling 3 keeps GPU on the legacy table; see GPU
-    //     tracking issue #1013.
+    //   - GPU kernel launches (src/backend/linalg_internal_gpu/*). The dtype
+    //     *dispatch* there is typed now -- as_storage_variant() + type_promote_t,
+    //     with the legacy cuAri_ii table removed (#1003, #1013) -- but the launch
+    //     helpers still reinterpret_cast this erased pointer to the CUDA-native
+    //     kernel type, because CUDA needs the raw device address.
     // New code should use StorageImplementation<T>::data() (typed) via
     // as_storage_variant()/storage_cast<T> instead. Do not add new callers of
     // this accessor.
--- a/tests/gpu/linalg_test/Cpr_test.cpp
+++ b/tests/gpu/linalg_test/Cpr_test.cpp
@@ -214,6 +214,46 @@ namespace cytnx {
         }
       }
 
+      // Cpr's promotion is invisible in the output dtype -- the result is always Bool -- so
+      // a wrong comparison type here cannot be caught by a dtype assertion the way it can
+      // for Add/Sub/Mul/Div. What pins it is a pair of values whose equality *depends* on
+      // the comparison type (#1003).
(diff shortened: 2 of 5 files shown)


manuschneider (approved):
Only Mul exercises the scalar (lconst/rconst) path with mixed complex/real operands — Sub/Div/Cpr scalar variants stay signed/unsigned only. Since the body argues scalar is a distinct path in the shared dispatch, the same row for the other ops might be worth a tiny follow-up (or a sentence on why one op suffices for the shared dispatch).

`ComplexDouble op Float` (the other direction of mixing widths) is also uncovered (but #1003's quoted criterion centers on the ComplexFloat/Double discriminator).

Issue #1142, opened 2026-07-29
Delete cuTNPerm_gpu.{cu,hpp}: orphaned source, never compiled and would not compile


yingjerkao:
`src/backend/utils_internal_gpu/cuTNPerm_gpu.{cu,hpp}` (1507 + 39 lines) is orphaned source. It is not merely uncalled — it is **never compiled**, and it **could not compile** if it were.

Found while converting the per-call `cutensorCreate` sites for #1132: I nearly reported its 11 cuTENSOR call sites as user-facing.

## It is not built

- `cuTNPerm` appears in **no** CMake file anywhere in the repo.
- `src/` CMakeLists use explicit source lists, no `file(GLOB …)`, so nothing sweeps it in implicitly.
- No `cuTNPerm` object file exists in a full CUDA build tree.

For contrast, its sibling `cuMovemem_gpu.cu` is listed at `src/backend/utils_internal_gpu/CMakeLists.txt:22`.

## It has no callers

- `cuTNPerm_gpu.hpp` is `#include`d by exactly one file — its own `.cu`.
- The symbols appear nowhere else in the repo.
- At link level, scanning the 336 object files of a CUDA build finds **0** undefined references to any `cuTNPerm` symbol. The same scan finds 1 for `cuMovemem_gpu` and 5 for `cuTensordot_internal`.

## It would not compile if re-added

Every declaration in the header has a trailing comma in its parameter list:

```cpp
boost::intrusive_ptr<Storage_base> cuTNPerm_gpu_cd(
  boost::intrusive_ptr<Storage_base> &in, const std::vector<cytnx_uint64> &old_shape,
  const std::vector<cytnx_uint64> &mapper, const std::vector<cytnx_uint64> &invmapper, );
//                                                                                    ^
```

Compile-probing the header directly:

```
cuTNPerm_gpu.hpp:18:92: error: expected identifier before ')' token
cuTNPerm_gpu.hpp:22:92: error: expected identifier before ')' token
cuTNPerm_gpu.hpp:27:99: error: expected identifier before ')' token
cuTNPerm_gpu.hpp:32:99: error: expected identifier before ')' token
```

4 errors for 4 declarations. That this has gone unnoticed is itself the proof the file is never built.

The header also declares only **4** functions while the `.cu` defines **11** (`_cd _cf _d _f _i64 _u64 _i32 _u32 _u16 _i16 _b`), so the two have drifted apart as well.

## It is superseded

[...]

PR #1175, opened 2026-08-04
fix(backend): detect failed CUDA kernel launches instead of returning zeros

yingjerkao:
## Problem

A CUDA kernel launch reports its status only through `cudaGetLastError()` — the
launch expression itself cannot fail visibly. Nothing in the GPU backend read
that status, so a launch that never ran left the output buffer holding whatever
it already contained, usually zeros from a fresh allocation. The failure was
indistinguishable from a numerical bug.

This is part 2 of #1171: on an sm_80 card, a default-architecture build returned
zeros across the entire GPU surface — including tests unrelated to any local
change — with no error raised. As that issue puts it, a launch-status check
turns "your kernel is wrong" into "no kernel image for this device".

## Fix

Add `CYTNX_CHECK_CUDA_LAUNCH` (`include/cytnx_error.hpp`, inside the `UNI_GPU`
guard) and call it after every kernel launch in the built GPU sources — 213
sites across 21 files.

Design notes:

- It reads the status **without synchronizing**, so it stays cheap enough to sit
  after every launch. That scopes it to launch-time failures (no image for the
  running device, bad launch configuration) rather than errors raised during
  execution, which surface asynchronously at the next synchronizing call.
- It raises through `cytnx_error_msg`, so the failure is catchable and names the
  kernel plus the launch site — unlike `checkCudaErrors`, which `exit()`s the
  process.
- The pre-existing `checkCudaErrors(cudaGetLastError())` in
  `cuTrace_internal.cu` becomes the same macro rather than a second read that
  could only ever see success.
- `cuTNPerm_gpu.cu` is deliberately left alone: it is not referenced by any build
  file (#1142).

Some `if`/`else` branches gained braces because the check becomes a second
statement; that is the only other change in the diff.

### Trade-off worth a reviewer's attention

30 of the 213 checked launch sites are followed by `cudaFree` of temporary
device buffers (114 free calls in all), which a throw now skips — so a failed
launch leaks device memory where it previously continued silently. An exception
plus a leak beats silently wrong numbers, but scoping those buffers with RAII is
a genuine follow-up rather than something to mix into this change.

Tracked as #1176, with the affected sites enumerated. The `DeviceBuffer<T>` type
that fixes it already exists (added by #1146 for the same failure mode in the
cuSOLVER wrappers), so the follow-up is reuse rather than new machinery.

## Testing

- **GPU suite:** `ctest -L gpu` on 2x RTX 4070 Ti SUPER (sm_89, arch 89 build) —
  **841/841 passed**, 0 failed, 1726 s. The 4 non-running tests are pre-existing
  source-level `GTEST_SKIP()`s (GPU `ExpM` unimplemented, cuQuantum-dependent QR,
  zero-heavy SVD). This is the no-false-positives evidence: the guard stays quiet
  across the whole GPU surface when launches genuinely succeed.
- **Guard fires correctly:** a standalone probe against the real header confirms
  a valid launch does not throw, while a launch rejected by the driver (2048
  threads/block, over the 1024 limit — the same *launch-rejected* class as "no
  kernel image for this device") throws a catchable `cytnx::error`:
  `CUDA kernel launch failed for 'probe_kernel': invalid argument`.
- **End-to-end arch mismatch**, from the session that wrote the patch: same tree,
  library built for `75-real` (SASS with no JIT-able PTX), run on sm_89.

  ```
  before: linalg::Add -> values = 0 0 0 0, no error
  after:  linalg::Add -> CytnxError "CUDA kernel launch failed for 'tn_kernel':
          named symbol not found" at cuArithmeticDispatch.cuh:122
  ```

- **Formatting:** `pre-commit run --files <changed>` clean (clang-format v14).
- **CPU presets not run, by design:** the only non-`.cu`/`.cuh` change is inside
  `#if defined(UNI_GPU)`, so this commit cannot affect a CPU build. CI's
  `ci-cmake_tests` covers that gate.

No automated regression test accompanies this. Reproducing the bug requires
building the whole library for an architecture the running device cannot execute,
which is not something the test suite can set up — hence the manual
demonstration above.

## Scope

Relates to #1171; does **not** close it. Part 1 of that issue — the multi-arch
`CMAKE_CUDA_ARCHITECTURES` default emitting LTO IR that is never device-linked —
is untouched and still wants a fix.

### DIFF
--- a/include/cytnx_error.hpp
+++ b/include/cytnx_error.hpp
@@ -543,6 +543,32 @@ void check(T result, char const *const func, const char *const file, int const l
 
   #endif
 
+  // Raise if the kernel launch immediately preceding this call failed.
+  //
+  // A launch that never runs leaves the output buffer holding whatever it
+  // already contained -- usually zeros from a fresh allocation -- so an
+  // unchecked failure is indistinguishable from a numerical bug. The launch
+  // status is not reported by the launch expression itself; it has to be read
+  // back with cudaGetLastError().
+  //
+  // This reads (and thereby clears) the pending status without synchronizing,
+  // so it stays cheap enough to sit after every launch. That also means it
+  // catches launch-time failures -- a fat binary with no image for the running
+  // device, a bad launch configuration -- and not errors raised while the
+  // kernel executes, which surface asynchronously at the next synchronizing
+  // call.
+  //
+  // Raising through cytnx_error_msg (rather than the exit() that checkCudaErrors
+  // performs) keeps the failure catchable and reports the calling function,
+  // file, and line of the launch site.
+  #define CYTNX_CHECK_CUDA_LAUNCH(kernel_name)                                \
+    do {                                                                      \
+      const cudaError_t cytnx_cuda_launch_status = cudaGetLastError();        \
+      cytnx_error_msg(cytnx_cuda_launch_status != cudaSuccess,                \
+                      "CUDA kernel launch failed for '%s': %s", #kernel_name, \
+                      cudaGetErrorString(cytnx_cuda_launch_status));          \
+    } while (0)
+
 #endif  // End of #if defined(UNI_GPU)
 
 #endif  // CYTNX_CYTNX_ERROR_H_
--- a/src/backend/linalg_internal_gpu/cuArithmeticDispatch.cuh
+++ b/src/backend/linalg_internal_gpu/cuArithmeticDispatch.cuh
@@ -109,17 +109,22 @@ namespace cytnx {
 
         if (Lin->size() == 1 and Rin->size() == 1) {
           constconst_kernel<op_code><<<NBlocks, 512>>>(_out, _Lin[0], len, _Rin[0]);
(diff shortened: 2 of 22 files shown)


manuschneider:
Review with AI help:

The branch now conflicts with master: #1184 landed first (2026-09-17, closing #1176) and rewrote 7 of the files this PR touches with DeviceBuffer RAII. So the intended ordering inverted — needs a rebase, re-placing the checks in those 7 files. This also makes the "Trade-off" section stale in the good direction: after the rebase, a throwing check no longer leaks in the #1184-covered files; it would be good to confirm during the rebase whether all 30 leak-prone sites fall in those files. Since CI doesn't execute GPU kernels, please re-run the local GPU suite on the rebased branch.

Minor, take or leave: cudaGetLastError() also returns (and clears) a sticky error latched by an earlier asynchronous failure, so the message can occasionally blame the named kernel for a predecessor's crash. Might be worth one extra line in the macro comment.

PR #1180, opened 2026-08-14
fix(utils): eliminate UB from empty-vector indexing in vec_concatenate

### LINKED ISSUE #1179: Bugs diagnosed by GCC 15
GCC 15.0 start to enable `_GLIBCXX_ASSERTIONS ` by default for unoptimized build (`-O0`). `_GLIBCXX_ASSERTIONS` raises errors for some STL precondition check. Refer to: https://gcc.gnu.org/pipermail/libstdc%2B%2B-cvs/2024q4/042125.html
Some existing bugs are address by GCC 15.0 and cause the CI test failed.

CI has failed since 2026.08.14 because conda-forge update `compilers` and `cxx-compiler` to 2.0.0 which bump g++ from 14 to 15.

The following tests FAILED:
	 54 - ZeroExtentLinalgTest.ProductsReturnEmptyOrAdditiveIdentity (Failed) cpu
	306 - ArnoldiUt.FermionicArnoldiFermionic (Failed)      cpu
	339 - LanczosGnd.LanczosGndTest (Failed)                cpu
	342 - LanczosGnd.FermionicLanczosFermionic (Failed)     cpu
	343 - LanczosExpUt.RealTypeTest (Failed)                cpu
	344 - LanczosExpUt.ComplexTypeTest (Failed)             cpu
	345 - LanczosExpUt.NonHermit (Failed)                   cpu
	346 - LanczosExpUt.NormVNot1 (Failed)                   cpu
	347 - LanczosExpUt.OneDimensionalKrylovSpace (Failed)   cpu
	348 - LanczosExpUt.FullKrylovSpaceDoesNotWarnAtDimensionLimit (Failed) cpu
	349 - LanczosExpUt.SmallResidualIsNotBreakdown (Failed) cpu
	350 - LanczosExpUt.FloatSmallResidualBelowRoundoffFloor (Failed) cpu
	351 - LanczosExpUt.FloatResidualAboveRoundoffFloorIsNotBreakdown (Failed) cpu
	352 - LanczosExpUt.ComplexFloatResidualAboveRoundoffFloorIsNotBreakdown (Failed) cpu
	353 - LanczosExpUt.FloatComplexTauReturnsComplexFloat (Failed) cpu
	368 - Lanczos.FermionicLanczosArpackFermionic (Failed)  cpu
	395 - VecTest.VecConcatenate (Failed)                   cpu
	396 - SearchTreeTest.BasicSearchOrder (Failed)          cpu
	404 - SearchTreeTest.IPEPSObservableNetworkContractsQuickly (Failed) cpu
	869 - DenseUniTensorTest.RankZeroDenseContractActsAsScalar (Failed) cpu
	1134 - DenseUniTensorTest.ContractDenseFullContractionBlockIsScalar (Failed) cpu
	1191 - BlockFermionicUniTensorTest.SimpleTensorContract (Failed) cpu
	1194 - BlockFermionicUniTensorTest.TraceRankZeroScalarPreservesSymmetryMetadata (Failed) cpu
	1268 - BlockUniTensorTest.Trace (Failed)                 cpu
	1297 - NconTest.NconDenseUtSpecifiedOrder (Failed)       cpu
	1298 - NconTest.NconDenseUtOptimalOrder (Failed)         cpu
	1303 - NetworkTest.NetworkDenseFindOptimal (Failed)      cpu
	1309 - NetworkTest.NetworkBlockNoOrder (Failed)          cpu
	1310 - NetworkTest.NetworkBlockSpecifiedOrder (Failed)   cpu
	1311 - NetworkTest.NetworkBlockFindOptimal (Failed)      cpu

For `VecConcatenate`

Changing https://github.com/Cytnx-dev/Cytnx/blob/e235906f73d38bc7d1dcd42255de25fdb116ce0f/src/utils/vec_concatenate.cpp#L6-L12
with
```cpp
static_assert(std::is_trivially_copyable_v<T>);

std::vector<T> out(inL.size() + inR.size());

if (!inL.empty())
  std::memcpy(out.data(), inL.data(), sizeof(T) * inL.size());

if (!inR.empty())
  std::memcpy(out.data() + inL.size(), inR.data(),
              sizeof(T) * inR.size());
```
fixes the test.

IvanaGyro:
## Problem

GCC 15 enables `_GLIBCXX_ASSERTIONS` by default for unoptimized (`-O0`) builds, which turns latent STL-precondition violations into hard aborts. `vec_concatenate` and `vec_concatenate_` in `src/utils/vec_concatenate.cpp` built their result with:

```cpp
memcpy(&out[0], &inL[0], sizeof(T) * inL.size());
```

`&inL[0]` (i.e. `operator[]` at index 0) is undefined behavior whenever `inL` is empty — the size check that `_GLIBCXX_ASSERTIONS` adds to `operator[]` now catches this and aborts. Since CI's `compilers`/`cxx-compiler` conda-forge packages bumped g++ from 14 to 15 (2026-08-14), this broke a long list of previously-passing tests that concatenate a possibly-empty vector somewhere in their call path — `Lanczos`, `Arnoldi`, `SearchTree`, `Network`, `UniTensor` contraction/trace, `ncon`, and `VecTest.VecConcatenate` itself — see #1179 for the full failing-test list and root-cause diagnosis.

A second instance of the same bug (`Storage::_from_vector` in `include/backend/Storage.hpp`) surfaced during this PR's own CI run, via `Storage.from_pylist([])` — see the second commit.

## Fix

Rebuilt both `vec_concatenate` functions on `std::vector::insert` against `begin()`/`end()` iterators, which stay valid (and no-op) on empty ranges — no raw indexing, no `memcpy`:

```cpp
template <class T>
std::vector<T> vec_concatenate(const std::vector<T> &inL, const std::vector<T> &inR) {
  std::vector<T> out;
  out.reserve(inL.size() + inR.size());
  out.insert(out.end(), inL.begin(), inL.end());
  out.insert(out.end(), inR.begin(), inR.end());
  return out;
}
```

`insert` works uniformly for every element type, so the hand-rolled index-loop specializations for `std::vector<bool>` and `std::vector<std::string>` are no longer needed and were folded into the generic template; their explicit instantiations (previously commented out, since the specializations covered them) are now enabled instead.

Also tidied the include block to Google-style grouping (own header, then C++ stdlib, then project headers) and made the `Type.hpp`/`cytnx_error.hpp` dependencies (used directly for `cytnx_error_msg` and the `cytnx_*` type aliases) explicit rather than relying on transitive includes.

For `Storage::_from_vector`, guarded each per-dtype `memcpy` with `if (!vin.empty())` and switched `&vin[0]` to `vin.data()` — `memcpy`'s own contract also disallows a null/dangling source pointer regardless of length, so skipping the call for size 0 is the correct fix rather than just silencing the assertion.

## Testing

- `pre-commit run` (clang-format v14) — clean on both changed files.
- `ctest --test-dir build/debug-openblas-cpu`: **1839/1839 passed**, 0 failed (11 pre-existing, unrelated skips).
- `ctest --test-dir build/debug-mkl-cpu`: **1873/1873 passed**, 0 failed (same 11 pre-existing skips).
- `pytest pytests/keep_set_dispatch_test.py::test_storage_from_pylist_empty_list_defaults_to_double`: passes (previously aborted the process with SIGABRT).
- `pytest pytests/ --doctest-modules`: 280 passed, 1 skipped (pre-existing).

Both ctest runs include every test listed as failing in #1179 (`VecTest.VecConcatenate`, `SearchTreeTest.*`, `DenseUniTensorTest.*`, `BlockFermionicUniTensorTest.*`, `BlockUniTensorTest.Trace`, `NconTest.*`, `NetworkTest.*`, `Lanczos*`, `ArnoldiUt.FermionicArnoldiFermionic`, `ZeroExtentLinalgTest.ProductsReturnEmptyOrAdditiveIdentity`), all passing.

Note: local toolchain here is GCC 13.3, so `_GLIBCXX_ASSERTIONS` isn't exercised locally the same way CI's GCC 15 does — but both fixes remove the underlying UB itself (invalid `operator[]` on an empty vector), which is a real bug independent of which compiler/assertion level catches it.

Closes #1179.


ianmccul:
why does vec_concatenate exist?

IvanaGyro:
> why does vec_concatenate exist?

I don't know. I suggest remove them in the future, like some other vec_ functions.

manuschneider (approved):

PR #1186, opened 2026-09-18
fixes a bug in contract when the two fermionic tensors have no common labels

### LINKED ISSUE #1033: Fermionic no-common-label contract suspicious code
Noticed while developting #1026:

Fermionic no-common-label contract appears to combine sign flips using output block index b, while the actual source blocks are 
`IDL[0]` and `IDR[0]`: src/BlockFermionicUniTensor.cpp:1310:

https://github.com/Cytnx-dev/Cytnx/blob/3268385451c9d57e9d056aea96dadb5eae374212/src/BlockFermionicUniTensor.cpp#L1298-L1311

That is probably a real bug, but it needs a focused test and review.

manuschneider:
Fixes #1033
Also minor fixes of comments, and removal of an unused variable.


src/BlockFermionicUniTensor.cpp
```diff
@@ -1391,9 +1393,8 @@ namespace cytnx {
           cytnx_error_msg(out_block.shape() != tmp->_blocks[b].shape(),
                           "[ERROR][BlockFermionicUniTensors][contract] Mismatching shape!%s", "\n");
           tmp->_blocks[b] = out_block;
-          const bool lhs_signflip = this->_signflip[b];
-          const bool rhs_signflip = signflip_rhs[b];
-          tmp->_signflip[b] = (lhs_signflip == rhs_signflip) ? EVEN : ODD;
+          // the output sign is the product of the two source-block signs
+          tmp->_signflip[b] = (this->_signflip[idl[0]] != signflip_rhs[idr[0]]);
```
IvanaGyro, line 1397: This bypass the meaning of the definition of parity enum.

IvanaGyro: Or maybe consider to use `^` to represent it is xor.

IvanaGyro (approved):

src/BlockFermionicUniTensor.cpp
manuschneider: I restored the explicit EVEN/ODD assignment

